Issue #5, opened 2021-09-03
Generating nextpnr model with wire pairs fails


TaoBi22:
Currently running fabric_gen.py -genNextpnrModel_pair throws a KeyError. This seems to be because of how program flags are checked - as the arguments of the program call are converted into a string from a list and then checked for substrings, if one valid flag is a substring of another then both commands will be executed when the superstring flag is used.

This means that when -genNextpnrModel_pair is added as a flag, FABulous picks up both -genNextpnrModel and -genNextpnrModel_pair, and executes both. We then get the KeyError - I'm not entirely sure why but it could potentially be a scope issue with variables being reused.

An easy fix would be to add an escaped ' character to the end of the check for genNextpnrModel, so that the end of the string must be the end of the argument, but this seems confusing and like it could lead to more issues - it could be worth overhauling how arguments are checked?

PR #9, opened 2021-09-10
Fix argument checks

TaoBi22:
As mentioned in #5, FABulous' current method of checking arguments is not ideal as it converts arguments to a string and checks for presence of a substring, meaning that valid flags that are substrings of other valid flags can cause issues. This PR resolves this by creating a processed list of all arguments (by removing surrounding whitespace and making lower case), and checking the presence of valid flags within this list.


TaoBi22:
PR now also updates tests to run the flow with pair generation to make sure it works, since this slipped under the radar previously.

PR #29, opened 2022-07-08
VPR Refactor for Pin Constraints

TaoBi22:
As discussed [here](https://groups.google.com/g/vtr-users/c/IkTgyfHLyQg), in order to have full control over placement of individual BELs through placement constraints, different BELs need to be on separate subtiles within each tile. This PR refactors the way that BELs are represented within VPR tiles to permit this - this also involves some changes to things such as pb_types (wrapper pb_types are now required, simply because each top level pb_type represents a BEL, but in VPR top-level pb_types cannot represent a primitive) and a minor change to the custom XML spec, documentation for which is incoming.

### DIFF
--- a/fabric_files/generic/custom_info.xml
+++ b/fabric_files/generic/custom_info.xml
@@ -1,7 +1,7 @@
 <custom_xml_spec>
 	<bel_info name="LUT4c_frame_config">
 		<bel_pb>
-			<pb_type name="LUT4c_frame_config" num_pb="8">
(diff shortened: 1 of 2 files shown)


TaoBi22:
Changed to WIP as I spotted some remaining necessary changes with FASM generation!

TaoBi22:
FASM metadata generation now supported - this actually works out a little nicer than the previous representation because the user no longer needs to annotate custom XML with FASM prefixes. Just docs left to update.

PR #54, opened 2022-09-06
Fix MULADD sign extension

### LINKED ISSUE #51: MULADD signedness
Something I noticed while looking at adding DSP inference support, which will be slightly harder than planned because Yosys' hard-fought rules for decomposing large multiplies to DSPs assume either configurable signendess - like Lattice, or fixed signed - like Xilinx.

It looks like there was an attempt to add configurable sign-extension: https://github.com/FPGA-Research-Manchester/FABulous/blob/4fe503fe97fdbdbd9374c082a8348a3b9a6dd7c5/fabric_files/fabric_init_verilog_example/MULADD.v#L94

but sign-extending the result of a multiply is _not_ the same as doing a signed multiply itself  - which aiui on a hardware level requires sign extending partial products, in Verilog it would be something like
```
wire signed [15:0] product;
assign product = $signed(OPA) * $signed(OPB);
```
an efficient way to do configurable signedness would require a bit more thought...

This can hopefully be dealt with for existing silicon using the MULADD primitive but we should also consider improving the primitive to support configurable signedness long term to avoid inefficient things happening to signed multiplies in synthesis.

gatecat:
To make a multiply signed, it's not sufficient just to sign extend the product - it's also necessary to sign extend both operands to the full width of the multiply.

Fixes #51

### DIFF
--- a/fabric_files/fabric_init_verilog_example/MULADD.v
+++ b/fabric_files/fabric_init_verilog_example/MULADD.v
@@ -86,9 +86,13 @@ module MULADD (A7, A6, A5, A4, A3, A2, A1, A0, B7, B6, B5, B4, B3, B2, B1, B0, C
 	assign OPB = ConfigBits[1] ? B_reg : B;
 	assign OPC = ConfigBits[2] ? C_reg : C;
 
+	// sign extend inputs when doing a signed multiply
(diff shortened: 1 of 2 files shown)


TaoBi22:
Looks good to me!

PR #55, opened 2022-09-28
Adding an example of simulating with Icarus

gatecat:
This is currently built atop old fabulous so we have a gold reference point to use for testing the refactor. By using Icarus rather than say xsim it's all open source and easy to test in CI, too (this will probably be a follow-up PR).

At the moment it's just using a simple counter design but it's intended to be extendable to more complex designs too, for example for testing the custom IP block integration.

The new upstream nextpnr-generic is used to build the user design, this means that we don't have to build nextpnr ourselves and instead can rely on what the friendly neighbourhood cats have done over at [oss-cad-suite](https://github.com/YosysHQ/oss-cad-suite-build).

Longer term it would be nice to have a suite of tests for different fabric shapes, IP types, etc; that perhaps could run overnight.


TaoBi22:
Looks good to me!

PR #58, opened 2022-10-24
VHDL testing

KelvinChung2000:
Adding all the required files for testing with VHDL. 
Adding a script to generate a top wrapper for VHDL.


fabric_files/fabric_ghdl_example/build_test_design.sh
```diff
@@ -0,0 +1,14 @@
...
+
+SYNTH_TCL=../../nextpnr/fabulous/synth/synth_fabulous.tcl
+BIT_GEN=../../nextpnr/fabulous/fab_arch/bit_gen.py
+
+
+DESIGN=counter
+
+set -ex
+# yosys -qp "tcl $SYNTH_TCL 4 top_wrapper test_design/${DESIGN}.json" test_design/${DESIGN}.v test_design/top_wrapper.v
+yosys -qp "tcl $SYNTH_TCL 4 top_wrapper test_design/${DESIGN}.json" ../generic/IO_1_bidirectional_frame_config_pass.v test_design/${DESIGN}.v test_design/top_wrapper.v
```
gatecat: hmm why is `IO_1_bidirectional_frame_config_pass` needed this should be part of the blackbox primitives already I think? adding this as a non-blackbox would break the IO constraints I think?

KelvinChung2000: I had problems before when synthesising the design, so I added that. It seems like it is still working, and I will change it back anyway, just in case.

PR #63, opened 2022-11-25
Prune and clean docs

TaoBi22:
There are a couple of pages in the docs that were left as placeholders a very long time ago - since they've not been filled yet, this PR removes them from the navigation bar so they can no longer be navigated to. The files are still there under the assumption that the pages will eventually be finished - if we decide against this they should obviously be removed. Note that this leads to a couple of warnings when generating the docs (as Sphinx detects that there's no navigation to the pages).

There's also a bit of general docs clean up included here.

### DIFF
--- a/docs/source/Usage.rst
+++ b/docs/source/Usage.rst
@@ -74,7 +74,7 @@ Example:
 
 The output <benchmark_name>_output.bin can be used in further simulation.
 
-VPR models can be found under ``/vproutput``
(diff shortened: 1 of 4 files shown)


gatecat (approved):

TaoBi22:
Looks like the warnings may show as a build fail on ReadTheDocs, so I'll remove the pages in another PR.

PR #65, opened 2022-12-04
Update models_pack_sim.v & Update models_pack.v

ruck314:
### Description
- Removing `reg X` from `cus_mux41` because not register and causing error when doing FPGA emulation


gatecat:
Thanks!

PR #66, opened 2022-12-04
Update fabulous_tb.v

ruck314:
### Description
- Bug fix to the enable the counter (active HIGH) after the reset pulse


ruck314:
Closing ticket.  My confusion was that the simulation and documentation instructions were using the same counter source code, which is not the case. The `counter.v` does not use enable while the `sequential_16bit_en.v` does.  Sorry for the noise. 
https://github.com/FPGA-Research-Manchester/FABulous/blob/master/fabric_files/fabric_icarus_example/test_design/counter.v

PR #71, opened 2022-12-08
Cadence genus syn fixes

ruck314:
### Description
- These updates fix the synthesis erros that I encountered when using Cadence Genus to synthesis the eFPGA_top.v

```
Error   : Illegal initial assignment for i/o declaration. [VLOGPT-79] [read_hdl]
        : Declaration 'active' in file '/u1/eFpgaFabric/hdl_output/bitbang.v' on line 8, column 20.
        : Illegal Verilog syntax is encountered.
        output reg active = 1'b0;

Error   : Illegal initial assignment for i/o declaration. [VLOGPT-79] [read_hdl]
        : Declaration 'LongFrameStrobe' in file '/u1/eFpgaFabric/hdl_output/ConfigFSM.v' on line 14, column 29.
        output reg LongFrameStrobe = 0;
```

### DIFF
--- a/fabric_files/fabric_simulation_example/ConfigFSM.v
+++ b/fabric_files/fabric_simulation_example/ConfigFSM.v
@@ -13,7 +13,7 @@ module ConfigFSM (CLK, WriteData, WriteStrobe, Reset, FrameAddressRegister, Long
 	input Reset;
 	
 	output reg [FrameBitsPerRow-1:0] FrameAddressRegister;
-	output reg LongFrameStrobe = 0;
+	output reg LongFrameStrobe;
 	output reg [RowSelectWidth-1:0] RowSelect;
 	
 	reg FrameStrobe = 0;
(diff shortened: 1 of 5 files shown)


gatecat:
Hi @ruck314,

Unfortunately this is breaking the simulation (probably by uncovering a hardware bug, as initial states will be ignored for VLSI synthesis anyway). This was being missed by CI because the uninitialised config logic and hence fabric caused it to output all `x`s and the checker didn't catch that - see https://github.com/FPGA-Research-Manchester/FABulous/pull/72. Once this PR is on top of that CI should (correctly) fail.

My suggested fix is to add an explicit reset at the top level, as is conventional in VLSI synthesis, and add async resets for all the signals that currently rely on initial values. Does this sound good to you too?

ruck314:
Hi @gatecat,

Refer to the recent commits that I did this morning. I think this would address both the output being initialized to zero (prevent breaking the CI) and the `[VLOGPT-79]` error message that I am getting.  What do you think?

gatecat:
This does indeed unbreak the CI, but, I think relying on initial values in this way in a way that will actually be ignored for ASIC synthesis is potentially covering up an issue.

ruck314:
Ok, I understand your point of view on this. I'll close the ticket

PR #73, opened 2022-12-09
config: Use resets everywhere instead of init values

gatecat:
Init values are ignored for ASIC synthesis and so dangerous to rely on. Instead, add async resets everywhere so we can start the config logic in a known good state. This is intended as an alternative solution to #71, and generally to better comply with VLSI good practice.

@ruck314 can you confirm this fixes the Cadence Genus synthesis?

### DIFF
--- a/fabric_files/fabric_icarus_example/fabulous_tb.v
+++ b/fabric_files/fabric_icarus_example/fabulous_tb.v
@@ -6,6 +6,7 @@ module fab_tb;
     wire [55:0] A_cfg, B_cfg;
 
     reg CLK = 1'b0;
+    reg resetn = 1'b1;
     reg SelfWriteStrobe = 1'b0;
     reg [31:0] SelfWriteData = 1'b0;
     reg Rx = 1'b1;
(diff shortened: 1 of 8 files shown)


ruck314:
@gatecat I got these two warning/error messages for ConfigFSM.v:
```
Error   : Illegal initial assignment for i/o declaration. [VLOGPT-79] [read_hdl]
        : Declaration 'LongFrameStrobe' in file '/fabulous-28nm/asic/targets/eFpgaFabric/hdl_output/ConfigFSM.v' on line 15, column 29.
        : Illegal Verilog syntax is encountered.
        output reg LongFrameStrobe = 0;
                                      |
Warning : Ignoring unsynthesizable construct. [VLOGPT-37]
        : Initial value assignment to reg in file '/fabulous-28nm/asic/targets/eFpgaFabric/hdl_output/ConfigFSM.v' on line 15, column 32.
        : For example, the following constructs will be ignored:
    - initial block
    - final block
    - program block
    - property block
    - sequence block
    - covergroup
    - checker block
    - gate drive strength
    - system task enable
    - reg declaration with initial value
    - specify block.
```

gatecat:
Sorry, missed that one, the `ConfigFSM_template.v` has now been fixed accordingly, can you try again?

ruck314 (approved):

ruck314:
@gatecat I have confirmed that [97a9b85](https://github.com/gatecat/FABulous/commit/97a9b858e1cec49210340fd1b91029ff650944ca) fixed the `[VLOGPT-79]` for ConfigFSM.v and bitbang.v.

Issue #79, opened 2023-01-13
VPR is not working with Yosys 0.19+0


KelvinChung2000:
When running the VPR flow using Yosys 0.19+0 yields the following error. 
```
Error 1: 
Type: Blif file
File: ../nextpnr/fabulous/fab_arch/sequential_16bit_en.blif
Line: 105
Message: Failed to find matching architecture model for 'LUTFF_ESR'
```
Looks like Yosys mapping has some problem causing this error.

gatecat:
`LUTFF_ESR` mapping shouldn't be enabled in the normal synthesis script, which one are you running?

KelvinChung2000:
Using `./nextpnr/fabulous/fab_arch/fabulous_flow.sh`, the `.json` changed to `.blif`. 
The yosys command is the following:
`yosys -p "tcl ../synth/synth_fabulous_dffesr.tcl 4 ${1} ${1}.blif" ${1}.v -l ${1}_yosys_log.txt`

TaoBi22:
Since we now have an upstream Yosys synthesis pass, is it worth just removing the old synthesis script and the surrounding documentation, and recommending the upstream Yosys pass instead? (As a note, based on a quick test just now I'm also getting issues with the upstream pass and VPR but I'll make another issue for that once I've explored it more).

TaoBi22:
Never mind, the upstream pass actually seems to work fine and the issues were on my end!

TaoBi22:
Should now be fixed in #80

PR #80, opened 2023-01-13
Switch to upstream Yosys synthesis pass

TaoBi22:
This makes the documentation changes I proposed in the #79 discussion, removing discussion of the previous synthesis script and giving instructions to use the pass that is now upstream in Yosys. If we merge this, then we can probably also get rid of all the synthesis stuff in the nextpnr fork. I think I've covered everywhere the old script was mentioned, but let me know if I've missed anything! I also suspect that the page on Yosys models (`docs/source/FPGA_CAD-tools/yosys.rst`) is pretty old, so it would be great if you could check that it's still accurate @gatecat.

### DIFF
--- a/docs/source/FPGA-to-bitstream/Yosys compilation.rst
+++ b/docs/source/FPGA-to-bitstream/Yosys compilation.rst
@@ -13,21 +13,21 @@ To build you may use the Makefile wrapper in the clone repository (https://githu
 User guide
 ----------
 
-A pre-defined Yosys TCL script is under ``$FAB_ROOT/nextpnr/fabulous/synth/synth_fabulous_dffesr.tcl`` for FABulous version3. If the output file, denoted below as ``<JSON_or_BLIF_file>``, has a ``.blif`` file extension, then output will be produced in the Berkeley Logic Interchange Format (BLIF) and will be synthesised appropriately for the VPR flow. Otherwise, output will be produced as JSON, synthesised for the nextpnr flow.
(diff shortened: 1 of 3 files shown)


docs/source/FPGA-to-bitstream/Yosys compilation.rst
```diff
@@ -13,21 +13,21 @@ To build you may use the Makefile wrapper in the clone repository (https://githu
 User guide
 ----------
 
-A pre-defined Yosys TCL script is under ``$FAB_ROOT/nextpnr/fabulous/synth/synth_fabulous_dffesr.tcl`` for FABulous version3. If the output file, denoted below as ``<JSON_or_BLIF_file>``, has a ``.blif`` file extension, then output will be produced in the Berkeley Logic Interchange Format (BLIF) and will be synthesised appropriately for the VPR flow. Otherwise, output will be produced as JSON, synthesised for the nextpnr flow.
+FABulous is supported by upstream Yosys, using the ``synth_fabulous`` pass. First, run ``yosys``, which will open up an interactive Yosys shell. If synthesizing for the nextpnr flow, run the following commands:
 
 .. code-block:: console
 
-	yosys -p "tcl <path/to/TCL/File>/synth_fabulous_dffesr.tcl <K-LUT> <top_module> <JSON_or_BLIF_file>" <benchmark_netlist>
```
gatecat, line 20: personally, I'd recommend using it non-interactively like
`yosys -p "synth_fabulous -top <toplevel> -json <out.json>" <files.v>`

as this is less typing and easier to include in scripts (like our examples).

TaoBi22: Agreed! I'll get that changed.

gatecat (approved):

PR #50, opened 2022-08-11
FABulous refactor

KelvinChung2000:
Refactored the whole fabric_gen.py script for fabric generation.

 File structure changes
- The `fabric_gen.py ` now only contains the function related to generating fabric and the bit stream specification.
- `fabric.py `  contains all the data structures used by the `fabric_gen.py `
- `file_parser.py` contains all the parsers which will convert things like CSV files and HDL files into objects that the `fabric_gen.py` can work with
- `code_generator.py`  is the abstract base class for code generation. The `code_generation_verilog.py` will be for outputting Verilog syntax and  `code_generation_VHDL.py` for outputting VHDL code. 
- `model_generation_npnr.py` and `model_generation_vpr.py` generate the Nextpnr model and VPR model.
- `utilities.py`, as the name suggests, all the helper functions. 

Architectural changes
- `fabric_gen.py ` is now generating code with the python object defined in `fabric.py ` whenever is possible
- `fabric_gen.py ` can generate VHDL code and Verilog with the same logic. 
- `fabric_gen.py ` now can also generate the top wrapper without using the `top_wrapper_generator_with_BRAM.py`
- We now do not need to generate Verilog code generation is independent of VHDL code generation 

Other changes
- Minor changes to the structure of the configMem CSV, now the header of the file is not commented, and the `ConfigBits_ranges` is now semicolon ";" separated rather than a comma separated, which allows for better use of the python csv module and better readability when reading the file with spreadsheets applications
- Due to both RTL codes being generated with a single logic source. The VHDL code will be equivalent to the Verilog code but different from the current version of the VHDL output code. 
- modified the "template" file in `fabulous_top_wrapper_temp` to propagate the parameter with the parameter feature provided by Verilog rather than doing string hacks with the files.  This also results in we can propagate all the variable changes with the top module, and the current `fabric_gen.py ` is taking advantage of that
- Change the code generation of Verilog to follow the Verilog 2001 standard 
- Make the MATRIX entry in the fabric.csv to accept `.list`, `.csv`, `.vhdl` and `.v` for initial parsing for calculating the amount of configuration bit in a tile. 
- Changed some of the Verilog module names so the name of the module will match the file name.

 
Tasks List:
- [x] Refactor model generation to work with object
- [x] Refactor bit stream spec generation 
- [x] Allow bit stream spec generation to work with user-defined bel
- [x] Make  `FABulous.py` using the `fabric_gen.py` as module
- [x] Testing and verifying the refactored code will generate a correct design
[...]

PR #83, opened 2023-02-23
Cleanup README and gitignore

TaoBi22:
Simple PR just cleaning up the README, which is currently pretty out of date (which is probably on me, sorry!) and wiping a load of ancient gitignore directories that haven't been relevant since the infancy of the project. Found these while reviewing #50, but this shouldn't create any nasty conflicts.


gatecat:
Thanks!

Issue #86, opened 2023-03-03
Documentation Builds Failing


TaoBi22:
As can be seen [here](https://readthedocs.org/projects/fabulous/builds/19664465/), the docs are failing to build as of #50 being merged. Looks to me like maybe the issue is not being able to find the FASM library when trying to load in the libraries for autodocs - I'll submit a PR to add FASM to the requirements, and hopefully that'll fix it, but I'm not 100% certain.

PR #87, opened 2023-03-03
Add FASM to docs requirements

TaoBi22:
Adds the `fasm` Python library to the docs requirements to try and fix the build failures, as discussed in #86.


gatecat:
Thanks!

PR #89, opened 2023-03-06
Add updated emulation support to refactored FABulous

gatecat:
This re-adds updated support for emulation; which enables fast testing of a bitstream against the fabric in simulation or on hardware by passing the bitstream via macros/parameters rather than simulating the entire configuration logic.

Compared to pre-refactor FABulous, there are two improvements:
 - Tiles with custom permuted bitstream layouts work properly (e.g. the LUT4AB)
 - The bitstream is passed through the hierarchy as a parameter rather than a wire, which will enable pre-flattening optimisations with Yosys, important for some of my future plans (like a cxxrtl sim experiment I did with the MPW2 demo design).

The first commit in the series is not related to the emulation but made debugging this along the way a lot easier, because I could actually see what exceptions occurred rather than just a generic Tcl error.

@KelvinChung2000 would you be able to take a look at this one?

### DIFF
--- a/FABulous.py
+++ b/FABulous.py
@@ -200,6 +200,17 @@ To run the complete FABulous flow with the default project, run the following co
             self.fabricLoaded = True
 
     def preloop(self) -> None:
+
+        def wrap_with_except_handling(fun_to_wrap):
+            def inter(*args, **varargs):
+                try:
+                    fun_to_wrap(*args, **varargs)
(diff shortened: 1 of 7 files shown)


gatecat:
Hopefully fixed now, let's see...

KelvinChung2000:
Looking all good to me

PR #91, opened 2023-03-13
fabric_gen: Fix user clock connectivity between supertiles

gatecat:
@KelvinChung2000 would you be able to review?


KelvinChung2000:
The rest is looking good for me

PR #93, opened 2023-03-24
Adding break_comb_loop to models_pack.v

ruck314:
### Description
For Synopsis DC compiler, you can add the following before elaboration and synthesis to deal with the combinatorial loops from the switch matrix
```TCL
set_disable_timing [get_pins -hierarchical *switch_matrix/*BEG*]
```
However for Cadence Genus, you can not apply such constraints until after elaboration and synthesis.  This makes even the simplest eFPGA design take 20+ hours to build.  After the 20+ hours, Cadence Genus addes `cdn_loop_breaker` buffers in all combinatorial loop path to address the timing issue. 

Similar to how you manually add the `clk_buf` with respect to the silicon process primitive, I propose we do a similar strategy for the combinatorial loop paths in the model_pack. 
```verilog
module break_comb_loop(input A, output X);
   NAME_OF_CONFIDENTIAL_PRIMITIVE_HERE cdn_loop_breaker1(
      .I (A),
      .Z (X));
endmodule
```
By manually inserting the loop breaker primitive into the combinatorial loop path, I can reduce my simple eFPGA design from 20+ hours to 40 minutes to elaboration and synthesis.

### DIFF
--- a/fabric_files/FABulous_project_template_verilog/Fabric/models_pack.v
+++ b/fabric_files/FABulous_project_template_verilog/Fabric/models_pack.v
@@ -158,9 +158,15 @@ module cus_mux41 (A0, A1, A2, A3, S0, S0N, S1, S1N, X);
 	input S1N;
 	output X; 
 	wire [1:0] SEL;
+	wire [3:0] AIN;
 
-	wire B0 = S0 ? A1 : A0;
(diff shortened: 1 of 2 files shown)


gatecat:
Thanks!

PR #99, opened 2023-04-13
Fail docs CI on warning

TaoBi22:
As mentioned in #98, our Read the Docs setup fails builds when they produce warnings - this keeps resulting in build failures on our Read the Docs site flying under the radar, as the CI is still passing. This passes a custom build command to the action we use to build the docs for CI that is the same as that run by the Makefile, save for the addition of a few flags Read the Docs gives - this means CI should now fail for builds that will fail on Read the Docs.

An alternative to this is to update the docs Makefile so that these flags are always provided, but maybe it's preferred that warnings don't affect users building them locally just because that's what Read the Docs happens to do. Equally though, you could argue that we don't want local doc builds out of sync with what's online - what's your opinion @gatecat?

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -14,4 +14,4 @@ jobs:
     - uses: ammaraskar/sphinx-action@master
       with:
         docs-folder: "docs/"
-        build-command: "sphinx-build -M html source build -T -W --keep-going"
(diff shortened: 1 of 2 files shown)


TaoBi22:
> LGTM - I expect the main reason for local docs builds is to work on them; and in that case failing on warnings seems sensible too
Good point! I'll revert and change the Makefile instead

PR #107, opened 2023-09-06
Fix docs build failing due to RTD deprecated attribute

TaoBi22:
[Our most recent docs build failed](https://readthedocs.org/projects/fabulous/builds/21823928/) as the system packages attribute that we set in `.readthedocs.yaml` is now deprecated. This just gets rid of the attribute which should hopefully fix it!


TaoBi22:
Seems to have fixed the problem!

Issue #104, opened 2023-06-14
No such file or directory: '.FABulous/maxWidth.txt' while running gen_model_vpr


Asma-Mohsin:
![image](https://github.com/FPGA-Research-Manchester/FABulous/assets/65780913/09548785-ee7f-48b0-8e1d-4b975f78e105)

KelvinChung2000:
Have you got the `.FABulous` folder? Since this is trying to write a file to the `.FABulous` folder. 

As a side note, VPR is currently not tested. I suggest working with Yosys and Nextpnr exclusively.

TaoBi22:
This issue should now be fixed in #108!

PR #108, opened 2023-09-06
Move `maxWidth.txt` handling to `FABulous.py`

TaoBi22:
This should fix the issue noted in #104 - `genVPRRoutingResourceGraph` has no knowledge of the project directory so can't write to the `.FABulous` directory to create the `maxWidth.txt` file. This just adds the max width as part of a return tuple so that the file handling can be done in `FABulous.py` as is done elsewhere.


gatecat:
Thanks!

PR #122, opened 2023-11-03
Fix file name of the bitstream specification file in the documentation.

IAmMarcelJung:
The file "meta_data.txt" does not seem to exist even after the flow is complete and the command works with "bitStreamSpec.bin". This file is also used in the function do_gen_bitStream_binary in FABulous.py.


IAmMarcelJung:
I revoked the change of the FASM file name since gen_bitstream_binary uses a "_gen.fasm" file for the bitstream generation, but does expect just a ".fasm" file to be specified in the command and adds the "_gen" internally.

docs/source/FPGA-to-bitstream/Bitstream generation.rst
```diff
@@ -46,11 +46,11 @@ With the FASM file generated by Nextpnr or VPR, you can run this command for bit
 Example
 ^^^^^^^
 
-The following are simple commands to synthesis the netlist ``sequential_16bit_en`` into a JSON netlist.
+The following are simple commands to synthesize the netlist ``sequential_16bit_en`` into a bitstream file.
```
TaoBi22: I think maybe synthesize is a misleading term, as we'd normally associate that with synthesis into a netlist. How about:
```suggestion
The following command lowers the FASM representation of the placed and routed ``sequential_16bit_en`` design into a bitstream file.
```

TaoBi22 (approved):
Thanks for this!! I've tried to pick up all the bits of documentation that are still hanging around from before we did our big CLI implementation and refactor but I guess I've still missed bits!

docs/source/FPGA-to-bitstream/Bitstream generation.rst
IAmMarcelJung: Yes your're right, the original sentence doesn't make any sense there, but I only saw the typo. Thanks!

TaoBi22:
Great, thanks!

PR #123, opened 2023-11-03
Fix the calculation of the addressable config bits

IAmMarcelJung:
The calculation of the addressable configuration bits seems to be wrong since j is given by 32 and i by 20 and i x j = 32 * 20 = 640. Also removed trailing whitespaces.


TaoBi22 (approved):
Thanks again!

IAmMarcelJung:
Glad I could help! :)

PR #125, opened 2023-11-06
Fix place_and_route_npnr help and error message

IAmMarcelJung:
Instead of the directory of the top module, the command expects the ".fasm" file to be specified.


KelvinChung2000:
The description of  `<dir_to_top_module>` is because running nextpnr needs multiple files, not just the `<fasm_file>`

IAmMarcelJung:
Thanks for your reply! Which files are needed in addition to the ```<fasm_file>```? For me it works with just the ```<fasm_file>``` and fails with anything else. If there are more files needed, this should be documented somehow or the the command should really accept just the ```<dir_to_top_module>```. Or am I doing something wrong?

IAmMarcelJung:
Sorry, I have to correct myself, you don't need a ```<fasm_file>```, but rather ```<dir_to_top_module>/<name_of_the_top_module>``` (```<sequential_16bit_en``` in the demo) since the file ending gets split in the script. The input file for running the nexpnr command is a ```<json_file>``` and I don't see any other files needed. So maybe this should be changed to ```<json_file>``` and in the script the stripping could be removed?

KelvinChung2000:
Yea, `<json_file>` might be better. The help message could also indicate the json file is generated by yosys. This should make things clearer.

IAmMarcelJung:
I changed it to ```<json_file>```, but I also had to change the call of ```do_place_and_route_npnr``` in ```do_run_FABulous_bitstream```, since there the command was called with a verilog file. So for the manual call this is (IMHO)  an improvement, but for the "automated" call, I think this makes things worse, since the other functions are called with the verilog file and it would be less uniform. IMHO we have two choices: Call all functions just with the top module name, so there are no irritations about the file ending/type or call all functions with the explicit file type/ending. 

I'm not entirely sure about this, so I'm really open for other opinions and suggestions!

KelvinChung2000:
If we use the top module name, we will approach the user design more like an "object". Then running the manual synthesis command will be like I want that specific "object" getting synthesis. This might be good when the user starts to work with a more complex design. But I am not sure how to get across this sort of "object" mindset similar to what we have in the CAD tools with the CLI tool. 

We could have a new command like `init_user_design <name>` and `list_object` to get across the idea of "object". But this feels very stupid. We start introducing unnecessary complexity because we want to make things more coherent.

IAmMarcelJung:
I agree, we should not introduce a new function just for this. So you suggest working with the explicit file endings in every function instead? Or would you rather use the "mixed" version as it is now?

Personally, I would suggest using the explicit version, even tough it would introduce some extra code (though no substantial extra complexitiy). It would align the functions better to the flow graph and it would also be more consistent than the "mixed" version.

KelvinChung2000:
Yes, explicit will be better.

IAmMarcelJung:
I made the changes so each of the commands now has to be called by specifying the correct file including the file ending.

KelvinChung2000:
I think we should move all the changes here to #130 since they are related, and #130 will have better handling than what we have here

IAmMarcelJung:
Yes I agree and applied the changes. Everything related to this will be handled in #130 from now on.

PR #126, opened 2023-11-06
Match wire names in the documentation of the wire shifting example

IAmMarcelJung:
The source and destination names in the wire shifting example did not match to the names in the corresponding image on the [fabric definition](https://fabulous.readthedocs.io/en/latest/fabric_definition.html) page for the wire shifting example.


TaoBi22 (approved):
Thanks! :)

PR #127, opened 2023-11-07
Add exception handling for the synthesis CLI commands

IAmMarcelJung:
The CLI always crashed when there was an error during synthesis (e.g. file not found or error in files). The traceback also "hid" the error with its output a few lines above. Now the user just gets notified that the synthesis failed, the error is just above the short logging message and the CLI does not crash so the user does not have to restart FABulous and reload the fabric after fixing the error. If it was just a typo in the path or file, the mistake can just be corrected from the history without completely retyping the command which is necessary after restarting FABulous.


TaoBi22 (approved):
Looks good to me! I also had issues with this testing #128 - I would guess it's probably worth making similar changes to the nextpnr call as well but I'm happy to put that in another PR. Requested review from Kelvin too since he wrote the CLI.

KelvinChung2000 (approved):
This is looking good to me.

Issue #124, opened 2023-11-06
Clarification of differences between ".fasm" and "_des.fasm" files


IAmMarcelJung:
In the flow, some commands use a ".fasm" file and some use a "_des.fasm" file. Some commands, like "gen_bitstream_binary" expect the specification of a ".fasm" file, but use "_des.fasm" internally. When comparing the CLI commands to the manual commands, the CLI way with "place_and_route_npnr"  creates a "_des.fasm" (although it expects just a ".fasm" file in the command), while the manual way  creates just a ".fasm" file (although this could obviously be changed by the user of course, [see documentation](https://fabulous.readthedocs.io/en/latest/FPGA-to-bitstream/Nextpnr%20compilation.html)).

The command gen_bitStream_binary also creates a ".fasm" file, but the content does not quite look like a ".fasm" file as specified in the  [FASM documentation](https://fasm.readthedocs.io/en/latest/index.html), at least for my understanding. The file consists of 14 blocks that look like this (all the same except for the first line, where just the 14 changes down to 1 until the last block):
```
X9Y14,RAM_IO,9,14
frame0,0,32,00000000000000000000000000000000
frame1,1,32,00000000000000000000000000000000
frame2,2,32,00000000000000000000000000000000
frame3,3,32,00000000000000000000000000000000
frame4,4,32,00000000000000000000000000000000
frame5,5,32,00000000000000000000000000000000
frame6,6,32,00000000000000000000000000000000
frame7,7,32,00000000000000000000000000000000
frame8,8,32,00000000000000000000000000000000
frame9,9,32,00000000000000000000000000000000
frame10,10,32,00000000000000000000000000000000
frame11,11,32,00000000000000000000000000000000
frame12,12,32,00000000000000000000000000000000
frame13,13,32,00000000000000000000000000000000
frame14,14,32,00000000000000000000000000000000
frame15,15,32,00000000000000000000000000000000
frame16,16,32,00000000000000000000000000000000
frame17,17,32,00000000000000000000000000000000
frame18,18,32,00000000000000000000000000000000
frame19,19,32,00000000000000000000000000000000
```

Could someone explain the difference between a "_des.fasm" and ".fasm" and when which one is used? Additionally, is there a reason why sometimes a ".fasm" file has to be specified, but internally a "_des.fasm" file is used and its existence is expected? Is the distinction even necessary?

Thanks in advance for any help!

IAmMarcelJung:
[...]

PR #128, opened 2023-11-09
Fix filetype confusion and CSV generation bug

TaoBi22:
#124 pointed out that the CSV representation of bitstreams is output into a FASM file instead of a CSV file - this fixes this and adds a couple of comments to be a little clearer. Also noticed a bug in the generation of this CSV, which was just thanks to some incorrect indentation so that's fixed here too.


IAmMarcelJung:
Thanks, this now works for me and also the bits are not all zeros and there are also all tiles included!

PR #129, opened 2023-11-10
Prevent FABulous crash by checking for existing directory in do_place_and_rout_npnr

IAmMarcelJung:
This checks if the specified directory exists and therefore prevents the CLI from crashing in ```if f"{name}.json" in os.listdir(f"{self.projectDir}/{path}"):``` if the specified directory does not exist. 

I don't know why git marks so many lines as changed, since I only added a surrounding ```if``` and an additional ```else```. Is this just do to the indention on python or is it my mistake? Sorry if the latter is the case and please tell me how an can correct it in that case.


TaoBi22 (approved):
LGTM! The big diff is nothing to worry about, like you said it's just that the lines inside your new if statement are now indented so Git picks them up as modified.

IAmMarcelJung:
Okay, thanks for your feedback!

KelvinChung2000 (approved):
Looking good to me.

PR #130, opened 2023-11-10
Prevent more crashes of the shell with more error handling

IAmMarcelJung:
With this modification, more errors of subprocesses are handled, leading to less crashes of the shell.
With the use of the return value it is still guaranteed that ```complete_gen_bitStream_binary``` stops when a subprocess resulted in an error.


KelvinChung2000:
I prefer using plain return than using  `return 1`. This is because the empty return will return `None`, and when testing `None` will be considered `False`. We will not need the `command() != 1` for the test case.

IAmMarcelJung:
Thanks for your reply! If we just used empty return, how would we distinguish between failure and success of a function? A successful function also returns ```None``` after completion.  if no ```return``` statement is hit. Or am I misunderstanding your point?

KelvinChung2000:
You are right. If the return is not hitting, it will still return None. I am now considering whether raising exceptions is better than handling it with return 1. What do you think?

IAmMarcelJung:
I made a first implementation with the exceptions. I'm not entirely sure if this is the best way to do it, but it works and we do not need the return values anymore. Let me know what you think about it. The overwriting of ```onecmd()``` is adapted from [this Stackoverflow post](https://stackoverflow.com/questions/50899713/python-cmd-module-return-to-prompt-after-catching-exception).

KelvinChung2000:
The current `onecmd` cannot capture all errors. I suggest just use `except` to capture any raised error. Then, in `do_run_FABulous_bitstream`, we will not need `try except` for each call since `onecmd` will catch any raised when calling the command.

IAmMarcelJung:
Changes from #125 have been applied here. Everything related to #125 will be handled here from now on.

KelvinChung2000:
Would you mind fixing the typo, please?

IAmMarcelJung:
Yeah sure, but I don't know which typo you mean. Could you please tell me were it is? Thanks!

KelvinChung2000:
I have mentioned this in a review. On line 807 should be `raise` not `return`

IAmMarcelJung:
Ok thanks, I will get that fixed! But I could not find the review you mentioned since it shows me ```No reviews``` in this PR.

KelvinChung2000:
Probably because I did not say myself as a reviewer, if you go into the file changed, you should be able to see my comment. Supprise you do not get pinged about it.

IAmMarcelJung:
Sorry I still could not find it. However I think I messed up some things on my side since I think I already committed the changes, but forgot to push them to the remote. Sorry for that! Was you comment a recent one or was it already there last year? I normally get notified about all kinds of stuff, but for this I didn't get anything.

KelvinChung2000:
Now is on line 867

IAmMarcelJung:
Yeah, you're right. Thanks for spotting it!

IAmMarcelJung:
I will change that!

Which tests do you mean? Are there any tests besides the GitHub Actions?

IAmMarcelJung:
I looked into the code and I saw that there are multiple occurrences of ```split(".")```, also in parts I did not change. I think that's why I used it in the first place. I personally think it would be better to move the substitution of ```split(".")``` into another PR, so we keep things separated, but this is of course debatable.

If there are tests available and they are failing, we should of course fix them in this PR.

If you again commented on a specific line in the code, unfortunately I still cannot see which line the comment is about, although I saw you assigned yourself as a reviewer.

KelvinChung2000:
It is fine now. For some reason, yesterday, when I tried the tool, something went wrong.

IAmMarcelJung:
Nice, but if you something went wrong yesterday, there might still be a bug. Can you recall what the error was and what you roughly did to get the error?

KelvinChung2000:
I figured it out now. The `run_FABulous_bitstream` when calling with a path like `./path/to/file.v `, using `file_ending = verilog_file_path.split('.')[1]` will not be able to find the file ending correctly. For similar reason 

```
verilog_file_path = args[1]
        if len(verilog_file_path.split('.')) != 2:
```
will also fail.

IAmMarcelJung:
Yes I can reproduce the error. Thanks for pointing it out. I will create a new PR were the file name information will be extracted using ```pathlib```, as you previously suggested.

PR #132, opened 2023-11-10
Add a persistent command history to the CLI

IAmMarcelJung:
This adds a persistent command history to the CLI, which is also available after closing and restarting the CLI.
Adapted from this [stackoverflow post](https://stackoverflow.com/questions/39495024/persistent-history-in-python-cmd-module).


KelvinChung2000:
I will suggest storing the history in the `.FABulous` folder. So, each project can have its own CLI history.

IAmMarcelJung:
Nice suggestion! It is now implemented.

KelvinChung2000:
We will not need the `.fabulous_history` in the `.gitignore` file then.

IAmMarcelJung:
Ah, yes of course. Fixed!

PR #141, opened 2023-11-30
Fix small typo in the documentation

IAmMarcelJung:
I think ```of``` is correct here instead of ```or```.


TaoBi22:
Yep, thanks for the catch!

Issue #148, opened 2024-02-02
Exception messages not printed


IAmMarcelJung:
Since #130, the exception message are not printed in the shell anymore. Since we already have logging, the exception message might also not be need and could therefore be removed. 

If we however need the messages, something like 
```
except Exception as e:
    print(e)
    return False
```
inside ```onecmd``` would again print the exception message. We should therefore decide if either omitting or printing the exception message is preferred.

IMHO it would be better to print the messages, but any other suggestions (also different ones to these two options) are welcome as always.

KelvinChung2000:
Printing the exception message will be better, as sometimes the cause of the error might be absent. Directly logging the error message will also be helpful.

KelvinChung2000:
fixed on #152

PR #152, opened 2024-02-16
Print traceback on error inside FABulous Shell

mole99:
If an error/exception occurs inside the FABulous Shell it is not printed, only `False` is returned:

```
    def onecmd(self, line):
        try:
            return super().onecmd(line)
        except:
            return False
```

You can easily test this by adding a syntax error to e.g. the `do_load_fabric` function:

```
    def do_load_fabric(self, args=""):
        "load csv file and generate an internal representation of the fabric"
        args = self.parse(args)
        asdf
        ....
```

The FABulous shell does not indicate that an error has occurred:

```
FABulous> load_fabric
FABulous> 
```

This PR fixes that:

```
FABulous> load_fabric
Traceback (most recent call last):
  File "/home/leo/Projects/FABulous_fix/FABulous.py", line 276, in onecmd
    return super().onecmd(line)
           ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/cmd.py", line 217, in onecmd
    return func(arg)
           ^^^^^^^^^
  File "/home/leo/Projects/FABulous_fix/FABulous.py", line 318, in do_load_fabric
    asdf
NameError: name 'asdf' is not defined

FABulous> 
```

### DIFF
--- a/FABulous.py
+++ b/FABulous.py
@@ -43,6 +43,7 @@ import logging
 import tkinter as tk
 from pathlib import PurePosixPath, PureWindowsPath
 import platform
+import traceback
 readline.set_completer_delims(' \t\n')
 histfile = ""
 histfile_size = 1000
@@ -274,6 +275,7 @@ To run the complete FABulous flow with the default project, run the following co
         try:
             return super().onecmd(line)
         except:
+            print(traceback.format_exc())
             return False
 
     # override the emptyline method, so empty command will just do nothing


IAmMarcelJung:
Nice! This should fix #148.

mole99:
Thanks for linking the issue! I haven't even seen it ^^

KelvinChung2000 (approved):

Issue #153, opened 2024-02-17
VHDL generation broken


mole99:
VHDL generation seems to be broken on current master. PR #152 now shows the proper error.

I ran the following commands:

```
python3 FABulous.py -c demo -w vhdl
python3 FABulous.py demo -w vhdl
load_fabric
run_FABulous_fabric
```

`run_FABulous_fabric` fails with:

```
[INFO]-2024-02-17 11:32:35,363 - Output file: demo/Tile/S_term_DSP/S_term_DSP.vhdl
Traceback (most recent call last):
  File "/home/leo/Repositories/FABulous/FABulous.py", line 276, in onecmd
    return super().onecmd(line)
           ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/cmd.py", line 217, in onecmd
    return func(arg)
           ^^^^^^^^^
  File "/home/leo/Repositories/FABulous/FABulous.py", line 526, in do_run_FABulous_fabric
    self.do_gen_fabric()
  File "/home/leo/Repositories/FABulous/FABulous.py", line 458, in do_gen_fabric
    self.do_gen_all_tile()
  File "/home/leo/Repositories/FABulous/FABulous.py", line 452, in do_gen_all_tile
    self.do_gen_tile(self.allTile)
  File "/home/leo/Repositories/FABulous/FABulous.py", line 441, in do_gen_tile
    self.fabricGen.genTile(t)
  File "/home/leo/Repositories/FABulous/FABulous.py", line 140, in genTile
    self.fabricGenerator.generateTile(tile)
  File "/home/leo/Repositories/FABulous/fabric_generator/fabric_gen.py", line 788, in generateTile
    raise ValueError(
ValueError: Could not find S_term_DSP_switch_matrix.vhdl in Tile/S_term_DSP/./ Need to run matrix generation first
```

The issue seems to be that `fabric_gen.py` searches for the file `Tile/S_term_DSP/S_term_DSP_switch_matrix.vhdl` from the current working directory, which is: `/home/leo/Repositories/FABulous`. But the file is in `/home/leo/Repositories/FABulous/demo/Tile/S_term_DSP/S_term_DSP_switch_matrix.vhdl`.

Once this is fixed, it would be good to add a test case for VHDL to the CI.

EverythingElseWasAlreadyTaken:
Fixed with #309 and #312

PR #154, opened 2024-02-17
Separate fabric and tile configuration

### LINKED ISSUE #149: Separate Fabric and Tile description
I think it would be clearer and more convenient to separate the fabric and tile descriptions in `fabric.csv`.

The file `fabric.csv` would then only contain the fabric description. The tile descriptions would reside in the tile-specific subfolders as e.g. `tile.csv`.

This makes sense as the tile configurations could be used for several fabrics and do not need to be duplicated.

mole99:
This PR separates the fabric and tile configuration. This has several benefits:

- It's cleaner:
    The fabric config only describes the fabric and the `Tile/` folder contains all information about tiles
    The git history for tile changes / fabric changes is separate
- It is easier easier to swap whole "tile libraries", just point to different tiles in `fabric.csv`
- Code is a bit more separated

I carefully tried to not break anything and tested this PR with both the Verilog and VHDL writers: `load_fabric` is successful for both of them. For Verilog all subsequent commands work as expected, for VHDL `run_FABulous_fabric` throws an error that is unrelated to this PR. See #153.

Fixes #149.


fabric_generator/file_parser.py
KelvinChung2000, line 1: We should keep the backward compatibility with the single CSV format.

fabric_generator/file_parser.py
```diff
@@ -245,7 +95,12 @@ def parseFabricCSV(fileName: str) -> Fabric:
         i = [j for j in i if j != ""]
         if not i:
             continue
-        if i[0].startswith("ConfigBitMode"):
+        if i[0].startswith("Tile"):
+            parseTile(os.path.abspath(os.path.join(filePath, i[1])), tileTypes, tileDefs, commonWirePair)
```
KelvinChung2000: Rather than passing the object and appending the item to them, returning the object is preferred. Returning new objects makes the code more readable.

fabric_generator/file_parser.py
```diff
@@ -245,7 +95,12 @@ def parseFabricCSV(fileName: str) -> Fabric:
...
+            tileDic = dict(zip(tileTypes, tileDefs))
+        elif i[0].startswith("Supertile"):
+            parseSupertile(os.path.abspath(os.path.join(filePath, i[1])), superTileDic, tileDic)
```
KelvinChung2000: same for here

fabric_generator/file_parser.py
mole99: That I hadn't thought of that! Should now support the old fabric configuration format. I had to change `parseTiles` and `parseSupertiles` to return a list of objects.

fabric_generator/file_parser.py
mole99: Done, but what should I do with `commonWirePair`? Keep it as it is or return a tuple?

fabric_generator/file_parser.py
mole99: Done

fabric_generator/file_parser.py
KelvinChung2000: We can do the same. Return as a tuple, then unpack directly. Something like `new_tiles, commonWirePair = parseTiles(os.path.abspath(os.path.join(filePath, i[1])))`

fabric_generator/file_parser.py
KelvinChung2000: Rather than deprecation, we can introduce a new keyword, for example, `TILE_PATH`, to indicate a path is provided. Will also need to update the Doc to let people know this is available.

mole99: The keywords that are used are `Tile` and `Supertile` to read the tile/supertile configuration from a path. Should I rename them?

I would keep the deprecation notice. It's not much work to move an existing fabric to the new format and the FABulous template uses the new format as well. Supporting both options is more work to maintain and I think it makes generally more sense when the configurations are split this way.

I would update the docs with the new way of describing tiles, if that's okay?

fabric_generator/file_parser.py
mole99: Done

fabric_generator/file_parser.py
KelvinChung2000: Yes, this makes sense.  

Let's keep the `Tile` and `Supertile` keywords since the `TILE` keyword will only happen in the "tile.csv". 

Update the Doc with the new tile description method and mark deprecation for the current method. When the "2.0" release is in place, we can entirely remove the current way.

mole99: I have now updated the docs to refer to the new method of tile description.

What I noticed: The example fabric uses the `CPU_IO` tile which is currently not available for Verilog and misses configurations in VHDL. I think it would be nice to have this available, should I open an issue regarding the missing files?

KelvinChung2000: Yes, open an issue about it. Someone is currently working on the automatic generation of the tile functionality. This should help with the situation.

mole99: Done! I think this PR should be ready then.

KelvinChung2000:
I think you will have other significant changes as well. We will need to maintain the current file structure for a tutorial shortly. I have started a new branch called `FABulous2.0-development`; let's have file structure changes happen there, and when it is all ready, we can put everything into master.

mole99:
Sounds good! I have changed the base branch to `FABulous2.0-development`.

PR #110, opened 2023-09-08
Build FABulous with setuptools

EverythingElseWasAlreadyTaken:
I've refactored FABulous to build it with setuptools. 

This would allow us to publish FABulous to PyPI.
I've already prepared a little POC and published it on  TestPyPI: 
https://test.pypi.org/project/FABulous-FPGA/

It was neccessary to rename the package ot FABulous-FPGA, since there is already a fabulous package on PyPI.

You can install it with:
```
$ python -m pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple FABulous-FPGA
```
The `--extra-index-url` is needed since some dependencies are not available on TestPyPI.

Afterward, you can use `FABulous` as command line tool:
```
$ FABulous -c demo
$ FABulous demo
$ FABulous --script ./demo/FABulous.tcl
```

PR #163, opened 2024-02-26
Add formatter 

KelvinChung2000:
I have added the black workflow. Please let me know if anything else needs to be done besides this `yml` file.


mole99:
Looks good! The CI worklow failed due to invalid formatting.

Now developers need to be instructed to use black before submitting their changes. One way to do this is to add a badge in the README, as OpenLane 2 does for example: https://github.com/efabless/openlane2
But I think it would also be good to mention that in the documentation.
And last but not least, this PR should probably change the formatting to be compatible with black.

KelvinChung2000:
Will need to write a section on contribution guidelines.

KelvinChung2000:
I also need to avoid significant code base changes. Even though this is only changing the look, who knows, I have broken something mid-way when the CI is not exhaustive enough. I will put this into 2.0 for now so the setup tool can proceed.

mole99:
Hi! I just wanted to ask if this is ready to merge?
It's easier to develop on top of these changes as they touch the entire codebase rather than fixing conflicts later on.

I would suggest merging #164 first and then this PR to master since they are fixes/formatting changes.
Then #110 can be rebased and also merged to master, so that a package of current FABulous can be uploaded to PyPI.
Then a new 2.0 development branch can be created where I can reapply #154 onto.
Further changes should then be done on the 2.0 branch only to avoid conflicts. You could also think about just calling the branch `dev` like other projects do.

Thanks!

KelvinChung2000:
Dirk would like to keep this merge happen later. How about you keep the development going with the formatting applied locally? When it's ready, I will merge this first and follow up with your updates. The amount of conflict should be minimal in that case.

mole99:
Okay, will do! Right now I am not actively working on FABulous, but will get back to it later.
I just thought it would be nice to finally have a package on PyPI and #110 is waiting for this PR.

mole99:
@KelvinChung2000 is there any news on this?

Would it be possible to merge this PR now and apply the formatter on the feature being worked on in the background? In that case, the conflicts should also be minimal. Then #110 could rebase on this PR and we could finally have a Python package of the current FABulous.

KelvinChung2000:
The workshop is over. I will ask Dirk what he thinks. I will at least merge this to development first so everyone has a common ground.

mole99:
Thanks! :+1:

PR #167, opened 2024-03-07
Add documentation of MID wires

### LINKED ISSUE #140: Missing documentation of MID wires
It seems that there is no documentation for MID wires. The documentation could include an image visualizing the feature with a short explanation.

IAmMarcelJung:
This would resolve #140. The image is a screenshot taken from inside [FABulator](https://github.com/FPGA-Research-Manchester/FABulator). Although it might not be the best visualization due to the thin wires, it's something and maybe could be replaced in the future. Either by an updated version if thicker lines are supported or if somebody wants to draw an image themselves, this would also to the trick.

The explanation might also be a bit flawed but this is as far as I understood it from explanation by  @dirk-koch and the visualization of FABulator. Also I'm not sure if I inserted the content into the correct place in the text. 

As always, any feedback is therefore highly appreciated!


docs/source/fabric_definition.rst
```diff
@@ -279,6 +279,17 @@ specifying:
   In summary, the example shows how a termination tile can be used to provide more complex interface blocks and all this can be easily modelled and implemented with FABulous.
   
   .. note::  The ``destination_name`` is refering to the port name used at the destination tile. FABulous will throw an error if the destination tile does not provide that port name.
+
+  Aside from ``BEGIN`` and ``END``, there also exist ``MID`` ports, which are used for wires spanning over two tiles.
```
TaoBi22: This phrasing might be slightly less ambiguous?

```suggestion
  Aside from ``BEGIN`` and ``END``, there also exist ``MID`` ports, which can be used for wires spanning more than two tiles.
```

TaoBi22 (approved):
This looks good to me, just a couple of minor suggested tweaks. Thanks for this!

docs/source/fabric_definition.rst
```diff
@@ -279,6 +279,17 @@ specifying:
...
+  Although they route over two tiles, they also have a tap on the middle tile.
+  On the middle tile, the sink is called ``MID`` while the source is still a ``BEG`` port with the addition of ``b``.
+  In the example below this is illustrated. The wire on the left goes from ``E2BEG0`` to ``E2MID`` which is the tap on the middle tile.
+  Inside the switch matrix, ``E2MID`` and ``E2BEGb`` are connected. The left wire the starts at ``E2BEGb`` and ends at ``E2END``.
```
TaoBi22:
```suggestion
  Inside the switch matrix, ``E2MID`` and ``E2BEGb`` are connected. The left wire then starts at ``E2BEGb`` and ends at ``E2END``.
```

dirk-koch:
Hi folks,
I would leave the mid ports.
The reason is that Xilinx used a lot so called MID ports in older FPGAs.
For instance Double wired had a MID port after one switch matrix and hex wires after 3 switch matrices respectively.
In FABulous, this is simply modeled by concatenating two shorter segments (two singles for a double or two triple for a hex wire).
So the E2BEGb is then the begin port of the second single leg forming a double with a MID port.
I hope this is not too confusing.
There is some documentation on this in the readthedocs.
The same method can be used to model the diagonal routing available in Virtex-V (if somebody wants to model that architecture in FABulpous).

docs/source/fabric_definition.rst
IAmMarcelJung: Thanks for catching this!

docs/source/fabric_definition.rst
IAmMarcelJung: Yeah seems better! Thanks!

IAmMarcelJung:
> Hi folks, I would leave the mid ports.

@dirk-koch We are not planning to remove the MID ports but rather wanted to add some documentation for them, since not everyone might be as familar with the Xilinx stuff as you are ;D With that said, do you agree with the short description that was added? Or is there anything you would improve or add?

TaoBi22:
@IAmMarcelJung I got the go-ahead from Dirk so I'll get this merged!

IAmMarcelJung:
Nice, thanks!

PR #176, opened 2024-05-07
README+Docs: Add virtual environments as recommended usage way

EverythingElseWasAlreadyTaken:
Add description how to use FABulous with virtual environments in README and docs, since it is recommended for working with python projects.

Introduce \$, (venv)\$ and FABulous> as commandprompt prefixes.


KelvinChung2000:
Since you are doing the document, could you add a line about FASM installation? The standard installation will yield a warning.

KelvinChung2000:
When I try to install with `pip install -e .` I also noticed that the tool fails with the following error. 
```
has a 'pyproject.toml' and its build backend is missing the 'build_editable' hook. Since it does not have a 'setup.py' nor a 'setup.cfg', it cannot be installed in editable mode. Consider using a build backend that supports PEP 660.
```

EverythingElseWasAlreadyTaken:
> I am doing a global installation. I will do it the venv way. The document will also need to mention this then.

For this reason I made this PR :)

It is generally no longer recommended to install python packages with pip globally, since it can break Python system packages and can cause problems with shared dependencies.

In newer Python versions (>=3.11) it will throw an error, if you try to install any packages globally with pip.
You can still bypass this error, but it is generally recommended to use venvs.

EverythingElseWasAlreadyTaken:
> When I try to install with `pip install -e .` I also noticed that the tool fails with the following error.
> 
> ```
> has a 'pyproject.toml' and its build backend is missing the 'build_editable' hook. Since it does not have a 'setup.py' nor a 'setup.cfg', it cannot be installed in editable mode. Consider using a build backend that supports PEP 660.
> ```

This issue seems to be caused by an old version of setuptools. 
Which would explain, why it works on my machine, if I install FABulous it globally.

I've updated the pyproject.toml, could you please try to install it again globally from this branch?

KelvinChung2000:
If the global install is deprecated, should we only stick to the venv installation?

IAmMarcelJung:
I still get the error on global install, even with the `setuptools>=64` requirement.

We could also change the documentation from recommending a venv to making a venv a prerequiste, since it is clearly a better way to install the package, for the reasons you mentioned @EverythingElseWasAlreadyTaken .

EverythingElseWasAlreadyTaken:
> I still get the error on global install, even with the `setuptools>=64` requirement.

Maybe you have to update setuptools manually and also update pip to make this work.
At least this is what people with similar issues did here: https://github.com/pypa/setuptools/discussions/4037
 
> We could also change the documentation from recommending a venv to making a venv a prerequiste, since it is clearly a better way to install the package, for the reasons you mentioned @EverythingElseWasAlreadyTaken .

The venv instaallation is the only way described in the documentation.
The global installation isn't described anymore.

EverythingElseWasAlreadyTaken:
> Since you are doing the document, could you add a line about FASM installation? The standard installation will yield a warning.

Should i just add a Note, that this warning is normal? Or is there a way to fix it?

KelvinChung2000:
To make the warning disappear, we will need to install fasm from the source. I think just putting a note is all we need.

IAmMarcelJung:
> The venv instaallation is the only way described in the documentation. The global installation isn't described anymore.

Okay then I will not try to make the global installation work since I won't use it personally.

IAmMarcelJung:
> Maybe you have to update setuptools manually and also update pip to make this work. At least this is what people with similar issues did here: [pypa/setuptools#4037](https://github.com/pypa/setuptools/discussions/4037)

Just for completion, @EverythingElseWasAlreadyTaken and I talked again about this and it was indeed the update of pip that made the global installation work.

PR #177, opened 2024-05-14
fabric_generator:file_parser: Refactor parseFileVHDL and parseFileVerilog

EverythingElseWasAlreadyTaken:
Refactor parseFileVHDL to make it more readable.
This fixes also the bug, that shared INPUT ports in VHDL files, were added as INOUT ports.

Add port direction variable in parseFileVerilog function to make it more readable.

Some cosmetic changes.

Move commit [8cb8aab](https://github.com/FPGA-Research-Manchester/FABulous/pull/177/commits/8cb8aabcf5631c92aa5ab8b25260fac4ee32a0c0) (Make expandListPorts public) from #176 to this PR.


KelvinChung2000:
I have done something similar recently. You can check out [here](https://github.com/KelvinChung2000/FABulous/commit/6db4e6fb2cd7fcc0c3af79152b8161aa2ebda001#diff-771374aab3db1156c58027c5be11f12a5bebac2f451f82b912fb1d8987e61b85). If this is helpful for you I will make a pull request publish that out.

EverythingElseWasAlreadyTaken:
> I have done something similar recently. You can check out [here](https://github.com/KelvinChung2000/FABulous/commit/6db4e6fb2cd7fcc0c3af79152b8161aa2ebda001#diff-771374aab3db1156c58027c5be11f12a5bebac2f451f82b912fb1d8987e61b85). If this is helpful for you I will make a pull request publish that out.

Thanks for the hint, but I see some issues with your changes for the parseFileVerilog/VHDL.
For the parseFIleVHDL, you still have this big chunk of if-else statements, which is already fixed in the parseFileVerilog. Also in the parseFileVerilog, you are treating INOUT ports always as OUTPUT ports, since you only check for INPUT ports. 

The other changes are quite nice, we should somehow combine them. 
Or you could accept this PR and rebase your commit to this. 

I've also though about, combining the parseFileVHDL and parseFileVerilog in one function, since the code is like 80% same. 
This would improve the maintainability and reduce the likelyhood for errors.

KelvinChung2000:
The main problem with combining is the subtle difference in parsing. You will have one blob for Verilog and another for VHDL, which might make the code look bulky again. But this should indeed reduce the amount of things that need maintenance. 

I will rebase mine on yours later. But I have other things to do first, so that won't happen soon. 

The ideal way would be to do it in a parser way, which reads in both Verilog and VHDL, and then we extract data under a common AST... But this in itself will be a project.

EverythingElseWasAlreadyTaken:
> The main problem with combining is the subtle difference in parsing. You will have one blob for Verilog and another for VHDL, which might make the code look bulky again. But this should indeed reduce the amount of things that need maintenance.

Yes, but I think this will mostly concern the regex lines. 

> I will rebase mine on yours later. But I have other things to do first, so that won't happen soon.

Okay. Sure, there is no rush. 

> The ideal way would be to do it in a parser way, which reads in both Verilog and VHDL, and then we extract data under a common AST... But this in itself will be a project.

Yes, this would be really nice, but this will be indeed a big project.

KelvinChung2000:
Since you are refactoring, you can include it here as well and rename the PR to mention refactoring both.

EverythingElseWasAlreadyTaken:
> Since you are refactoring, you can include it here as well and rename the PR to mention refactoring both.

I've merged them into one parseFile function now.
Would be nice if you could review my changes. :)

FABulous/fabric_generator/file_parser.py
```diff
@@ -737,42 +583,70 @@ def parseFileVerilog(filename: str, belPrefix: str = "") -> tuple[
...
+            if result := re.search(r"(.*):(.*)", result):
+                portName = f"{belPrefix}{result.group(1)}"
+                if result.group(2).upper() == "IN":
+                    direction = IO["INPUT"]
+                elif result.group(2).upper() == "OUT":
+                    direction = IO["OUTPUT"]
+                elif result.group(2).upper() == "INOUT":
+                    direction = IO["INOUT"]
+                else:
+                    direction = None
```
KelvinChung2000: We can raise a value error here since we will never reach this point.  With the raise, it should also notify the static checker that `direction` is always defined.

FABulous/fabric_generator/file_parser.py
```diff
@@ -737,42 +583,70 @@ def parseFileVerilog(filename: str, belPrefix: str = "") -> tuple[
...
-            isConfig = False
-            isShared = False
+        if "UserCLK" in portName:
+            userClk = True
+
+        isExternal = False
+        isConfig = False
+        isShared = False
 
     return internal, external, config, shared, noConfigBits, userClk, belMapDic
```
KelvinChung2000: This is a historical mess that I made. We can create the `Bel` object here instead of doing it at the main parser logic. This will also allow the function to be used standalone, and much easier to use.

EverythingElseWasAlreadyTaken:
@KelvinChung2000 have you already had some time to look at it again?

I have some other works that are partly based on this.

KelvinChung2000:
I have put some comments on it. I am not sure why when I made an review it did not notify you

KelvinChung2000 (changes requested):

KelvinChung2000:
Does it work now?

KelvinChung2000 (approved):

EverythingElseWasAlreadyTaken:
> Does it work now?

Yes, saw the Review now, but I didn't get any notifications before this. 

Thanks, I'll look into this :)

FABulous/fabric_generator/file_parser.py
EverythingElseWasAlreadyTaken: Fixed :)

FABulous/fabric_generator/file_parser.py
EverythingElseWasAlreadyTaken: Fixed :)

PR #194, opened 2024-06-12
init commenting

A-Kibats:
initial expansions of comments and familiarisation with Git workflow.


A-Kibats:
Was put to main via Error. Request pulled immediately.

Issue #201, opened 2024-06-27
Switch-Matrix contains commented-out ports


hoyer-ims:
Trying to integrate some custom-tiles, I have used the existing tiles and just commented out some ports in the .v of the new BEL with /* */ and //. I have adjusted the switch_matrix.list and .csv for the BEL and removed all of the obsolete ports. However, they still appear in the switch_matrix.csv after FABulous has run, triggering a warning as well: 

[WARNING]-2024-06-27 10:01:12,999 - Input port C15 of switch matrix in Tile MS is not used
[WARNING]-2024-06-27 10:01:12,999 - Input port C16 of switch matrix in Tile MS is not used

I am not sure whether the issue impacts the overall flow of FABulous or whether it is just a thing that can be ignored. It took me a while to figure out that it was the commented-out ports, that caused the issue. Deleting the commented-out part fixed it. 

It can easily be reproduced by adding a commented-out additional port to e. g. the MULADD.v
![grafik](https://github.com/FPGA-Research-Manchester/FABulous/assets/165947311/dedad6ea-6017-4157-99af-2d1df36edf9d)
![grafik](https://github.com/FPGA-Research-Manchester/FABulous/assets/165947311/0b0e0797-6124-4b86-9610-383c83e1ab09)

hoyer-ims:
Naming ports double in the BEL.v (one time commented-in, one time commented-out) results in no warning, but in an issue when nextpnr tries to route. It gives an assertion failure: 

terminate called after throwing an instance of 'nextpnr_generic::assertion_failure'
  what():  Assertion failure: bi.pins.count(name) == 0 (/home/hoyer/GIT/Fabulous/work/fabulous_efpna/nextpnr_efpna/generic/arch.cc:147)
[ERROR]-2024-06-27 10:53:43,504 - Placement and Routing failed.

KelvinChung2000:
Only commenting will keep the port in because the current implementation uses a regular expression to find the port, and commenting will not work. Hopefully, we will use an HDL parser to extract information soon.

Whether a write is double depends on the `.csv` file using the offset attribute. You should be able to name it anything.

PR #202, opened 2024-06-30
Vhdl synthesis

Biswajitks1:
The VHDL synthesis is accomplished using the GHDL Yosys plugin. However, the top wrapper (constraints) must be in Verilog because GHDL does not support attributes like (* BEL ...*). The documentation has also been updated accordingly. Currently, VHDL synthesis is only compatible with nextpnr.


KelvinChung2000:
I know that VHDL also has attributes, but I don't know how to use them or how VHDL creates custom attributes. If this is non-trivial, we can stick with this hybrid method. However, if it is straightforward, we can apply the same method for BEL port parsing.

Biswajitks1:
I attempted to define custom attributes in VHDL and used GHDL to process them. However, GHDL issued a warning stating that it didn't recognize these attributes and synthesis was not working.I will try to find a way to tell ghdl interpret custom attributes.

KelvinChung2000:
You probably have looked at this, but I will still include it here https://stackoverflow.com/questions/56444320/ghdl-simulator-doesnt-support-vhdl-attributes-without-error

Biswajitks1:
Thank you for sharing. I tried that.
The synth_fabulous command, specifically when it takes the top module with the following command: 
synth_fabulous -top top_wrapper. Does synth_fabulous support VHDL files? In the website: https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/synth_fabulous.html
they mentioned the commands executed by running this is read_verilog.So I kept the top_wrapper as verilog.

docs/source/Usage.rst
```diff
@@ -49,18 +49,22 @@ with ``source venv/bin/activate`` to use FABulous.
...
 
 
+Install FABulous with "editable" option:
+.. code-block:: console
+
+    (venv)$ pip install -e .
+
 The following packages need to be installed for the CAD toolchain
 
+If you are using yosys oss-cad-suite, no need to install the following packages.
```
IAmMarcelJung, line 75: I would suggest to merge this sentence with the previous one to something like:

The following packages need to be installed for the CAD toolchain if you are not using the yosys oss-cad-suite.

Biswajitks1:
By using the [synthesis script](https://github.com/FPGA-Research-Manchester/FABulous/pull/fabric_cad/synth/synth_fabulous_dffesr.tcl) it may be possible to proceed. However, I am concerned about how attributes are handled. They could be turned into warnings. I need to verify whether GHDL writes the attributes into the .o and elaboration files it creates. I will look into this.
It is better to have the top_wrapper in VHDL. I will explore the possibilities for this approach.

Also I found out that yosys can output the synthesized netlist in a pictorial/vector format should we add this? There are also some tools like [netlistsvg](https://github.com/nturley/netlistsvg) which can convert the json files into svg.

KelvinChung2000:
Creating warnings is fine as long as we know what is causing the warning. The attribute you created in VHDL should get picked up by yosys and GHDL if they both confirm the exact specification. 

Showing a post-synthesis netlist is not that important at the moment since most people will not be viewing it. However, if you plan to spend time on IDE integration, you can revisit the topic. 

If the custom attribute is not causing problems, we can change the VHDL BEL file to attribute base. This should then enable us to unify our BEL parsing by using yosys, which solves issues like #201. But this will be in another pull request.

KelvinChung2000:
@Biswajitks1 Hi, I would like to know if you have got the VHDL attribute working?

Biswajitks:
I didn't give my 100% in it. I started working on the simulation part, which is now complete. I will create a pull request for it soon. For now, I think we can proceed with the mixed synthesis approach.  I have a few doubts about the VHDL simulation, which I will include in my pull request. Once the simulation is finalized, I'll shift my full attention to the synthesis.

IAmMarcelJung:
Sorry I currently don't have the time to do this. Also as mentioned I had problems with the dev branch version (see #245) , which however could be fixed. I can look into this in April the earliest. Until then it should be enough to simulate it, as we do for all other changes :)

Biswajitks1:
ok I will check the simualtion.

Biswajitks1:
I checked the simulation it works.

docs/source/FPGA-to-bitstream/Yosys compilation.rst
```diff
@@ -38,7 +38,17 @@ file is located at ``user_design/sequential_16bit_en.v`` then the result of the
 
 Manual Synthesis
 ^^^^^^^^^^^^^^^^
-FABulous is supported by upstream Yosys, using the ``synth_fabulous`` pass. First, run ``yosys``, which will open up an interactive Yosys shell. If synthesizing for the nextpnr flow, run this command: ``yosys -p "synth_fabulous -top <toplevel> -json <out.json>" <files.v>``.
+FABulous is supported by upstream Yosys, using the ``synth_fabulous`` pass. First, run ``yosys``, which will open up an interactive Yosys shell.  
+
+For Verilog projects:
+ If synthesizing for the nextpnr flow, run this command: ``yosys -p "synth_fabulous -top <toplevel> -json <out.json>" <files.v>``.
+
+ If you are synthesizing for the VPR flow, then run this command: ``yosys -p "synth_fabulous -top <toplevel> -blif <out.blif> -vpr" <files.v>``.
```
KelvinChung2000: We do not support VPR anymore, so this is not needed.

KelvinChung2000 (approved):

docs/source/FPGA-to-bitstream/Yosys compilation.rst
Biswajitks1: Yeah missed that one. Removed it now.

PR #211, opened 2024-07-05
Fix for VHDL Project Creation Issue

Biswajitks1:
This PR resolves an issue where VHDL projects created with `FABulous -c <name> -w vhdl` were generating `verilog` fabric and tile files when `FABulous <name>` was used followed by `run_FABulous_fabric`. 

To address this, the project should be started with `FABulous <name> -w vhdl`. The solution involved creating a `.env` file to store the project's language setting upon creation. This language setting is then read from the `.env` file to configure the writer accordingly. Subsequently, executing `FABulous <name>` now functions correctly.
In future the .env file can be used to store some specifications of the project.


FABulous/FABulous.py
KelvinChung2000, line 1: Can you use `pathlib` to handle the paths and avoid using `os`. For this use case, it probably makes no difference, but `pathlib` is more portable than `os`

KelvinChung2000 (approved):

FABulous/FABulous.py
Biswajitks1: ok I will do that.

EverythingElseWasAlreadyTaken:
Hey
Is there any progress on this?
I'd like to use the dot-env for some other ideas. :)

PR #198, opened 2024-06-18
:fabric_gen: Fix VHDL ConfigMem gets instantiated twice per tile

EverythingElseWasAlreadyTaken:
Remove the double call of addComponentDeclarationForFile, which caused the ConfigMem for a VHDL tile got instantiated twice, while the tile code was generated.

@KelvinChung2000: I'm not sure how this affects with the shift register based configuration type. Could you please check, if I should rather remove the upper "addComponentDeclarationForFile" call in the generateTile function?

PR #214, opened 2024-07-29
Vhdl simulation

Biswajitks1:
While working on the simulation of a VHDL design, I encountered an error related to the resetn signal. I realized that the VHDL fabric files were not updated to match the current Verilog files, so I updated them accordingly. However, when I attempted to run the simulation, it quit before executing the run simulation command of ghdl.

To diagnose the problem, I used GDB (GNU Project Debugger) and discovered that the issue was related to the system's stack size. The default stack size on Linux is 8192 kB. By increasing the stack size to unlimited, the simulation was able to run, but it took approximately 2 hours and 30 minutes to complete.

I'm looking for ways to reduce the simulation time. I used the optimization flags of ghdl but they didn't help much. Any ideas or suggestions would be greatly appreciated.

### DIFF
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -46,6 +46,7 @@ readline.set_completer_delims(" \t\n")
 histfile = ""
 histfile_size = 1000
 
+simulation_decide = 0  # for verilog
 MAX_BITBYTES = 16384
 
 logger = logging.getLogger(__name__)
@@ -101,6 +102,15 @@ def copy_verilog_files(src, dst):
     for root, _, files in os.walk(src):
         for file in files:
             if file.endswith(".v"):
+                source_path = os.path.join(root, file)
+                destination_path = os.path.join(dst, file)
+                shutil.copy(source_path, destination_path)
+
+
+def copy_vhdl_files(src, dst):
+    for root, _, files in os.walk(src):
+        for file in files:
+            if file.endswith(".vhdl"):
                 source_path = os.path.join(root, file)
                 destination_path = os.path.join(dst, file)
                 shutil.copy(source_path, destination_path)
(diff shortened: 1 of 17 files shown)


KelvinChung2000:
I have also had a long simulation time problem before. The long simulation and the large stack size might be due to how the mux is written. If you check the Verilog simulation pack, you can see the mux is coded in a specific way to minimise the number of events generated, which should help speed up the simulation.  

What machine specs are you using by the way?

Biswajitks1:
I have implemented buffer similar to break comb loop of verilog in vhdl that didn't help.I will once again go through the implementation of the mux in verilog. My system is i5 10th gen with 32 GB of RAM. The same user_design in the verilog does the simulation in few minutes on the same system.

KelvinChung2000:
The comb loop brake is needed to make the simulation run. The mux design will reduce the amount of `x` propagated within the design, which should aid the problem. If the simulation is still taking too long, we might need to investigate what is slowing the simulation down in detail.

Biswajitks1:
ok thank you. I will look into the mux implementation.

Biswajitks1:
Hi, I implemented the muxes in my_lib.vhdl in a different way using the verilog lib file as reference but that didn't solve the issue.....

KelvinChung2000:
What sort of memory usage are you seeing? 

Maybe drop an issue on GHDL GitHub and see if they can offer any help. 

You can observe the waveform and find cell is leading to a lot of signal change, then optimise from there. You can also try CXXRTL simulation or translate the design to Verilog and then feed it to Verilator, which might undo a lot of the work you have done. Other than that, I cannot think of any solution.

Biswajitks1:
I was busy with my academics for the past week didn't spend enough time on this. I planned to work on this weekend. This simulation branch is built on vhdl synthesis. So that needs to be merged first.

KelvinChung2000:
Yes. That's fine. But can you still change it so it merges with the master?

Biswajitks1:
ok I will do that.I will create a single pull request for both synthesis and simulation for the master branch.

Biswajitks1:
The pull request #198 should also need to be merged in master. @EverythingElseWasAlreadyTaken can you please look into this.

EverythingElseWasAlreadyTaken:
> Yes. That's fine. But can you still change it so it merges with the master?

Are you sure, that we should merge this into master and not the development branch? 
We should avoid adding too much in the master, while the dev branch is under active development, or it will be a lot of work to merge master and dev later, without breaking the commit history. 
In my opinion, we should only add truly necessary fixes in the master, anything else should be in dev.

KelvinChung2000:
I want it in master because this is testing. Whether it's in master or development, it should be the same. But if the testing depends on something currently in development, then let's keep this in development.

Biswajitks1:
Ok.So then we are sticking to development branch right?
I found an old discussion on [Ghdl-discuss](https://ghdl-discuss.gna.narkive.com/zSSLBBye/huge-simulation-speed-slowdown) where it was mentioned that there was an issue with the threshold value in GHDL's source code, which was later fixed. This seems somewhat related to the issue I'm currently facing: the simulation isn't consuming more than 1GB of RAM. I'm not sure whether this is a problem or not. I've opened an issue on GHDL, and hope to get a solution.

KelvinChung2000:
Is this your [issue](https://github.com/ghdl/ghdl/issues/2732)? I think you can provide them a zip of all the files, which should help them to look at the problem

Biswajitks1:
yes. ok I will do that thank you.

Biswajitks1:
Hi, apologies for the delay. I got caught up with other tasks and was also waiting for a reply from the GHDL developers. Regarding the issue we discussed, here's the link to the relevant GitHub issue: [GHDL Issue #2732](https://github.com/ghdl/ghdl/issues/2732).

The developer replied with the following:  "I haven't worked on it yet. It's a long time project."

I don't have much experience with Ada, but I tried tweaking the GHDL code to address the memory issue. Unfortunately, none of the changes I made resolved the problem. I think we can just go ahead with the current simulation method. Atleast we have a running simulation, it just takes long time.

KelvinChung2000:
Sure. I will merge this once all the conflicts are resolved.

Biswajitks1:
Before merging this the [VHDL Synthesis](https://github.com/FPGA-Research/FABulous/pull/202). pull request should be merged. I tried for vhdl attributes to work didn't got any proper solution. Should we go ahead with the current synthesis?

EverythingElseWasAlreadyTaken:
> Hi, apologies for the delay. I got caught up with other tasks and was also waiting for a reply from the GHDL developers. Regarding the issue we discussed, here's the link to the relevant GitHub issue: [GHDL Issue #2732](https://github.com/ghdl/ghdl/issues/2732).
> 
> The developer replied with the following: "I haven't worked on it yet. It's a long time project."
> 
> I don't have much experience with Ada, but I tried tweaking the GHDL code to address the memory issue. Unfortunately, none of the changes I made resolved the problem. I think we can just go ahead with the current simulation method. Atleast we have a running simulation, it just takes long time.

If GHDL is the issue for simulation, we could maybe try out [NVC](https://github.com/nickg/nvc) for VHDL simulation and only do synthesis with GHDL. 

For #202 could you please fix the merge conflicts there? Otherwise, we can't merge it.

Biswajitks1:
Ok sure. I will fix the merge conflicts and try the NVC.

Biswajitks1:
I tried the nvc, it is slower compared to GHDL. I got some idea I will try that and update accordingly.

Issue #213, opened 2024-07-29
Waveform


CarlosQbit:
Hello, I hope you are having a great day. I am trying to compare the waveforms of the test circuit (I found out it was a counter) with the waveform of the ouput of the eFPGA after simulation, after following all the Quick Start Documentation and Simulation Setup Documentation:

```
python3 FABulous.py -c <name_of_project>
python3 FABulous.py <name_of_project>
# inside the FABulous shell
load_fabric
run_FABulous_fabric
run_FABulous_bitstream npnr user_design/sequential_16bit_en.v

cd demo/Test
./build_test_design.sh
./run_simulation.sh
```

 I did what is on the picture to generate the waveform because otherwise I could not find how to. iI would be glad if someone could explain if there is a better way to generate the waveform and compare it. Also if there is some documentation or an explanation on how to try a different circuit besides the implemented counter, for examle an AND gate or a multiplier.  Also if there is a way to see the waveform of the bitstream. Thanks in advance!
![Captura desde 2024-07-28 18-21-17](https://github.com/user-attachments/assets/47ad5350-bc05-45c8-8d65-1541776c4c20)

KelvinChung2000:
Since we are using Icarus, this is the only way to generate a waveform. If you are trying to compare different waveforms, the only thing I am aware of is a relatively recent paper on waveform comparison language tools, which I am not familiar with. The main problem is that two waveforms can be different but perform the same thing. Furthermore, the waveform generated by different simulators will be different as well. What might be more practical is to have the simulator print value out and do the comparison that way. 

The simplest thing you can do is keep the counter's interface and replace the internal logic with what you want to experiment with. The current waveform you are getting also contains the bitstream loading action. 

Based on the type of question you are asking, you have no previous experience in RTL design. If you are planning to use FABulous for your research, email me at king.chung@manchester.ac.uk, and we can see how to better assist you.

IAmMarcelJung:
Hi, I don't know how much you already solved in personal communication, but I'll just leave my two cents here:

> I am trying to compare the waveforms of the test circuit [...] with the waveform of the ouput of the eFPGA after simulation [...].

[...]

PR #217, opened 2024-07-30
Fix .fst generation in run_simulation and correct signal generation for the demo design in the test bench

IAmMarcelJung:
As mentioned in #213, the generation of `.fst` files is currently broken, since the `-fst` flag is missing when calling `vvp`. 
This is fixed here. The second fix concerns the test bench `sequential_16bit_en_tb.v`, where the enable bit was not set for the user design, resulting in undefined signals. Since the simulation is meant for the fabric, it did still not fail, but the output values of the DUT were useless. 

The fix for the `.fst` file was not applied to `run_simulation.sh` since no waveform will be generated by default in this file.
In the future `sequential_16bit_en_tb.v` could be made more generic, but for now this is just a quick fix with the minimum required changes, so that at least the demo design is correctly simulated.

### DIFF
--- a/FABulous.py
+++ b/FABulous.py
@@ -1079,7 +1079,10 @@ To run the complete FABulous flow with the default project, run the following co
         make_hex(f"{self.projectDir}/{path}/{bitstream}", f"{self.projectDir}/{path}/{bitstream_hex}")
 
         try:
-            runCmd = ["vvp", f"{self.projectDir}/{path}/{vvp_file}"]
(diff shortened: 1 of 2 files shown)


KelvinChung2000:
We should use the simulation command for testing instead of calling a script. But before we do that, we need to find a way to propagate the error to the top instead of letting the program die gracefully—something like a test mode for the shell. But that's a problem for another day.

PR #221, opened 2024-08-02
drop vpr support and clean up

KelvinChung2000:
Maintaining a toolchain is very difficult, and maintaining two is not something viable. Since we do not have the resources to do so, it is better to give up supporting it. 

Please let me know what else will need to be removed.


FABulous/FABulous_API.py
```diff
@@ -16,7 +15,7 @@ class FABulous:
     """Class for managing fabric and geometry generation.
 
     This class parses fabric data from 'fabric.csv', generates fabric layouts,
-    geometries, models for both place and route tools (npnr/VPR), as well as
+    geometries, models for place and route tool, as well as
```
TaoBi22: Nit: might be clearer to just say nextpnr if we're only supporting one

FABulous/FABulous_API.py
```diff
@@ -203,7 +202,7 @@ def genBitStreamSpec(self):
         specObject = self.fabricGenerator.generateBitsStreamSpec()
         return specObject
 
-    def genModelNpnr(self):
+    def genModel(self):
```
TaoBi22: I still think it's worth having a function name that clarifies this is a model for nextpnr, just `genModel` is pretty vague

docs/source/FPGA-to-bitstream/Yosys compilation.rst
```diff
@@ -23,10 +23,8 @@ calling the following command:
 .. code-block:: console
 
         # Nextpnr backend synthesis (JSON)
-        synthesis_npnr <path_to_user_design>
+        synthesis <path_to_user_design>
```
TaoBi22, line 26: Is the change to the CLI code also made in this PR? Couldn't see it at a quick skim, sorry if I've missed it

PR #222, opened 2024-08-06
Vector Support for User BEL in Verilog.

A-Kibats:
Currently in the verilog files containing user BEL, ports must be declared individually, an example of this would be as follows: 
```
module RegFile_32x4 (D0, D1, D2, D3, W_ADR0, W_ADR1, W_ADR2, W_ADR3, W_ADR4, W_en, AD0, AD1, AD2, AD3, A_ADR0, A_ADR1, A_ADR2, A_ADR3, A_ADR4, BD0, BD1, BD2, BD3, B_ADR0, B_ADR1, B_ADR2, B_ADR3, B_ADR4, UserCLK, ConfigBits);
	parameter NoConfigBits = 2;// has to be adjusted manually (we don't use an arithmetic parser for the value)
	// IMPORTANT: this has to be in a dedicated line
	input D0; // Register File write port
	input D1;
	input D2;
	input D3;
	input W_ADR0;
	input W_ADR1;
	input W_ADR2;
	input W_ADR3;
	input W_ADR4;
	input W_en;
	
	output AD0;// Register File read port A
	output AD1;
	output AD2;
	output AD3;
	input A_ADR0;
	input A_ADR1;
	input A_ADR2;
	input A_ADR3;
	input A_ADR4;

	output BD0;//Register File read port B
	output BD1;
	output BD2;
	output BD3;
	input B_ADR0;
	input B_ADR1;
	input B_ADR2;
	input B_ADR3;
	input B_ADR4;

	(* FABulous, EXTERNAL, SHARED_PORT *) input UserCLK;// EXTERNAL // SHARED_PORT // ## the EXTERNAL keyword will send this sisgnal all the way to top and the //SHARED Allows multiple BELs using the same port (e.g. for exporting a clock to the top)
	// GLOBAL all primitive pins that are connected to the switch matrix have to go before the GLOBAL label
	(* FABulous, GLOBAL *) input [NoConfigBits-1:0] ConfigBits;
```

With these modifications  input can now be as follows:

```
module RegFile_32x4 #(parameter NoConfigBits = 2)(
	// ConfigBits has to be adjusted manually (we don't use an arithmetic parser for the value)
	input [3:0] D, // Register File write port
	input [4:0] W_ADR,
	input W_en,
	
	output [3:0] AD, // Register File read port A
	input [4:0] A_ADR,

	output [3:0] BD, //Register File read port B
	input [4:0] B_ADR,

	(* FABulous, EXTERNAL, SHARED_PORT *) input UserCLK, // EXTERNAL // SHARED_PORT // ## the EXTERNAL keyword will send this sisgnal all the way to top and the //SHARED Allows multiple BELs using the same port (e.g. for exporting a clock to the top)
	// GLOBAL all primitive pins that are connected to the switch matrix have to go before the GLOBAL label
	(* FABulous, GLOBAL *) input [NoConfigBits-1:0] ConfigBits
);
```

This is primarily done by using yosys to convert the verilog into a JSON netlist and parsing the JSON instead. This also has the added benefits of passing off some of the error handling to yosys and potentially being able to process VHDL files using the same logic in the future.

Backwards compatibility has been added for the individual deceleration of ports and as long as files that rely on each other have the same format, all should be well. However this does not include mixing styles within the same file.

Current Verilog template tiles have been changed to vector to reflect these changes.

### Testing

The following scenarios were tested using the demo tiles:

- All tiles kept as original. | **Passed**
- All tiles converted to use vectors. | **Passed**
- Mixed formats (MULADD.v & LUT4c_frame_config_dffesr.v were kept original, rest vector). | **Passed**
- Mixed formats (RAM_IO/config_access.v & W_IO/config_access.v had different formats). | **Failed**

Please review the changes and let me know if there are any questions or further modifications needed. Thank you!

### DIFF
--- a/FABulous/fabric_files/FABulous_project_template_verilog/Tile/RAM_IO/Config_access.v
+++ b/FABulous/fabric_files/FABulous_project_template_verilog/Tile/RAM_IO/Config_access.v
@@ -7,14 +7,11 @@ C_bit3=3
 module Config_access #(parameter NoConfigBits = 4)(
 	// ConfigBits has to be adjusted manually (we don't use an arithmetic parser for the value)
 	// Pin0
-	(* FABulous, EXTERNAL *)output [3:0] C, // EXTERNAL
+	(* FABulous, EXTERNAL *)output [3:0] C_bit, // EXTERNAL
 	// GLOBAL all primitive pins that are connected to the switch matrix have to go before the GLOBAL label
 	(* FABulous, GLOBAL *)input [NoConfigBits-1:0] ConfigBits
 );
 	// we just wire configuration bits to fabric top
-	assign C[0] = ConfigBits[0];
-    assign C[1] = ConfigBits[1];
-    assign C[2] = ConfigBits[2];
-    assign C[3] = ConfigBits[3];
+	assign C_bit = ConfigBits;
 
 endmodule
--- a/FABulous/fabric_files/FABulous_project_template_verilog/Tile/RAM_IO/Config_access.v
+++ b/FABulous/fabric_files/FABulous_project_template_verilog/Tile/RAM_IO/Config_access.v
@@ -4,20 +4,17 @@ C_bit1=1,
 C_bit2=2,
 C_bit3=3
 *)
-module Config_access (C_bit0, C_bit1, C_bit2, C_bit3, ConfigBits);
-	parameter NoConfigBits = 4;// has to be adjusted manually (we don't use an arithmetic parser for the value)
+module Config_access #(parameter NoConfigBits = 4)(
+	// ConfigBits has to be adjusted manually (we don't use an arithmetic parser for the value)
 	// Pin0
-	(* FABulous, EXTERNAL *)output C_bit0; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit1; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit2; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit3; // EXTERNAL
+	(* FABulous, EXTERNAL *)output [3:0] C, // EXTERNAL
 	// GLOBAL all primitive pins that are connected to the switch matrix have to go before the GLOBAL label
-	(* FABulous, GLOBAL *)input [NoConfigBits-1:0] ConfigBits;
-
+	(* FABulous, GLOBAL *)input [NoConfigBits-1:0] ConfigBits
+);
 	// we just wire configuration bits to fabric top
-	assign C_bit0 = ConfigBits[0];
-	assign C_bit1 = ConfigBits[1];
-	assign C_bit2 = ConfigBits[2];
-	assign C_bit3 = ConfigBits[3];
+	assign C[0] = ConfigBits[0];
+    assign C[1] = ConfigBits[1];
+    assign C[2] = ConfigBits[2];
+    assign C[3] = ConfigBits[3];
 
 endmodule
--- a/FABulous/fabric_files/FABulous_project_template_verilog/Tile/W_IO/Config_access.v
+++ b/FABulous/fabric_files/FABulous_project_template_verilog/Tile/W_IO/Config_access.v
@@ -7,14 +7,11 @@ C_bit3=3
 module Config_access #(parameter NoConfigBits = 4)(
 	// ConfigBits has to be adjusted manually (we don't use an arithmetic parser for the value)
 	// Pin0
-	(* FABulous, EXTERNAL *)output [3:0] C, // EXTERNAL
+	(* FABulous, EXTERNAL *)output [3:0] C_bit, // EXTERNAL
 	// GLOBAL all primitive pins that are connected to the switch matrix have to go before the GLOBAL label
 	(* FABulous, GLOBAL *)input [NoConfigBits-1:0] ConfigBits
 );
 	// we just wire configuration bits to fabric top
-	assign C[0] = ConfigBits[0];
-    assign C[1] = ConfigBits[1];
-    assign C[2] = ConfigBits[2];
-    assign C[3] = ConfigBits[3];
+	assign C_bit = ConfigBits;
 
 endmodule
--- a/FABulous/fabric_files/FABulous_project_template_verilog/Tile/W_IO/Config_access.v
+++ b/FABulous/fabric_files/FABulous_project_template_verilog/Tile/W_IO/Config_access.v
@@ -4,20 +4,17 @@ C_bit1=1,
 C_bit2=2,
 C_bit3=3
 *)
-module Config_access (C_bit0, C_bit1, C_bit2, C_bit3, ConfigBits);
-	parameter NoConfigBits = 4;// has to be adjusted manually (we don't use an arithmetic parser for the value)
+module Config_access #(parameter NoConfigBits = 4)(
+	// ConfigBits has to be adjusted manually (we don't use an arithmetic parser for the value)
 	// Pin0
-	(* FABulous, EXTERNAL *)output C_bit0; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit1; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit2; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit3; // EXTERNAL
+	(* FABulous, EXTERNAL *)output [3:0] C, // EXTERNAL
 	// GLOBAL all primitive pins that are connected to the switch matrix have to go before the GLOBAL label
-	(* FABulous, GLOBAL *)input [NoConfigBits-1:0] ConfigBits;
-
+	(* FABulous, GLOBAL *)input [NoConfigBits-1:0] ConfigBits
+);
 	// we just wire configuration bits to fabric top
-	assign C_bit0 = ConfigBits[0];
-	assign C_bit1 = ConfigBits[1];
-	assign C_bit2 = ConfigBits[2];
-	assign C_bit3 = ConfigBits[3];
+	assign C[0] = ConfigBits[0];
+    assign C[1] = ConfigBits[1];
+    assign C[2] = ConfigBits[2];
+    assign C[3] = ConfigBits[3];
 
 endmodule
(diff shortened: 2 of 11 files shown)


FABulous/fabric_files/FABulous_project_template_verilog/Tile/RAM_IO/Config_access.v
```diff
@@ -4,20 +4,14 @@ C_bit1=1,
...
-module Config_access (C_bit0, C_bit1, C_bit2, C_bit3, ConfigBits);
-	parameter NoConfigBits = 4;// has to be adjusted manually (we don't use an arithmetic parser for the value)
+module Config_access #(parameter NoConfigBits = 4)(
+	// ConfigBits has to be adjusted manually (we don't use an arithmetic parser for the value)
 	// Pin0
-	(* FABulous, EXTERNAL *)output C_bit0; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit1; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit2; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit3; // EXTERNAL
+	(* FABulous, EXTERNAL *)output [3:0] C_bit, // EXTERNAL
```
KelvinChung2000, line 10: name `C_bit` to `C_bits` will be better

FABulous/fabric_files/FABulous_project_template_verilog/Tile/W_IO/Config_access.v
```diff
@@ -4,20 +4,14 @@ C_bit1=1,
...
-module Config_access (C_bit0, C_bit1, C_bit2, C_bit3, ConfigBits);
-	parameter NoConfigBits = 4;// has to be adjusted manually (we don't use an arithmetic parser for the value)
+module Config_access #(parameter NoConfigBits = 4)(
+	// ConfigBits has to be adjusted manually (we don't use an arithmetic parser for the value)
 	// Pin0
-	(* FABulous, EXTERNAL *)output C_bit0; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit1; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit2; // EXTERNAL
-	(* FABulous, EXTERNAL *)output C_bit3; // EXTERNAL
+	(* FABulous, EXTERNAL *)output [3:0] C_bit, // EXTERNAL
```
KelvinChung2000, line 10: same in here

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -1085,23 +1086,61 @@ def generateTile(self, tile: Tile) -> None:
...
-            # internal port
+            # Internal ports
             for port in bel.inputs + bel.outputs:
-                port = port.removeprefix(bel.prefix)
-                portsPairs.append((port, f"{bel.prefix}{port}"))
-
-            # external port
+                port_name = port.removeprefix(bel.prefix)
+                match = re.match(r"([a-zA-Z_]+)(\d*)", port_name)
+                if match:
```
KelvinChung2000: a more compact way to write will be
```
if r := re.match(r"([a-zA-Z_]+)(\d*)", port_name):
  portname, number = r.groups()
  port_dict[portname].append((port, number)) 
```

FABulous/fabric_generator/file_parser.py
```diff
@@ -1,10 +1,12 @@
 import csv
 import os
 import re
+import subprocess
+import json
 from loguru import logger
 from copy import deepcopy
 
-from typing import Literal
+from typing import Literal, Dict
```
KelvinChung2000: will not need `Dict` typing, simply using `dict` will work

FABulous/fabric_generator/file_parser.py
```diff
@@ -579,45 +582,30 @@ def parseFile(filename: str, belPrefix: str = "", filetype: str = "") -> Bel:
         exit(-1)
 
     if filetype not in ("verilog", "vhdl"):
```
KelvinChung2000: You can add a `Literial` type at the function argument, which make the function more readable

FABulous/fabric_generator/file_parser.py
```diff
@@ -664,6 +653,82 @@ def parseFile(filename: str, belPrefix: str = "", filetype: str = "") -> Bel:
...
+                    continue
+                if "UserCLK" in port_name:
+                    userClk = True
+                if port_name[-1].isdigit():
+                    individually_declared = True
+                direction = IO[details["direction"].upper()]
+                bits = details.get("bits", [])
+                filtered_ports[port_name] = (direction, bits)
+        # Passed attributes dont show in port list, checks for attributes in other lists.
+        for submodule in sub_modules:
```
KelvinChung2000: I might be wrong, but I notice all the attributes are on the `netnames`

FABulous/fabric_generator/file_parser.py
```diff
@@ -664,6 +653,82 @@ def parseFile(filename: str, belPrefix: str = "", filetype: str = "") -> Bel:
...
+            if submodule not in module_info:
+                continue
```
KelvinChung2000: Since the sub part is defined, this is not needed

FABulous/fabric_generator/file_parser.py
```diff
@@ -664,6 +653,82 @@ def parseFile(filename: str, belPrefix: str = "", filetype: str = "") -> Bel:
...
+            item_info = module_info.get(submodule, {})
+            for item, details in item_info.items():
+                if item == "NoConfigBits":
+                    noConfigBits = details
+                if item in filtered_ports:
+                    direction, bits = filtered_ports[item]
+                    attributes = details.get("attributes", {})
+                    for index in range(len(bits)):
+                        if len(bits) > 1:
+                            new_port_name = (
```
KelvinChung2000: this can stay on the same line

FABulous/fabric_generator/file_parser.py
```diff
@@ -664,6 +653,82 @@ def parseFile(filename: str, belPrefix: str = "", filetype: str = "") -> Bel:
...
+                                f"{item}{index}"  # Multi-bit ports get index
+                            )
+                        else:
+                            new_port_name = (
+                                item  # Single-bit port uses the original name
```
KelvinChung2000: same here

FABulous/fabric_generator/file_parser.py
```diff
@@ -664,6 +653,82 @@ def parseFile(filename: str, belPrefix: str = "", filetype: str = "") -> Bel:
...
+                                internal.append(
+                                    (f"{belPrefix}{new_port_name}", direction)
+                                )
+        belMapDic = verilog_belMapProcessing(module_info)
+        if len(belMapDic) != noConfigBits:
+            raise ValueError(
+                f"NoConfigBits does not match with the BEL map in file {filename}, length of BelMap is {len(belMapDic)}, but with {noConfigBits} config bits"
+            )
+
     return Bel(
```
KelvinChung2000, line 715: The return here is better done with a name argument, like `varA=A`

FABulous/fabric_generator/file_parser.py
```diff
@@ -674,12 +739,61 @@ def parseFile(filename: str, belPrefix: str = "", filetype: str = "") -> Bel:
...
+
+        belMapDic[new_key] = {0: {0: "1"}}
+
+    # yosys reverses belmap, reverse back to keep original belmap.
+    belMapDic = dict(reversed(list(belMapDic.items())))
+
+    return belMapDic
+
+
+def vhdl_belMapProcessing(file: str, filename: str, syntax: Literal["vhdl"]) -> dict:
```
KelvinChung2000: If already know is VHDL then the syntax part is not needed

FABulous/fabric_generator/file_parser.py
```diff
@@ -664,6 +653,82 @@ def parseFile(filename: str, belPrefix: str = "", filetype: str = "") -> Bel:
...
+                    for index in range(len(bits)):
+                        if len(bits) > 1:
+                            new_port_name = (
+                                f"{item}{index}"  # Multi-bit ports get index
+                            )
+                        else:
+                            new_port_name = (
+                                item  # Single-bit port uses the original name
+                            )
+                        match attributes:
```
KelvinChung2000:
```
for attr in attributes:
  match attr:
    case "EXTERNAL":
      ....
```
will look better

FABulous/fabric_generator/file_parser.py
A-Kibats: My understanding from yosys documentation for [write_json](https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/write_json.html) was that `cells` and `memories` also contain attribute details. I'm not sure if ports and declarations always end up in `netnames` but I thought it would be best to be thorough in-case through optimisation yosys does put them else where.

KelvinChung2000: Yes, it is also better to check, but for example, if `cells` will never contain the FABulous related attributes, then we will never need to check them.

FABulous/fabric_files/FABulous_project_template_verilog/Tile/RAM_IO/Config_access.v
A-Kibats: changing to `C_bits` causes issues with `generateTopWrapper()` in fabric_gen.py:

```
            # some backwards compat
            if basename.endswith("_bit"):
                # _bit is part of the indexing rather than the name
                basename = basename[:-4]
            # top level IO has A and B parts combined and reverse order
            if len(basename) == 7 and basename[1:] in ("_I_top", "_O_top", "_T_top"):
                assert basename[0] in "ABCDEFGH"
                indices.append(-(ord(basename[0]) - ord("A")))
                basename = basename[2:]
```
If you still want me to change it and make this compatible let me know.

FABulous/fabric_generator/file_parser.py
A-Kibats: I've streamlined it now to just check netnames, i'll add a note though if it does cause issues in future.

FABulous/fabric_generator/file_parser.py
A-Kibats: Trying this way causes issues, see more detail in teams.

FABulous/fabric_files/FABulous_project_template_verilog/Tile/RAM_IO/Config_access.v
KelvinChung2000: Keep it the way as is then

FABulous/fabric_generator/file_parser.py
A-Kibats: Change now to if else for clarity.

PR #223, opened 2024-08-08
Warning For Individual Port Declaration

A-Kibats:
PR adds warning for Individual Port Declaration as suggested by @KelvinChung2000.

Example:

image

-Edit- 

Message is now `...Individually declared rather than as a vector.`

### DIFF
--- a/FABulous/fabric_generator/file_parser.py
+++ b/FABulous/fabric_generator/file_parser.py
@@ -667,10 +667,10 @@ def parseFile(
 
         with open(f"{json_file}", "r") as f:
             data_dict = json.load(f)
-        # Default yosys list names added.
+
         modules = data_dict.get("modules", {})
         filtered_ports: dict[str, IO] = {}
-        # Gatheres port name and direction, filters out configbits as they show in ports.
+        # Gathers port name and direction, filters out configbits as they show in ports.
         for module_name, module_info in modules.items():
             ports = module_info["ports"]
             for port_name, details in ports.items():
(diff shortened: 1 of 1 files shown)


FABulous/fabric_generator/file_parser.py
```diff
@@ -667,7 +667,7 @@ def parseFile(
 
         with open(f"{json_file}", "r") as f:
             data_dict = json.load(f)
-        # Default yosys list names added.
+
         modules = data_dict.get("modules", {})
         filtered_ports: dict[str, IO] = {}
         # Gatheres port name and direction, filters out configbits as they show in ports.
```
IAmMarcelJung:
```suggestion
        # Gathers port name and direction, filters out configbits as they show in ports.
```

FABulous/fabric_generator/file_parser.py
```diff
@@ -684,6 +684,12 @@ def parseFile(
                 bits = details.get("bits", [])
                 filtered_ports[port_name] = (direction, bits)
 
+        if individually_declared:
+            logger.warning(
+                f"Ports in {filename} have been individually declared rather than vector."
```
IAmMarcelJung:
```suggestion
                f"Ports in {filename} have been individually declared rather than as a vector."
```
Reads a bit nicer IMHO.

IAmMarcelJung (changes requested):
Thanks for this! I just have these two small remarks.

A-Kibats:
Added the suggestions and squashed commit. Cheers.

PR #225, opened 2024-08-09
Add new print feature

KelvinChung2000:
Allow users to print the Bel and Tile information in the loaded fabric.

Some extra clean up.


IAmMarcelJung:
A lot of docstrings where removed. Was this on done on purpose?

IAmMarcelJung:
Okay then it's fine, just wanted to be sure :)

KelvinChung2000:
If the rest looks good to you, I will merge this in

IAmMarcelJung:
Yes I think it looks good, also tried it out and it worked fine!

PR #229, opened 2024-08-20
Update python to 3.12 and update documentation

### LINKED ISSUE #228: Using python3.10 feature but python3.9 is specified in README.md (FABulous2.0-development branch)
When I wanted to test the changes in #227, I saw that the python version to use is specified as `python3.9` in the `FABulous2.0-development`  branch. Therefore I set my venv to `python3.9` and set it up as usual. When I wanted to create the project, using `FABulous -c demo` I got the following error (shortened):

```console
File "FABulous/fabric_definition/Fabric.py", line 249, in Fabric
    def getTileByName(self, name: str) -> Tile | None:
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```
I then found out that the `union` type was introduced in [PEP 604](https://peps.python.org/pep-0604/) and is only supported in `python3.10` and higher. Therefore I suggest we either remove the union type or change the suggested python version in `README.md` to at least `python3.10` in the `FABulous2.0-development` branch (which should be preferred IMHO).

If there is an error on my side please let me know, but I think the PEP is quite clear.

IAmMarcelJung:
This fixes #228 by changing the required python version to 3.12. Since `Usage.rst` had to be updated, small improvements were also made to that file which are listed below:

- The note about the fasm warning was hidden and is now visible.
- The subsection headings for the docker section were wrapped into two lines and are now changed to just bold text. They could have also be changed to an actual heading, but the next heading level was too big compared to the previous solution using fields and the level seem to be managed automatically in restructredText. If anyone knows a better solution suggestion are appreciated!
- Adding a note about possible issues with tkinter and virtualenv (I ran into this, since Ubuntu 22.04 uses 3.10 by default, so others might also run into this). I also added this to `README.md` (thanks to @EverythingElseWasAlreadyTaken for the hint with turning it into a note!).
- Small typo and layout fixes.

I hope I did catch all spots where the python version needs to be updated. I kindly ask @EverythingElseWasAlreadyTaken to also look over this since he has done a lot of the python setup and might know where to look at. Thanks in advance to any reviewer!

And as always I'm open to any kind of feedback!


docs/source/Usage.rst
```diff
@@ -11,14 +11,24 @@ The following packages need to be installed for generating fabric HDLs
 
 :Python:
 
-Version >= 3.9
+Version >= 3.12
```
KelvinChung2000, line 14: Even though we do everything in Python 3.12, the minimum requirement is Python 3.10. Should we say something like the minimum requirement is 3.10, but recommend using Python3.12? Or should we support all 3.10-3.12?

pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "FABulous-FPGA"
 authors = [{name = "Jing, Nguyen, Bea, Bardia, Dirk", email = "dirk.koch@manchester.ac.uk"}]
 description = "FABulous FPGA Fabric generator"
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.12"
```
KelvinChung2000, line 10: I think we also do the version assertion [in the shell](https://github.com/FPGA-Research-Manchester/FABulous/blob/FABulous2.0-development/FABulous/FABulous.py#L1499)

docs/source/Usage.rst
```diff
@@ -29,13 +39,13 @@ Version >= 3.9
...
 If you are not sure what this is and why you should use it, please read the
-:`virtualenv documentation <https://virtualenv.pypa.io/en/latest/index.html>`_:
+`virtualenv documentation <https://virtualenv.pypa.io/en/latest/index.html>`_.
 
 .. code-block:: console
 
    $ cd FABulous
-    $ virtualenv venv
-    $ source venv/bin/activate
+   $ virtualenv venv
```
KelvinChung2000, line 47: Does the copy code block also copy the `$`? Sometimes, it is annoying when copying a command and getting an invalid command because of the `$` sign.  
Some references might help solve the problem: 
https://github.com/executablebooks/sphinx-book-theme/issues/411
https://sphinx-copybutton.readthedocs.io/en/latest/use.html#using-regexp-prompt-identifiers

docs/source/Usage.rst
IAmMarcelJung: I think we should specifiy the version that we use for the GitHub actions (GHA) since this version is guaranteed to be working. We could say something like "older versions might also work but are not officially supported" or we could also explcitely say ">= 3.10 might also work but is not officially supported". The alternative would be keep using 3.10 in the GHA and also change all docs and checks to 3.10 since we cirrently don't use features of >=3.11 anyway.

pyproject.toml
IAmMarcelJung: Thanks, I will update this as soon as we agree on a minimum version!

docs/source/Usage.rst
IAmMarcelJung: Until now we don't use the copybutton feature. Should I add it, along with the suggested configuration? I tried it out and it worked perfectly. Using the copybutton without the configuration does indeed also copy the `$` as you suggested. I think it would be a useful feature to have a copy button, but then we would also have to adjust the config so that `(venv)` and `(FABulous>)` are also ignored.

KelvinChung2000: You can decide including it in this pull request or leave it later. The adjustment should be simple, is just some extra regex need to match.

IAmMarcelJung: I just added  (not yet commited) `|FABulous\> |\(venv\)\$` to the configuration to also ignore `(venv)` and `FABulous>` (but not `FABulous`!). Maybe there's a better/smarter way, but it worked. 

One thing that didn't work was ignoring comments starting with `#` which we have e.g. [here](https://github.com/FPGA-Research-Manchester/FABulous/blob/FABulous2.0-development/docs/source/FPGA-to-bitstream/Yosys%20compilation.rst?plain=1#L25). I tried to add `|^\#.*` which does work on https://regex101.com/ but in the actual documentation I had no success.

docs/source/Usage.rst
KelvinChung2000: Feature like fstring improvement and parametric typing might be useful later. 

How about we do official 3.12, so CI also on 3.12, Then saying 3.10, 3.11 is unofficial supported, but also should work. In the toml we do requirements as >=3.12. We should also remove the version check in the shell since the install will handle version checks.

IAmMarcelJung: Sounds good to me, I will adjust it tomorrow!

docs/source/Usage.rst
KelvinChung2000: Maybe try doing as raw string. 

The `^` means start of the line. The line it try to match might have space, which make it not start with `#`, I think without the `^` should work.

IAmMarcelJung: > Maybe try doing as raw string.

You mean with r""? I already did that, since they already suggest it in the sphinx docs.

 After trying out a lot of different stuff, I found out that the comment in `Usage.rst` is automatically ignored without explicitely mentioning it in the regex. So I just added the 'FABulous> ` prefix to the other comment and then it also worked there. I will push the changes later.

pyproject.toml
IAmMarcelJung: Like [this](https://github.com/FPGA-Research-Manchester/FABulous/pull/229/commits/34c3b9c6a0d51cd8245adbdaa0f5c70906b20257#diff-965514c6ec039ffc7c9ead6c416589047a46ce6bf1eff968ba825543e25cde0eR1499-R1500), right?

KelvinChung2000: We can remove the check. If the user uses the pip install method, the install should ensure it runs at python3.12. We will also have fewer things to maintain.

IAmMarcelJung: You mean removing the check in the shell?

KelvinChung2000: Yes. With the installation, we should not need to manage the version within the shell anymore.

IAmMarcelJung:
> We should do an install script of the OSS cad suit or even install it via the setup tools — future problems.

Yes this would be the preferred approach to  make the setup even more straightforward.

Thanks for reviewing and merging! :)

PR #232, opened 2024-09-10
Add better handling for environment variables

EverythingElseWasAlreadyTaken:
Add functions to handle global and project specific environment variables and .env files.
Add commandline parameter to pass global or project specific .env files.
Add docs for .env. 

Remove unnecessary load_writer_from_env function.

### DIFF
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -44,8 +44,8 @@ from FABulous.FABulous_API import FABulous
 readline.set_completer_delims(" \t\n")
 histfile = ""
 histfile_size = 1000
-
 MAX_BITBYTES = 16384
+metaDataDir = ".FABulous"
 
 
 def setup_logger(verbosity: int):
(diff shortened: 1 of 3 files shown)


KelvinChung2000 (commented):
Can you please change the `os.path` with `pathlib`. 

Would using the `dotenv_values` better? Since we can avoid touching the users' OS.

For the global setting, would it be better to do it via setuptools?

EverythingElseWasAlreadyTaken:
> I am thinking of generating the global env from the setup tool, which will be the default, then loading in user env and overwriting things when needed.

Ok, this can still be done. All used env vars have defaults, so no known FABulous behavior is broken. 
Maybe setup tools could just generate a .env file, which could still be changed.

KelvinChung2000:
Forgot default value exists🤦🏻Nothing needs worrying then.

Kind regards,
Kelvin Chung

FABulous/FABulous.py
```diff
@@ -67,24 +67,86 @@ def setup_logger(verbosity: int):
...
+        else:
+            logger.info(f"No global .env file found at {gde}")
+    elif (
+        Path(os.getenv("FAB_ROOT")).joinpath(".env").exists()
+        and Path(os.getenv("FAB_ROOT")).joinpath(".env").is_file()
+    ):
+        load_dotenv(Path(os.getenv("FAB_ROOT")).joinpath(".env"))
+        logger.info(f"Loaded global .env file from {fabulousRoot}/.env")
+    else:
+        logger.info("No global .env file found")
```
KelvinChung2000: Should terminate with error logging since reaching this point means something about the installation is wrong

FABulous/FABulous.py
```diff
@@ -67,24 +67,86 @@ def setup_logger(verbosity: int):
...
+    if args.globalDotEnv:
+        gde = Path(args.globalDotEnv)
+        if gde.is_file():
+            load_dotenv(gde)
+            logger.info(f"Loaded global .env file from {gde}")
+        elif gde.joinpath(".env").exists() and gde.joinpath(".env").is_file():
+            load_dotenv(gde.joinpath(".env"))
+            logger.info(f"Loaded global .env file from {gde.joinpath('.env')}")
+        else:
+            logger.info(f"No global .env file found at {gde}")
```
KelvinChung2000: Should terminate with an error when reaching this point

FABulous/FABulous.py
```diff
@@ -67,24 +67,86 @@ def setup_logger(verbosity: int):
...
+    if args.projectDotEnv:
+        pde = Path(args.projectDotEnv)
+        if pde.exists() and pde.is_file():
+            load_dotenv(pde)
+            logger.info(f"Loaded global .env file from pde")
+    elif fabDir.joinpath(".env").exists() and fabDir.joinpath(".env").is_file():
+        load_dotenv(fabDir.joinpath(".env"))
+        logger.info(f"Loaded project .env file from {fabDir}/.env')")
+    else:
+        logger.info("No project .env file found")
```
KelvinChung2000: Same in here.

KelvinChung2000 (approved):
One last final thing, once this is sorted, we can get this merged.

FABulous/FABulous.py
EverythingElseWasAlreadyTaken: I would maybe make this a warning, since you could also use FABulous it entirely without a .env file, since we have Default values for everything, and you can set everything through environment variables in your terminal.

KelvinChung2000: That will do as well

EverythingElseWasAlreadyTaken: Done :)

PR #233, opened 2024-09-18
Fix backward compatibility to old project structure and some housekeeping 

EverythingElseWasAlreadyTaken:
Fix backward compatibility to old project structure: 
- Add case for old tcl scripts in run_FABulous_bitstream.
- ConfigMemPath in fabric_gen:generateBitstreamSpec was set wrong, if a project used the old fabric.csv structure.

Remove unneeded get_path() function and use Path() instead of pathlib.Path()

Add check if external programs are installed

### DIFF
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -1509,11 +1509,26 @@ To run the complete FABulous flow with the default project, run the following co
             run_FABulous_bitstream <top_module_file>
 
         """
-        if len(args) != 1:
+        if len(args) == 1:
+            verilog_file_path = get_path(args[0])
+        elif len(args) == 2:
+            # Backwards compatibility to older scripts
+            if "npnr" in args[0]:
+                verilog_file_path = get_path(args[1])
+            elif "vpr" in args[0]:
+                logger.error(
+                    "run_FABulous_bitstream does not support vpr anymore, please use npnr or try a older FABulous version."
+                )
+                return
+
+            else:
+                logger.error(f"run_FABulous_bitstream does not support {args[0]}")
+                return
+
+        else:
             logger.error("Usage: run_FABulous_bitstream <top_module_file>")
             return
 
-        verilog_file_path = get_path(args[0])
         file_path_no_suffix = verilog_file_path.parent / verilog_file_path.stem
 
         if verilog_file_path.suffix != ".v":
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -31,7 +31,7 @@ import tkinter as tk
 import traceback
 from contextlib import redirect_stdout
 from glob import glob
-from pathlib import PurePosixPath, PureWindowsPath, Path
+from pathlib import PurePath, Path
 from typing import List, Literal
 from dotenv import load_dotenv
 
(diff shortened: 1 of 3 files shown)


FABulous/FABulous.py
```diff
@@ -1509,11 +1519,26 @@ def do_run_FABulous_bitstream(self, *args):
...
-        if len(args) != 1:
+        if len(args) == 1:
+            verilog_file_path = PurePath(args[0])
+        elif len(args) == 2:
+            # Backwards compatibility to older scripts
+            if "npnr" in args[0]:
+                verilog_file_path = PurePath(args[1])
+            elif "vpr" in args[0]:
+                logger.error(
+                    "run_FABulous_bitstream does not support vpr anymore, please use npnr or try a older FABulous version."
```
IAmMarcelJung:
```suggestion
                    "run_FABulous_bitstream does not support vpr anymore, please use npnr or try an older FABulous version."
```

FABulous/FABulous.py
```diff
@@ -252,6 +252,37 @@ def make_hex(binfile, outfile):
...
+    """
+    path = shutil.which(application)
+    if path is not None:
+        return Path(path)
+    else:
+        logger.error(
+            f"{application} is not installed. Please install it or set FAB_<APPLICATION>_PATH in the .env file."
+        )
+        if throw_exeption:
+            raise Exception(f"{application} is not installed.")
```
IAmMarcelJung:
```suggestion
def check_if_application_exists(application: str, throw_exception: bool = True) -> Path:
    """Checks if an application is installed on the system.

    Parameters
    ----------
    application : str
        Name of the application to check.
    throw_exception : bool, optional
        If True, throws an exception if the application is not installed, by default True

    Returns
    -------
    Path
        Path to the application, if installed.

    Raises
    ------
    Exception
        If the application is not installed and throw_exception is True.
    """
    path = shutil.which(application)
    if path is not None:
        return Path(path)
    else:
        logger.error(
            f"{application} is not installed. Please install it or set FAB_<APPLICATION>_PATH in the .env file."
        )
        if throw_exception:
            raise Exception(f"{application} is not installed.")
```

FABulous/fabric_generator/file_parser.py
```diff
@@ -555,7 +551,8 @@ def parseTiles(fileName: pathlib.Path) -> tuple[list[Tile], list[tuple[str, str]
             else:
                 logger.error(f"Unknown tile description {temp[0]} in tile {t}.")
                 raise ValueError
-
+            # TODO: if old fabric.csv tileDir is set wrong
```
IAmMarcelJung: Is the TODO supposed to be still in there to be fixed in the future? Or was it just missed?

IAmMarcelJung (changes requested):
Thanks for this! :) As always I just found some typos.

KelvinChung2000:
I think we can stop supporting the 1.0 file structure format. We can do all the breaking changes in 2.0.

EverythingElseWasAlreadyTaken:
> I think we can stop supporting the 1.0 file structure format. We can do all the breaking changes in 2.0.

Generally I'm okay with this, but then we should find a proper handling for this, like throwing proper Errors for this.
Since currently, with old projects, the ConfigMem generation just stopped working, but the flow just broke at Place and Route step. 
 And this fix was a relatively easy to implement,

EverythingElseWasAlreadyTaken:
@KelvinChung2000 how should we proceed with this?

KelvinChung2000:
We can keep the fix. I think it is all good now.

Issue #236, opened 2024-10-02
Dependency Revaluation


mole99:
Hello! I would like to package FABulous with Nix, but there are two dependencies that complicate things. They are listed in `pyproject.toml` as `fasm` and `docker` (I am looking at `FABulous2.0-development`).

As far as I can see, the [`docker`](https://pypi.org/project/docker/) Python package is never used in FABulous. Maybe the Python library was confused with the Docker engine, which is indeed necessary to build and run the FABulous docker container? If so, could this dependency be removed?

With `fasm` it's a bit more complicated as it is used to generate the bitstream. From what I see `fasm` is imported in two files: `fabric_gen.py` and `bit_gen.py`. There is a note at each import: `# Remove this line if you do not have the fasm library installed and will not be generating a bitstream`. Would it be possible to import `fasm` with a `try / except`? This would make it possible to run FABulous without `fasm` installed and at least generate a fabric.

Another question, does `fabric_gen` actually make use of fasm? I don't see any functions being used. It would be helpful if instead of `from fasm import *` only selected functions could be imported `from fasm import ..., ...` this makes it easier to see which functions are actually used.

On a side note, I also don't see `numpy` being used in FABulous. Since it is another huge dependency, maybe it could be removed too? Lastly, I think `python-dotenv` and `loguru` should be added to `pyproject.toml`.

I would be happy to help with any changes!

KelvinChung2000:
The `docker` dependency is related to something I worked on but did not release, so it is safe to remove. 

I know that `fabric_gen.py` should not be using `fasm` so this should be safe to remove. I think `fasm` is not being used by `bit_gen.py` as well and have consider removing it but I cannot confirm it is not being used. The `from fasm import *` is also from pre-refactor. The `bit_gen.py` file should be removed and changed to fit the new software architecture...   

The `numpy` package is something from pre-refactor, which I don't think is currently being used.  

`python-dotenv` and `loguru` should be added to the `pyproject.toml`.

I have this discussion of migrating towards Nix for better Openlane2 integration; I don't know if this is being worked on at the moment. Ultimately, we aim to support driving the Openlane2 flow from FABulous.

EverythingElseWasAlreadyTaken:
It would also be nice to add required package versions in the `requirements.txt` and `docs/requirements.txt`.

mole99:
@KelvinChung2000 Well, I'm working on the opposite right now :) Driving FABulous from OpenLane 2. 
[...]

PR #239, opened 2024-10-09
Update Dependencies

mole99:
Update the dependencies as discussed in #236.

`numpy` and `docker` have been removed and for the others the minimum required versions have been set.
If `fasm` cannot be imported, a warning is issued that bitstream generation is not supported in that case.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,9 +18,9 @@ classifiers = [
 
 
 dependencies = [
-    'numpy',
-    'fasm',
-    'docker',
+    'python-dotenv>=1.0.0',
+    'loguru>=0.7.0',
+    'fasm>=0.0.2',
 ]
 
 [project.urls]
--- a/FABulous/fabric_cad/bit_gen.py
+++ b/FABulous/fabric_cad/bit_gen.py
@@ -4,8 +4,17 @@ import pickle
 import re
 import sys
 
-from fasm import *  # Remove this line if you do not have the fasm library installed and will not be generating a bitstream
 from loguru import logger
+
+try:
+    from fasm import (
+        parse_fasm_filename,
+        fasm_tuple_to_string,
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -18,9 +18,9 @@ classifiers = [
 
 
 dependencies = [
-    'numpy',
-    'fasm',
-    'docker',
+    'python-dotenv>=1.0.0',
```
EverythingElseWasAlreadyTaken, line 21: Could you please remove the version pinning in pyproject.toml?

Usually, you only pin the versions in the requiremtens.txt, for deployable/standalone applications. In the pyproject.toml you only add unpinned dependencies, that you don't get any dependency versioning troubles when importing the project as a dependency itself.

EverythingElseWasAlreadyTaken (commented):
Looks good, thanks for the PR! :+1: 

I just left a small comment. :)

pyproject.toml
mole99: Hi, I only added a lower boundary as shown in the user guide here: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements
If dependencies have versions lower than that, then FABulous probably won't work anyway.

Nix's buildPythonPackage looks in pyproject.toml for the dependencies. If I remove the version range from there, I wouldn't get a warning if the dependencies don't have the right versions.

EverythingElseWasAlreadyTaken: Sorry, I overlooked that, I wasn't fully awake yet. :sweat_smile: 
As long as it's only the lower limit, it should be fine. :+1:

EverythingElseWasAlreadyTaken (approved):

pyproject.toml
mole99: No worries, thanks for the review 😃

PR #242, opened 2024-10-21
Fixing non power of 2 mux generation 

KelvinChung2000:
Now can correctly generate non power of 2 mux.

```verilog
 //switch matrix multiplexer E6BEG4 MUX-5
assign E6BEG4_input = {A_O,W6END3,WW4END3,W1END2,W1END0};
cus_mux81_buf inst_cus_mux81_buf_E6BEG4 (
    .A0(E6BEG4_input[0]),
    .A1(E6BEG4_input[1]),
    .A2(E6BEG4_input[2]),
    .A3(E6BEG4_input[3]),
    .A4(E6BEG4_input[4]),
    .A5(GND0),
    .A6(GND0),
    .A7(GND0),
    .S0(ConfigBits[76+0]),
    .S0N(ConfigBits_N[76+0]),
    .S1(ConfigBits[76+1]),
    .S1N(ConfigBits_N[76+1]),
    .S2(ConfigBits[76+2]),
    .S2N(ConfigBits_N[76+2]),
    .X(E6BEG4)
);
```

I have also updated the name of the my_mux2 to cus_mux21, to make them follow the same naming conversion for the other muxes.

### DIFF
--- a/FABulous/fabric_files/FABulous_project_template_verilog/Fabric/models_pack.v
+++ b/FABulous/fabric_files/FABulous_project_template_verilog/Fabric/models_pack.v
@@ -192,7 +192,7 @@ module cus_mux41_buf (A0, A1, A2, A3, S0, S0N, S1, S1N, X);
 	assign X =  S1 ? B1 : B0;
 endmodule
 
-module cus_mux21_buf (A0, A1, S, X);
+module cus_mux21 (A0, A1, S, X);
 	input A0;
 	input A1;
 	input S;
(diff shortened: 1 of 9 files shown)


EverythingElseWasAlreadyTaken (changes requested):
I've run some tests and the general output looks good. 
I compared the default Fabric with the current dev branch, also looks good.

I currently try to run an equivalence check of the outputs with yosys, but this seems to run for a long time. But this is more out of curiosity and for future testing.  

But there is one thing, it would be nice if you maybe squash your commits together and provide a meaningful commit message, why this was changed.
For changes like this, it could be good to know on a later point, why this was changed.

KelvinChung2000:
The repo is default to squash and merge when merging which will use the pull request title as the commit message.

PR #253, opened 2024-11-21
Add generative IOs

EverythingElseWasAlreadyTaken:
Adds a GEN_IO primitive to generate IOs in the top level design and
route them trough to the defined tile, where they are connected to
the switchmatrix.

GEN_IOs can be used as following in the fabric.csv:
GEN_IO,Number of Pins,Direction,Prefix

``` fabric.csv::
GEN_IO,2,OUTPUT,A_O,,,,,,,,,,,,,,
GEN_IO,2,INPUT,A_I,,,,,,,,,,,,,,
```
This will generate four IOs, two INPUT (A_I0, A_I1) and two OUTPUT(A_O0
and A_O1), which can be a accessed in the tile through the switchmatrix.
This will also generate four external ports, (A_I0_top, A_I1_top,
A_O0_top, A_O1_top) which are routed to the top level and are connected
to the equivalent tile ports.

### DIFF
--- a/FABulous/fabric_definition/Tile.py
+++ b/FABulous/fabric_definition/Tile.py
@@ -69,6 +69,10 @@ class Tile:
         for b in self.bels:
             self.globalConfigBits += b.configBit
 
+        for gio in self.gen_ios:
+            if gio.configAccess:
+                self.globalConfigBits += gio.configBit
+
     def __eq__(self, __o: Any) -> bool:
         if __o is None or not isinstance(__o, Tile):
             return False
--- a/FABulous/fabric_definition/Tile.py
+++ b/FABulous/fabric_definition/Tile.py
@@ -22,7 +22,7 @@ class Tile:
         The list of ports of the tile
     matrixDir : str
         The directory of the tile matrix
-    gen_ios : (List[Gen_IO])
+    gen_ios : List[Gen_IO]
         The list of GEN_IOs of the tile
     globalConfigBits : int
         The number of config bits the tile has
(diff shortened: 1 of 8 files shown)


EverythingElseWasAlreadyTaken:
@KelvinChung2000 Please no squash merge on this, since there are a lot of change, so it would be good to keep the commits separated, for a better traceability and to keep the commit messages.

FABulous/fabric_definition/Tile.py
```diff
@@ -21,6 +22,8 @@ class Tile:
         The list of ports of the tile
     matrixDir : str
         The directory of the tile matrix
+    gen_ios : (List[Gen_IO])
```
KelvinChung2000: should not need bracket

FABulous/fabric_definition/Tile.py
```diff
@@ -35,6 +38,7 @@ class Tile:
     portsInfo: list[Port]
     bels: list[Bel]
     matrixDir: pathlib.Path
+    gen_ios = list[Gen_IO]
```
KelvinChung2000: should be ":"

FABulous/fabric_generator/code_generation_VHDL.py
```diff
@@ -108,20 +108,38 @@ def addLogicStart(self, indentLevel=0):
     def addLogicEnd(self, indentLevel=0):
         self._add("\n" f"end" "\n", indentLevel)
 
-    def addAssignScalar(self, left, right, delay=0, indentLevel=0):
+    def addRegister(self, reg, regIn, clk="CLK", inverted=False, indentLevel=0):
```
KelvinChung2000, line 111: For large text templates, I think putting the text at the top and then using `.format` should make things look cleaner. We can do this sort of "cosmetic" update later

docs/source/fabric_definition.rst
KelvinChung2000, line 1: Adding information about the created port still needs information would be useful.

KelvinChung2000 (commented):
I assume we will no longer need the W_IO tile with this? Maybe should also update the demo to use the feature.

PR #256, opened 2024-12-16
workflows: Update workflows for ubuntu 24

EverythingElseWasAlreadyTaken:
Remove legacy python3.10 call.
Remove installation of python-tk since it is already included in the default binares since Python3.7.
Update docs and README.

Since we are using 'ubunut-latest' for the workflows, the workflows failed since python-tk was removed from the Ubunut PPA.

### DIFF
--- a/.github/workflows/fabric_gen.yml
+++ b/.github/workflows/fabric_gen.yml
@@ -20,10 +20,7 @@ jobs:
         run: |
           python3 -m pip install --upgrade pip
           pip3 install flake8 pytest
-          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
-          if [ -f requirements.txt ]; then python3.10 -m pip install -r requirements.txt; fi
-          sudo apt update
-          sudo apt install python-tk -y
+          if [ -f requirements.txt ]; then pip3 install -r requirements.txt; fi
       - name: Lint with flake8
         run: |
           # stop the build if there are Python syntax errors or undefined names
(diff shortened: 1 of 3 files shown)


KelvinChung2000:
Is TK installed by default at  python3.12?

EverythingElseWasAlreadyTaken:
> Is TK installed by default at python3.12?

Yes, since Python 3.7 it is shipped with the default binaries, seems that Ubuntu removed it from their PPAs with their latest update. 

https://tkdocs.com/tutorial/install.html#installlinux

KelvinChung2000:
Ok. I think should leave a note on things earlier than 24 might still need to do it via apt install.

EverythingElseWasAlreadyTaken:
> Ok. I think should leave a note on things earlier than 24 might still need to do it via apt install.

We require at least Python 3.12 to install and use FABulous, which already includes TK. 
If someone is using an older Ubuntu, you have to update your Python anyway.

KelvinChung2000:
My disk got corrupted, which gave me a chance to test a fresh Python install.
```
kelvin@pop-os:~$ python3.12
Python 3.12.8 (main, Dec  4 2024, 08:54:12) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tkinter 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'tkinter'
>>> 
```
I am using 22.04, and you still need to install it manually.

EverythingElseWasAlreadyTaken:
Interesting, seems that they built their own Python binaries without tk support, before Ubuntu 24. 
I'll add a note for older Systems and do a new PR.

PR #274, opened 2025-01-13
Pre commit hooks and format

KelvinChung2000:
Adding pre-commit hooks to the repo to aid development and apply the pre-commit settings. 

Please let me know if you have any other hooks you would like to add, and I will include them all. 

One specific one in mind is [file header](https://github.com/Quantco/pre-commit-mirrors-insert-license?tab=readme-ov-file). But I want to check what we want first.


KelvinChung2000:
Our current doc coverage is at 40%. We should put some work into improving it.

IAmMarcelJung:
Looks cool! Probably we should split adding the hooks and the actual formatting if that's possible? Or at least mention it in the PR?

[This one](https://github.com/Lucas-C/pre-commit-hooks) also has support to forbid/remove tabs, maybe we could also add that.

PR #281, opened 2025-01-21
fabric_gen: Generate configMem files only if needed

EverythingElseWasAlreadyTaken:
Add check if configMem files needed to be generated for a tile. Tile_configMem.csv and also Tile_configMem.v were always generated, even if they are not needed for a tile.

Remove hardcoded NoConfigBit parameters in eFPGA.v and supertile tops, since they are not needed and cause linter issues.


IAmMarcelJung (approved):
Thanks for this! Looks good to me.

PR #293, opened 2025-01-23
code_gen_VHDL: Fix issue with VHDL module instantiation

EverythingElseWasAlreadyTaken:
The generated VHDL code for module instantiations of switch matrix multiplexers, concatenated vectors with {}, which is not valid VHDL.

Replace {} with () for aggregation in VHDL.

### DIFF
--- a/FABulous/fabric_generator/code_generation_VHDL.py
+++ b/FABulous/fabric_generator/code_generation_VHDL.py
@@ -147,16 +147,15 @@ class VHDLWriter(codeGenerator):
         self._add(f"Port map(", indentLevel=indentLevel + 1)
         connectPair = []
         for i in portsPairs:
-            if "[" in i[0]:
-                port = i[0].replace("[", "(").replace("]", ")").replace(":", " downto ")
-            else:
-                port = i[0]
-            if "[" in i[1]:
+            # NOTE: This is a temporary fix for the issue of curly braces in the port names and needs to be fixed properly a later refactoring of the code generation
+            port = i[0].replace("{", "(").replace("}", ")")
+            signal = i[1].replace("{", "(").replace("}", ")")
+            if "[" in port:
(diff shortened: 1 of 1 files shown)


EverythingElseWasAlreadyTaken:
This is a diff of the changes in a VHDL project before and after this commit: 
[diff.txt](https://github.com/user-attachments/files/18523834/diff.txt)

KelvinChung2000 (commented):
I don't like how I done this. We should change it to something more robust one day. Are we getting the port pair from a formated string? Because I do not recall we are doing it this way...

EverythingElseWasAlreadyTaken:
> I don't like how I done this. We should change it to something more robust one day. Are we getting the port pair from a formated string? Because I do not recall we are doing it this way...

Yes, I know, this is not really the best way. This is just a quick fix since it was broken and didn't want to start a refactoring project. 

The other easy way would be fixing it here:
https://github.com/FPGA-Research/FABulous/blob/FABulous2.0-development/FABulous/fabric_generator/fabric_gen.py#L1137
But this would introduce more Language dependent code and does not catch the general case.

In general would it be better, if portsPair would be just a better data structure, that the writer can handle accordingly to the language and not a list of sting-tuples, that needs to be formatted in the right way.  
But this should in the best case be done in a general refactoring of the writers.

KelvinChung2000:
Refactoring the writer will be my solution. But also the most time-consuming option. 

Maybe put a comment on this as a temporary fix, get this merged, and deal with it during refactoring.

EverythingElseWasAlreadyTaken:
> Refactoring the writer will be my solution. But also the most time-consuming option.
> 
> Maybe put a comment on this as a temporary fix, get this merged, and deal with it during refactoring.

Yes, this would be the best solution, and needs to be done at some point. 

I've added a comment.

PR #301, opened 2025-01-24
Fix #300

### LINKED ISSUE #300: Shell is entered instead of throwing an error when the script file is misspelled
**Describe the bug**
As the title says, when calling FABulous with the script argument (either `-ts` or `-fs`) and the argument is not given correctly, the tool enters the shell instead of throwing an error. This is just a minor thing and should be handled with low priority but I still want to leave this here. 

**To Reproduce**
1. Go to `FAB_ROOT` and create a new project or use an exisiting onr
2. Execute command with a wrong script argument (e.g. strip off the "l"): `FABulous <project_name> -fs=./<project_name>/FABulous.tc
3. The FABulous shell should

**Expected behavior**
I would expect the tool to throw an error instead of entering the shell.

**Screenshots**

Image

**My Setup:**
 - OS: Ubuntu 24.04 LTS
 - Kernel version: 6.8.0-51-generic
 - Python version: 3.12.03
 - FABulous version: dev branch @ 1daab1ff24ae5cc95f985d329eb96fe0731a9ec9

KelvinChung2000:
Correctly exit shell when the provided path does not exists.

### DIFF
--- a/FABulous/FABulous_CLI/FABulous_CLI.py
+++ b/FABulous/FABulous_CLI/FABulous_CLI.py
@@ -196,10 +196,16 @@ class FABulous_CLI(Cmd):
         if not TCLScript.is_dir() and TCLScript.exists():
             self._startup_commands.append(f"run_tcl {TCLScript}")
             self._startup_commands.append("exit")
+        elif not TCLScript.is_dir() and not TCLScript.exists():
+            logger.error(f"Cannot find {TCLScript}")
+            exit(1)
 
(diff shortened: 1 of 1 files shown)


IAmMarcelJung (approved):
Thanks for this, it fixes the mentioned issue. However, I noticed that the shell is still entered if a valid directory is given and also FABulous tries to execute a script even if its just a .txt file. I will open a new issue for this.

PR #304, opened 2025-01-26
Add support for SystemVerilog in BELs

mole99:
This PR allows BELs to be written in SystemVerilog.

At first I thought of adding a separate entry for `"systemverilog"`, but since Verilog is a subset of SystemVerilog, it is safe and sane to treat both cases the same.

This enables using the SV support in yosys and thus the use of e.g. `always_comb` and `always_ff` to avoid shooting yourself in the foot.

Let me know if you agree with these changes.


KelvinChung2000:
Looks good to me

PR #306, opened 2025-01-27
Improve FrameData and FrameStrobe handling

### LINKED ISSUE #296: Tiles with no config bits have no `FrameData` signals
This may seem like a feature at first: Tiles that do not use any configuration bits have no `FrameData`/`FrameData_O` signals. But for the end user this is a actually a bug:

Tiles without any config bits and tiles with config bits can be mixed in a row. This requires all tiles to have `FrameData`/`FrameData_O` signals, even those without any config bits. Currently, this will result in a broken fabric.

### LINKED ISSUE #297: Special handling of first and last row of fabric
When `ConfigBitMode` is set to `FRAME_BASED`, FABulous will:

1. Unconditionally generate `FrameStrobe` signals for all columns ✅
2. Generate `FrameData` signals for all rows except for the first and last ❌

Point 2 may look like a feature at first, if none of the tiles in the first and last row use configuration bits (as is the case in the example fabric), then there is no need to generate `FrameData` signals, right?

- What if you want to use different tiles in the first and last row, e.g. `N_IO` and `S_IO`?
- There is no way to configure this behavior.
- There is no way to disable this behavior.

Your first reaction might be to make this behavior optional, but the better approach is to remove this feature **now** and reimplement it *correctly*[^1] at a later time.

My PR in #161 fixes the `FrameData` generation for the first and last row by removing the hardcoded exception. Additionally, if the first tile in a row is `NULL`, than `Tile_Y{y}_FrameData` is correctly connected to the second tile.

This works fine if you are using e.g. only `N_IO` and `S_IO` tiles in the first and last row (as in my case). But if these tiles are termination tiles without any configuration bits (as in the example fabric) or any mix of them, then some tiles have no `FrameData` inputs and `FrameData_O` outputs, resulting in a broken fabric.

This means my PR depends on every tile, no matter if it has config bits or not, having `FrameData`/`FrameData_o` inputs/outputs, which is a sensible assumption. Remember: for `FrameStrobe` signals this is already the case, why should `FrameData` be any different? See #296 for tracking this dependent issue.

I would like to hear your feedback before I bring PR #161 up to date and integrate the fixes for #296 into it. Thanks!

[^1]: A correct reimplementation would be one that handles `FrameData` as well as `FrameStrobe` signals: If all tiles of a row or column don't use any configuration bits, only then FABulous should disable the `FrameData`/`FrameStrobe` signals for that row/column. The problem is that tiles will always need to be generated with `FrameData`/`FrameStrobe` signals (independent of whether they use any config bits at all), since there is no way of knowing in advance how the tiles will get stitched (and that's a good thing, because it keeps FABulous flexible). So it's probably best to forget about this “feature” altogether, as it really makes no real difference whether the first and last row have `FrameData` signals or not ;)

mole99:
This PR fixes/improves how FrameData and FrameStrobe signals are connected within the fabric.

1. Tiles: Always generate FrameData ports
    - Fixes #296
    - This allows mixing tiles with no config bits and tiles with config bits in the same row
2. Remove special handling of first and last row
    - Fixes #297
    - This allows to have tiles with config bits in the first and last row (e.g. `N_IO`, `S_IO`)
3. Generalize handling of NULL tiles
    - Supersedes #161
    - A major improvement to how NULL tiles are currently handled

The general behavior is as follows:

For each tile connect the FrameData input port to the FrameData output port of the previous tile. If the previous tile is `NULL`, the tile before is used (if that one is `NULL` repeat again). If all tiles before the tile are `NULL`, connect directly to the FrameData port of the fabric. The same behavior applies to the FrameStrobe signals.

In theory, this makes it possible to create donut-shaped fabrics (I haven't tried it as this would require more types of termination tiles).

This PR also includes fixes to supertiles:

- The port direction for external I/Os was incorrect
- Connect external I/Os to the ports

The demo fabric is working as before.

I look forward to your feedback to get this merged soon. Thanks!

### DIFF
--- a/FABulous/fabric_generator/fabric_gen.py
+++ b/FABulous/fabric_generator/fabric_gen.py
@@ -840,18 +840,18 @@ class FabricGenerator:
 
         if self.fabric.configBitMode == ConfigBitMode.FRAME_BASED:
             self.writer.addPortVector(
-                "FrameData", IO.INPUT, "FrameBitsPerRow -1", indentLevel=2
+                "FrameData", IO.INPUT, "FrameBitsPerRow-1", indentLevel=2
             )
             self.writer.addComment("CONFIG_PORT", onNewLine=False, end="")
             self.writer.addPortVector(
-                "FrameData_O", IO.OUTPUT, "FrameBitsPerRow -1", indentLevel=2
+                "FrameData_O", IO.OUTPUT, "FrameBitsPerRow-1", indentLevel=2
             )
             self.writer.addPortVector(
-                "FrameStrobe", IO.INPUT, "MaxFramesPerCol -1", indentLevel=2
+                "FrameStrobe", IO.INPUT, "MaxFramesPerCol-1", indentLevel=2
             )
             self.writer.addComment("CONFIG_PORT", onNewLine=False, end="")
             self.writer.addPortVector(
-                "FrameStrobe_O", IO.OUTPUT, "MaxFramesPerCol -1", indentLevel=2
+                "FrameStrobe_O", IO.OUTPUT, "MaxFramesPerCol-1", indentLevel=2
             )
 
         elif self.fabric.configBitMode == ConfigBitMode.FLIPFLOP_CHAIN:
@@ -1690,12 +1690,12 @@ class FabricGenerator:
             # FrameBitsPerRow : integer := 32;
             for y in range(self.fabric.numberOfRows):
                 self.writer.addConnectionVector(
-                    f"Tile_Y{y}_FrameData", "FrameBitsPerRow -1"
+                    f"Row_Y{y}_FrameData", "FrameBitsPerRow -1"
                 )
 
             for x in range(self.fabric.numberOfColumns):
                 self.writer.addConnectionVector(
-                    f"Tile_X{x}_FrameStrobe", "MaxFramesPerCol - 1"
+                    f"Column_X{x}_FrameStrobe", "MaxFramesPerCol - 1"
                 )
 
             for y in range(self.fabric.numberOfRows):
@@ -1741,14 +1741,14 @@ class FabricGenerator:
         if self.fabric.configBitMode == ConfigBitMode.FRAME_BASED:
             for y in range(len(self.fabric.tile)):
                 self.writer.addAssignVector(
-                    f"Tile_Y{y}_FrameData",
+                    f"Row_Y{y}_FrameData",
                     "FrameData",
                     f"FrameBitsPerRow*({y}+1)-1",
                     f"FrameBitsPerRow*{y}",
                 )
             for x in range(len(self.fabric.tile[0])):
                 self.writer.addAssignVector(
-                    f"Tile_X{x}_FrameStrobe",
+                    f"Column_X{x}_FrameStrobe",
                     "FrameStrobe",
                     f"MaxFramesPerCol*({x}+1)-1",
                     f"MaxFramesPerCol*{x}",
@@ -2008,7 +2008,7 @@ class FabricGenerator:
                         # FrameData_O signals.
                         # If the previous tile is NULL, continue the search.
                         # If all previous tiles are NULL, connect to the fabrics
-                        # Tile_Y{y}_FrameData signals.
+                        # Row_Y{y}_FrameData signals.
 
                         done = False
 
@@ -2038,10 +2038,10 @@ class FabricGenerator:
                             break
 
                         # No non-NULL tile was found, and tile is not part of a supertile.
-                        # Connect to the fabrics Tile_Y{y}_FrameData signals.
+                        # Connect to the fabrics Row_Y{y}_FrameData signals.
                         if not done:
                             portsPairs.append(
-                                (f"{pre}FrameData", f"Tile_Y{supertile_y}_FrameData")
+                                (f"{pre}FrameData", f"Row_Y{supertile_y}_FrameData")
                             )
 
                         # Connecting FrameData_O is easier:
@@ -2059,7 +2059,7 @@ class FabricGenerator:
                         # FrameStrobe_O signals.
                         # If the previous tile is NULL, continue the search.
                         # If all previous tiles are NULL, connect to the fabrics
-                        # Tile_X{x}_FrameStrobe signals.
+                        # Column_X{x}_FrameStrobe signals.
 
                         done = False
 
@@ -2093,12 +2093,12 @@ class FabricGenerator:
                             break
 
                         # No non-NULL tile was found, and tile is not part of a supertile.
-                        # Connect to the fabrics Tile_X{x}_FrameStrobe signals.
+                        # Connect to the fabrics Column_X{x}_FrameStrobe signals.
                         if not done:
                             portsPairs.append(
                                 (
                                     f"{pre}FrameStrobe",
-                                    f"Tile_X{supertile_x}_FrameStrobe",
+                                    f"Column_X{supertile_x}_FrameStrobe",
                                 )
                             )
 
--- a/FABulous/fabric_generator/fabric_gen.py
+++ b/FABulous/fabric_generator/fabric_gen.py
@@ -2013,7 +2013,7 @@ class FabricGenerator:
                         done = False
 
                         # Get all x-positions before this tile
-                        for search_x in range(supertile_x - 1, 0 - 1, -1):
+                        for search_x in range(supertile_x - 1, -1, -1):
 
                             # Previous tile is part of the same supertile.
                             # FrameData signals are connected internally.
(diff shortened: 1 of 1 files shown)


FABulous/fabric_generator/fabric_gen.py
```diff
@@ -978,24 +973,24 @@ def generateTile(self, tile: Tile) -> None:
...
-            for i in range(self.fabric.frameBitsPerRow):
-                self.writer.addInstantiation(
-                    "my_buf",
-                    f"data_outbuf_{i}",
-                    portsPairs=[
-                        ("A", f"FrameData_O_i[{i}]"),
-                        ("X", f"FrameData_O[{i}]"),
-                    ],
-                )
+        # buffer FrameData signals
```
KelvinChung2000, line 976: I am unsure if we should always add the buffer. I guess they won't be optimised away if they are not needed?

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -1990,69 +2000,118 @@ def generateFabric(self) -> None:
...
+                            # Previous tile is part of the same supertile.
+                            # FrameStrobe signals are connected internally.
+                            # Stop the search and be done.
+                            if (supertile_x, search_y) in superTileLoc:
+                                done = True
+                                break
+
+                            # Previous tile is NULL, continue search
+                            if self.fabric.tile[search_y][supertile_x] == None:
+                                print("skip")
```
KelvinChung2000: Can put to `logger.debug`

FABulous/fabric_generator/fabric_gen.py
mole99: Since other tiles with config bits will have them, I would opt to keep the buffers both for consistency and to ensure that the timing of the FrameData signals of all rows is the same in an ASIC.

FABulous/fabric_generator/fabric_gen.py
mole99: Whoops, that was left over from testing. I'll remove it.

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -836,28 +839,20 @@ def generateTile(self, tile: Tile) -> None:
...
+            self.writer.addPortVector(
+                "FrameData_O", IO.OUTPUT, "FrameBitsPerRow -1", indentLevel=2
+            )
+            self.writer.addPortVector(
+                "FrameStrobe", IO.INPUT, "MaxFramesPerCol -1", indentLevel=2
+            )
+            self.writer.addComment("CONFIG_PORT", onNewLine=False, end="")
+            self.writer.addPortVector(
+                "FrameStrobe_O", IO.OUTPUT, "MaxFramesPerCol -1", indentLevel=2
+            )
```
IAmMarcelJung: It's unrelated to the changes you did, but maybe we also remove the whitespace in all of these occurences?
```suggestion
            self.writer.addPortVector(
                "FrameData", IO.INPUT, "FrameBitsPerRow-1", indentLevel=2
            )
            self.writer.addComment("CONFIG_PORT", onNewLine=False, end="")
            self.writer.addPortVector(
                "FrameData_O", IO.OUTPUT, "FrameBitsPerRow-1", indentLevel=2
            )
            self.writer.addPortVector(
                "FrameStrobe", IO.INPUT, "MaxFramesPerCol-1", indentLevel=2
            )
            self.writer.addComment("CONFIG_PORT", onNewLine=False, end="")
            self.writer.addPortVector(
                "FrameStrobe_O", IO.OUTPUT, "MaxFramesPerCol-1", indentLevel=2
            )
```

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -1990,69 +2000,116 @@ def generateFabric(self) -> None:
...
-                            elif (x + i - 1, y + j) not in superTileLoc:
-                                portsPairs.append(
-                                    (
-                                        f"{pre}FrameData_O",
-                                        f"Tile_X{x+i}Y{y+j}_FrameData_O",
-                                    )
+                        done = False
+
+                        # Get all x-positions before this tile
+                        for search_x in range(supertile_x - 1, 0 - 1, -1):
```
IAmMarcelJung: `0 - 1` looks a bit weird, is this intended?

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -1990,69 +2000,116 @@ def generateFabric(self) -> None:
...
+                            break
+
+                        # No non-NULL tile was found, and tile is not part of a supertile.
+                        # Connect to the fabrics Tile_X{x}_FrameStrobe signals.
+                        if not done:
                             portsPairs.append(
                                 (
                                     f"{pre}FrameStrobe",
-                                    f"Tile_X{x+i}Y{y+j+1}_FrameStrobe_O",
+                                    f"Tile_X{supertile_x}_FrameStrobe",
```
IAmMarcelJung: Same question as above, just with `Row` and `Column` switched :)

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -1990,69 +2000,116 @@ def generateFabric(self) -> None:
...
-                        if y + 1 >= self.fabric.numberOfRows:
+                            done = True
+                            break
+
+                        # No non-NULL tile was found, and tile is not part of a supertile.
+                        # Connect to the fabrics Tile_Y{y}_FrameData signals.
+                        if not done:
                             portsPairs.append(
-                                (f"{pre}FrameStrobe", f"Tile_X{x}_FrameStrobe")
+                                (f"{pre}FrameData", f"Tile_Y{supertile_y}_FrameData")
```
IAmMarcelJung: Maybe we could use `Row` instead of `Tile` here since this does refer to the strobe for the whole row instead of a single tile (I hope I don't interchange row/column...)?

IAmMarcelJung (commented):
Thanks for this! I also tried it out for the demo fabric. I just have a suggestion and a few remarks/questions :)

FABulous/fabric_generator/fabric_gen.py
mole99: Done!

FABulous/fabric_generator/fabric_gen.py
mole99: It's intended to show that the end value is 0, but to actually include 0 we need to subtract one.

FABulous/fabric_generator/fabric_gen.py
mole99: I agree, that makes it clearer, done!

FABulous/fabric_generator/fabric_gen.py
mole99: Also done!

FABulous/fabric_generator/fabric_gen.py
IAmMarcelJung: I get your point (and both like and appreciate that your think about this), but isn't -1 pretty standard for the range function? Asking rather out of curiousity than doing a suggesting here :) 

Or we could introduce a variable for the westernmost tile and set it to -1? Don't know if this would be good practice though...

mole99: I have no issue with changing it, just stated that it was intended ;)
Done!

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -1990,69 +2000,116 @@ def generateFabric(self) -> None:
...
 
-                            elif (x + i - 1, y + j) not in superTileLoc:
-                                portsPairs.append(
-                                    (
-                                        f"{pre}FrameData_O",
-                                        f"Tile_X{x+i}Y{y+j}_FrameData_O",
-                                    )
+                        done = False
+
+                        # Get all x-positions before this tile
```
IAmMarcelJung:
```suggestion
                        # Get all x-positions to the west of this tile
```

Maybe this a bit more precise?

FABulous/fabric_generator/fabric_gen.py
IAmMarcelJung: Ok thanks :)

FABulous/fabric_generator/fabric_gen.py
mole99: Yes, good point! I also updated the other comments.

EverythingElseWasAlreadyTaken (approved):
Thanks for the PR, this looks good to me!

mole99:
My PR was specifically targeting FrameData and FrameStrobe signals and does not consider routing wires. This means you can use NULL tiles to change the shape of the fabric, e.g. make a notch, but you cannot place NULL tiles where the adjacent tile has routing wires at that side.

For DRC (Design Rule Check) it does not matter whether you buffer those signals or not.

KelvinChung2000:
Yes, that's a new issue, which means more stuff might need fixing on our side or behaviour-defining.  

The buffer is for antenna checks. This does not happen with hardening just a single tile but when the fabric is stitched.

mole99:
True! But the PnR tool will automatically add an antenna diode or insert a jumper (switching the layer) to prevent antenna issues. I'm not sure if OpenROAD already has the capability to insert jumpers.

PR #307, opened 2025-01-27
Fix run_simulation output

### LINKED ISSUE #305: non-matching result from run_simulation
The `run_simulation` command is creating:

```fabric(I_top) = 0x000xxxx gold = 0x000xxxx, fabric(T_top) = 0xffffffe gold = 0xffffffe```

instead of:

```fabric(I_top) = 0x48c018c gold = 0x48c018c, fabric(T_top) = 0xffffffe gold = 0xffffffe```

which does not match the test script result.

### PR DESCRIPTION
**Pull request by IAmMarcelJung**
This fixes #305 by correctly enabling and resetting the counter in the testbench used in the `run_simulation` command.

It should be noted that this still does not match the output of the script `run_simulation.sh`, since the command simulates the user design instead of the `counter.v` test design. I will open a discussion on what we want the `run_simulation` command to actually test.

@mole99 could you check if this resolves the issue of the wrong output on your end (according to the things I mentioned above)? I ofc did check it on my end, but better safe than sorry :)

### DIFF
--- a/FABulous/fabric_files/FABulous_project_template_verilog/Test/sequential_16bit_en_tb.v
+++ b/FABulous/fabric_files/FABulous_project_template_verilog/Test/sequential_16bit_en_tb.v
@@ -88,9 +88,11 @@ module sequential_16bit_en_tb;
         end
 `endif
         repeat (100) @(posedge CLK);
-        O_top = 28'b1; // reset
+        // Enable and reset the counter
+        O_top = 28'b0000_0000_0000_0000_0000_0000_0011;
         repeat (5) @(posedge CLK);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mole99** [comment]
@IAmMarcelJung Yes, this work now on my end :)

```
fabric(I_top) = 0x0000000 gold = 0x0000000, fabric(T_top) = 0xffffffe gold = 0xffffffe
fabric(I_top) = 0x0000001 gold = 0x0000001, fabric(T_top) = 0xffffffe gold = 0xffffffe
fabric(I_top) = 0x0000002 gold = 0x0000002, fabric(T_top) = 0xffffffe gold = 0xffffffe
```

PR #309, opened 2025-01-29
fabric_generator: Fix broken VHDL fabric generation

EverythingElseWasAlreadyTaken:
fabric_gen:
  - Fix: Add missing `begin` / LogicStart in ConfigMem generation.
  - Generate NoConfigBits Parameter in Tiles only if it has config bits.
  - Remove unused/typo signal `resten` in top wrapper generation.

file_parser:
  - Fix: Add individually_declared setting also in VHDL bel port parsing, since individual declared VHDL bel ports were detected and connected as vectors, which resulted in port mismatches.

### DIFF
--- a/FABulous/fabric_generator/fabric_gen.py
+++ b/FABulous/fabric_generator/fabric_gen.py
@@ -420,6 +420,7 @@ class FabricGenerator:
             self.writer.addPreprocElse()
         self.writer.addNewLine()
         self.writer.addNewLine()
+        self.writer.addLogicStart()
         self.writer.addComment("instantiate frame latches", end="")
         for i in configMemList:
             counter = 0
@@ -494,12 +495,15 @@ class FabricGenerator:
         # we pass the NumberOfConfigBits as a comment in the beginning of the file.
         # This simplifies it to generate the configuration port only if needed later when building the fabric where we are only working with the VHDL files
 
-        # VHDL header
+        # Generate header
         self.writer.addComment(f"NumberOfConfigBits: {noConfigBits}")
         self.writer.addHeader(f"{tile.name}_switch_matrix")
-        self.writer.addParameterStart(indentLevel=1)
-        self.writer.addParameter("NoConfigBits", "integer", noConfigBits, indentLevel=2)
-        self.writer.addParameterEnd(indentLevel=1)
+        if noConfigBits > 0:
+            self.writer.addParameterStart(indentLevel=1)
+            self.writer.addParameter(
+                "NoConfigBits", "integer", noConfigBits, indentLevel=2
+            )
+            self.writer.addParameterEnd(indentLevel=1)
         self.writer.addPortStart(indentLevel=1)
 
         # normal wire input
@@ -2224,7 +2228,6 @@ class FabricGenerator:
         self.writer.addConnectionVector("LocalWriteData", 31)
         self.writer.addConnectionScalar("LocalWriteStrobe")
         self.writer.addConnectionVector("RowSelect", "RowSelectWidth-1")
-        self.writer.addConnectionScalar("resten")
 
         if isinstance(self.writer, VHDLWriter):
             basePath = Path(self.writer.outFileName).parent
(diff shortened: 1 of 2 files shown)


FABulous/fabric_generator/fabric_gen.py
```diff
@@ -420,6 +420,7 @@ def generateConfigMem(self, tile: Tile, configMemCsv: str) -> None:
             self.writer.addPreprocElse()
         self.writer.addNewLine()
         self.writer.addNewLine()
+        self.writer.addLogicStart()
```
KelvinChung2000, line 423: I have started working on the new code generator since I need it to generate techmap map automatically. The new generator can do 
```python
with codeRegion():
   statement()
   statement()
```
Which then can avoid problem like this.

FABulous/fabric_generator/file_parser.py
```diff
@@ -829,6 +829,9 @@ def parseBelFile(
             else:
                 internal.append((portName, direction))
 
+            if portName[-1].isdigit():
+                individually_declared = True
```
KelvinChung2000, line 847: I think we should move away from "individually declare" since this is just not natural to write and also is valid to have a digit as the character of the name. This will do for now.

FABulous/fabric_generator/file_parser.py
```diff
@@ -903,6 +900,12 @@ def parseBelFile(
                 f"NoConfigBits does not match with the BEL map in file {filename}, length of BelMap is {len(belMapDic)}, but with {noConfigBits} config bits"
             )
 
+    if individually_declared:
+        logger.warning(
+            f"Ports in {filename} have been individually declared rather than as a vector."
+        )
+        logger.warning("Ports will not be concatenated during fabric generation.")
```
KelvinChung2000, line 923: If this will actually result in an error, should we raise an exception?

KelvinChung2000 (approved):

FABulous/fabric_generator/fabric_gen.py
EverythingElseWasAlreadyTaken: Sounds cool! 
Looking forward to this, since the current code generation can really cause a lot of issues.

FABulous/fabric_generator/file_parser.py
EverythingElseWasAlreadyTaken: Yes, this is not really a good approach in general, which can also cause problems.
This was just a quick fix to get the fabric running again.

KelvinChung2000: Let's mark it with a TODO/FIXME.

FABulous/fabric_generator/file_parser.py
EverythingElseWasAlreadyTaken: I think it is a warning is fine, since this is not an error itself, but it can lead to some errors.
This is more an issue of how the code parsing and generation is currently working.

In general, we should maybe also switch to the yosys-ghdl plugin to parse the VHDL bels.

FABulous/fabric_generator/file_parser.py
EverythingElseWasAlreadyTaken: Yes, I'll add a TODO there.

FABulous/fabric_generator/file_parser.py
EverythingElseWasAlreadyTaken: Yes you are right, at least for VHDL this is an issue, since we don't support vectors here.
I'll update the commit.

In general, we should maybe also switch to the yosys-ghdl plugin to parse the VHDL bels.

KelvinChung2000: That will require us to update all the bels to use the VHDL attribute. It's annoying to do, but we can work on that later.

EverythingElseWasAlreadyTaken: > That will require us to update all the bels to use the VHDL attribute. It's annoying to do, but we can work on that later.

I'll open an issue for this.

EverythingElseWasAlreadyTaken: Issue #310

PR #311, opened 2025-02-03
Include first/last row in bitstream generation and ConfigFSM

mole99:
Now that #306 has been merged, we need to actually make use of the first and last row of the fabric.

Therefore this PR:

- Modifies `bit_gen.py` to dump the config bits for the first and last row
- Modifies `ConfigFSM` to write that information to the FrameData's of the first and last row
  - I tested the Verilog implementation of `ConfigFSM`, the VHDL version is untested but should behave the same

As a bonus, I finally added a desync frame to `bit_gen.py`. This means after configuration has finished, `ConfigFSM` will actually go back into the idle state and it is possible to upload a different bitstream without resetting the circuit.

This is shown here working:

Bildschirmfoto vom 2025-02-03 09-32-35


KelvinChung2000 (commented):
The changes logically make sense to me. But I have no knowledge in the configuration side.

mole99:
Thanks for taking a look. I would like to add that I have successfully configured and simulated a fabric with I/Os in the top row and other tiles requiring config bits in the bottom row.

FABulous/fabric_files/FABulous_project_template_verilog/Fabric/ConfigFSM.v
```diff
@@ -53,15 +53,15 @@ module ConfigFSM (CLK, resetn, WriteData, WriteStrobe, FSM_Reset, FrameAddressRe
...
                                 state <= 2; //writing frame data
                             end
                         end
                     end
                     2: begin
                         if(WriteStrobe == 1'b1) begin// if writing enabled
-                            FrameShiftState <= FrameShiftState -1 ;
-                            if(FrameShiftState == 1) begin // on last frame
+                            FrameShiftState <= FrameShiftState - 1;
+                            if(FrameShiftState == 0) begin // on last frame
```
IAmMarcelJung, line 64: I noticed that this causes the FrameShiftState to become 0x1F/31 (an invalid value). This also causes an invalid value for RowSelectin even if we have a WriteStrobe:
image
This is how it previously looked like for the transition from bottom to top row:
image


@KelvinChung2000 do you think this could be an issue or is this fine?
The simulation runs without errors, but since it's not really sophisticated, I don't trust it too much.

FABulous/fabric_files/FABulous_project_template_vhdl/Fabric/ConfigFSM.vhdl
```diff
@@ -79,14 +79,14 @@ begin
...
-                FrameShiftState <= to_unsigned(NumberOfRows, FrameShiftState'length);
+                FrameShiftState <= to_unsigned(NumberOfRows + 1, FrameShiftState'length);
                 state <= "10";
               end if;
             end if;
           when "10" =>
             if WriteStrobe = '1' then
               FrameShiftState <= FrameShiftState - "00001";
-              if FrameShiftState = X"00000001" then
+              if FrameShiftState = X"00000000" then
```
IAmMarcelJung: I didn't check this but it probably causes the same behavior as mentioned above.

IAmMarcelJung (changes requested):
Thanks for this! See my comment for why I marked this as "changes requested". The rest looks fine so if this is resolved we can merge this.

As a sidenote, I don't like our current handling of "NumberOfRows". In `fabric_gen.py` the value for the generated `eFPGA_top.v` is set to the actual number of rows - 2. Probably this is a leftover from the previous special treatment of the top and bottom row? We should  maybe think about setting this value to the actual number of rows and change all calculations done with this variable (or parameter on the Verilog/VHDL side).

IAmMarcelJung:
> I would like to add that I have successfully configured and simulated a fabric with I/Os in the top row and other tiles requiring config bits in the bottom row.

If this works, then it's probably fine, but I would rather have this double checked since this is touching core parts of the config logic :)

mole99:
@IAmMarcelJung Sorry for the confusion, with this comment I was mainly referring to the changes in `bit_gen`. I used the generated bitstream for that fabric, but I wrote my own configuration logic for it. 

Thus the changes to `ConfigFSM` were only tested with the demo fabric, therefore I would really appreciate another pair of eyes on it.

dirk-koch:
I think making the config bits mandatory is not a good idea.
The first one is that we cannot use recent FABulous to generate bitstreams for older fabrics.
The simple solution so far was to use configuration bits in the top and bottom tiles is by adding an extra row of NULL tiles in the top and bottom rows. Not beautiful but safe.

The reason why we haven't foreseen config bits by default is that in "just eFPGA" mode, we will likely not need the top and bottom columns for something "fancy" and just do a loopback (as done, for instance, by most Xilinx FPGAs). Connections to a fabric should not be just applied arbitrary but following the bottom to top vectorization scheme in mind.
Meaning: carry chains , DSP-blocks etc. all use a connection scheme where vectors have LSBs in the bottom and MSBs at the top. This is also the reason why "main connections" sit at the left or right border of the fabric.

Anyway, a clean approach is scanning through the fabric row-by-row and count the number of configuration bits needed (after we generated all tiles) and then decide for each row it it needs configuration bits. We could do that with thresholds for, let's say 8, 16, 32 FrameData bits.
This way, our old fabrics would still work as usual and nothing should break.
Additionally, we could also foresee setting an optional variable to set the FrameData bits manually. That would still require a feasibility check before generating the fabric.
There are changes to the configuration logic planned that will be rolled out (hopefully) soon and we will definitely look into the first/last row issue.

mole99:
@dirk-koch Thank you for your reply. I understand your points, but I do not fully agree with them:

>  The first one is that we cannot use recent FABulous to generate bitstreams for older fabrics.

There is an easy way to still be able to generate bitstreams for older FABulous fabrics: by simply adding a "legacy" mode to `bit_gen`. If `--legacy` is passed, then the config bits of the first and last row are not output to the bitstream and the bitstream is therefore backwards compatible.

I can add that change to this PR right away.

> The simple solution so far was to use configuration bits in the top and bottom tiles is by adding an extra row of NULL tiles in the top and bottom rows. Not beautiful but safe.

That workaround wasn't directly obvious to me and probably to other users of FABulous either. Aside from not being pretty, as you mentioned, it wouldn't work well with my automatic stitching approach as the row of NULL-tiles would add empty space to the fabric macro.

> The reason why we haven't foreseen config bits by default is that in "just eFPGA" mode, we will likely not need the top and bottom columns for something "fancy" and just do a loopback (as done, for instance, by most Xilinx FPGAs). Connections to a fabric should not be just applied arbitrary but following the bottom to top vectorization scheme in mind.

Yes, I understand that in some situations you don't need configuration bits in the top and bottom row. But what if you need them, as in my case?

> Meaning: carry chains , DSP-blocks etc. all use a connection scheme where vectors have LSBs in the bottom and MSBs at the top. This is also the reason why "main connections" sit at the left or right border of the fabric.

True, but in my fabric I am already pretty tile-space limited and I just would like to add some simple ADC or DAC tiles at the bottom. I know, it might not be best practice, but nextpnr can succesfully route my user design and that's good enough for me.

---

I understand why FABulous introduced the exception for the top and bottom row, but I strongly believe that it creates more issues than it solves. The current exception is not a generalised solution (one could even call it a *hack*). It already cost me a lot of time to work around a lot of issues caused by it. It needlessly complicates many of the subsystems in FABulous such as the fabric description, fabric generation, bitstream generation and even the configuration FSM. Exceptions have to be made everywhere just because these two rows **must not** have configuration bits.

That's why I worked on removing these exceptions everywhere, making sure the fabric is still generated correctly, and in turn making the code easier to read and therefore making it easier for people to contribute to FABulous.

With these changes users have the *choice* of whether they want any config bits in those rows or not. If they don't use any config bits in those rows, that's fine too. Just implement a loopback and the physical macros can be very small.

*Hopefully soon* can be a long time, and there are issues caused by the current implementation right now, so I urge you too reconsider these changes that improve the current situation.

If you like, we can have a short call where I can show you my current fabric and explain in detail why these changes are important to me and for the FABulous project.
Leo

EverythingElseWasAlreadyTaken:
Hey @mole99,
Thanks for your work! 
I haven't been involved in that topic, but I followed the discussion about it and I also think this is the right way to go!

Just wanted to mention, that we also should add a note to the README about the legacy mode and also some documentation about it.

We might also need to update the docs regarding these changes, but this does not need to happen in this PR. 
I think most of the stuff you touched was not really documented in the first place, but we should at least add an Issue to keep track of blind spots in the docs.

mole99:
@EverythingElseWasAlreadyTaken thanks, I appreciate it!

I've added a note about legacy bitstream generation to the README  (and removed an outdated mention of `npnr`).
To keep track of any needed documentation, I opened an issue here: #323

FABulous/fabric_files/FABulous_project_template_verilog/Fabric/ConfigFSM.v
IAmMarcelJung: @mole99 did you check if still persists? I don't know if this is an issue at all, but since it differs from the previous behavior I would either like to have this "fixed" or acknowledged by someone that this does not cause a problem.

IAmMarcelJung:
So I looked and it and tried it out and everything looks good :)  Except for that one thing I mentioned in the comment, which still needs to be resolved or acknowledged.
Special thanks for spotting and removing the deprecated `npnr` flag in the README!

mole99:
Hi @IAmMarcelJung, sorry for my delayed reply!

I had a closer look at the top-level integration and found an issue with the current implementation: I didn't realize that the top-level only generates `Frame_Data_Reg`s for all rows except the top and bottom one, those would always receive `0x12345678` for their FrameData. This would prevent the config bits for the top and bottom rows from actually reaching the tiles.
So I fixed that up, ran the simulation and verified that the frames are correctly loaded.

Regarding your question whether the `RowSelect` being `0x1F` at the same time as a `WriteStrobe` is an issue: that is not the case. The `WriteStrobe` is actually not responsible for writing a new value into the `Frame_Data_Reg`s, it only notifies the ConfigFSM that a new `WriteData` is available. The `Frame_Data_Reg`s only react to the `RowSelect` signal, that's why it is set to `0x1F` while waiting for new `WriteData`: `0x1F` is an invalid frame and therefore does not change anything. Here's the code snippet:

```verilog
    always @ (*) begin
        if(WriteStrobe) begin // if writing active
            RowSelect = FrameShiftState; // we write the frame
        end else begin
            RowSelect = {RowSelectWidth{1'b1}}; //otherwise, we write an invalid frame
        end
    end
```

You could prevent `FrameShiftState` from going to `0x1F` during a `WriteStrobe` if you only decrement when != 0, but this is not necessary as explained above.

All in all, I think the whole configuration logic needs a major overhaul at some point in time, as it's hard to read and hard to debug, but that's beyond the scope of this PR... :)

KelvinChung2000:
I agree. I happen to be working on it now, and it is unclear how they all work together. 

Creating a standard specification for frame-based configuration bitstream might be a good idea. Like AXI, it tells you how it should behave but not how it should be implemented.

IAmMarcelJung:
Hi @mole99, thanks, I did just briefly read what you wrote and it makes sense to me, but I will still try to take a closer look with your explanations this week :) Hope this is fine or do you need this merged asap?

mole99:
@IAmMarcelJung No hurry, this or next week sounds fine :)

IAmMarcelJung:
Hi @mole99,

so I took a closer look into how the configuration works and as far as my understanding goes it is indeed fine to have the `RowSelect` signal at `0x1F` while writing new data.

However I couldn't fully follow your explanation. You wrote
> The WriteStrobe is actually not responsible for writing a new value into the Frame_Data_Regs, it only notifies the ConfigFSM that a new WriteData is available. The Frame_Data_Regs only react to the RowSelect signal, that's why it is set to 0x1F while waiting for new WriteData: 0x1F is an invalid frame and therefore does not change anything.

This is a bit contradictory to me, since a WriteStrobe should signalize that new WriteData is available or did I get this wrong? So if both WriteStrobe is 1 and FrameShiftReg is `0x1F`, RowSelect will be set to `0x1F` for valid data too, not only during the "wait state", where it is usually `0x1F` as you pointed out, right?

Anyway, this one occurence of WriteData == `1` and RowSelect == `0x1F` just replaces what was previously the row `0`, when it had no config bits and the WriteData also had no effect. That being said, should we consider to reset the FrameShiftState to `{RowSelectWidth{1'b1}}` instead of `0`, since otherwise the bitstream header is written into row `0`? This should not really be an issue since the data should be overwritten again during configuration, but just from a logical point of view it might make sense to reset to this invalid frame state. What do you think?

I hope my explanations make a bit sense and are somewhat understandable, although it was a bit hard to describe what I actually mean. Feel free to ask if anything is unclear :) But the tl;dr is that I'm now also convinced that your changes will not cause issues, if my understanding is correct!

mole99:
Hi @IAmMarcelJung, let me try to clarify!

> This is a bit contradictory to me, since a WriteStrobe should signalize that new WriteData is available or did I get this wrong?

That is almost correct. `WriteStrobe` only notifies the `ConfigFSM` that new data is available (e.g. from the bitbang or uart interface), but it is not responsible for writing to the `Frame_Data_Reg`s.

If we take a look at the `Frame_Data_Reg` module, it is clear that it is only sensitive to `RowSelect`, there is no `strobe` signal:

```Verilog
module Frame_Data_Reg (FrameData_I, FrameData_O, RowSelect, CLK);
    parameter FrameBitsPerRow = 32;
    parameter RowSelectWidth = 5;
    parameter Row = 1;
    input [FrameBitsPerRow-1:0] FrameData_I;
    output reg [FrameBitsPerRow-1:0] FrameData_O;
    input [RowSelectWidth-1:0] RowSelect;
    input CLK;

    always @ (posedge CLK) begin
        if (RowSelect==Row)
            FrameData_O <= FrameData_I;
    end//CLK
endmodule
```

If `RowSelect` matches a specific instance of `Frame_Data_Reg`, then the data is written. That's why when `FrameShiftReg` is `0x1F` no data is written, because no `Frame_Data_Reg` is selected.
 
 > That being said, should we consider to reset the FrameShiftState to {RowSelectWidth{1'b1}} instead of 0, since otherwise the bitstream header is written into row 0? This should not really be an issue since the data should be overwritten again during configuration, but just from a logical point of view it might make sense to reset to this invalid frame state. What do you think?
 
Absolutely, resetting `FrameShiftState` to `{RowSelectWidth{1'b1}}` would actually make more sense to prevent writing all of the metadata and `0xFAB0FAB1` and the bitstream header into row 0. But as you mentioned it does not matter as it will be overwritten anyways.

I wouldn't change the current implementation too much until a proper rewrite. I think it would make more sense for each `Frame_Data_Reg` to also have a dedicated `strobe` signal. Of course that's a bit more logic, but the synthesis tool can apply clock gating to those regs and it will be quite efficient. The advantage is that we don't need to rely on an invalid `Frame_Data_Reg` address and it is clear at which point *actual* data is written.

I'm currently working on an alternative config controller for my fabric, you can check it out once it's ready :)

IAmMarcelJung:
Hi @mole99,

thanks again for your clarifications! 

> If RowSelect matches a specific instance of Frame_Data_Reg, then the data is written. That's why when FrameShiftReg is 0x1F no data is written, because no Frame_Data_Reg is selected.

IAmMarcelJung:
Ups sorry, shift + enter is not creatin a new line, but actually postin the comment, i was still writing...

FABulous/fabric_files/FABulous_project_template_verilog/Fabric/ConfigFSM.v
IAmMarcelJung: This is indeed not an issue. The discussion about it can be read in the main thread.

FABulous/fabric_files/FABulous_project_template_vhdl/Fabric/ConfigFSM.vhdl
IAmMarcelJung: This is therefore also not an issue, also see main discussion.

IAmMarcelJung (approved):

mole99:
@IAmMarcelJung I don't know the specifics on what Dirk wants to change. Perhaps the bitstream format could be overhauled too, it feels a bit awkward. At the moment the header looks as follows:

```
32-bit header for demo fabric:

| 5-bit column select | ... unused ... | 20th bit sync | 20-bit frame strobe |
```

The column select is the width of `frameSelectWidth`, which is currently hardcoded to 5, meaning that fabrics can only every be 32 columns wide. This can easily be fixed by calculating `frameSelectWidth` based on `numberOfColumns`. The 20-bit frame strobe can be customized with `maxFramesPerCol`, but the sync bit is currently hardcoded to bit 20. Also, support for initialising BRAMs would be great. So maybe it could be reworked with future extensions/features in mind.

But that's for another time ;)

---

It would be great if these changes could be included in the next release since my fabric builds on them. Or do you mean we simply merge this after #324? That's fine for me, just let me know if I need to rebase this PR.

Thanks a lot for your extensive code reviews!

mole99:
@KelvinChung2000 Alright, that sounds fine! Btw. could you rename the development branch to just `dev` afterwards? That's more common with software projects.

I've added a few comments to #325 👍

mole99:
Hi @KelvinChung2000, just wanted to ask when this is going to get merged :)

FABulous/fabric_cad/bit_gen.py
```diff
@@ -181,6 +192,10 @@ def genBitstream(fasmFile: str, specFile: str, bitstreamFile: str):
             bitStr += bitstring_to_bytes(frame_select_temp)
             bitStr += bit_array[i][j]
 
+    # Add desync frame
+    # 20th bit is desync flag
+    bitStr += bytes.fromhex("00100000")
+
```
IAmMarcelJung: @KelvinChung2000 @mole99 should we maybe move this to a separate PR since it's unrelated to the other changes and also fixes the upload for old fabrics? What do you think?

mole99: If you plan to release a version of FABulous before this merge, then yes it might make sense to "backport" this change.

mole99:
Hi @KelvinChung2000, did you have time to look into setting the switch automatically depending on whether the first or last row of the fabric contains configuration bits? Thanks!

KelvinChung2000:
I have added the code for auto-switching by adding a new entry to the specData. More of a temporary fix before we move on to something more robust, and less data re-parsing. Just do a rebase. Once our CI passes, we can merge this.

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -2509,6 +2510,14 @@ def generateBitsStreamSpec(self) -> Dict[str, Dict]:
 
         specData["TileMap"] = tileMap
         configMemList: List[ConfigMem] = []
+
+        # Temporary fix for legacy support
```
IAmMarcelJung: If this is just temporary, we should maybe add a TODO or NOTE here.

FABulous/fabric_cad/bit_gen.py
```diff
@@ -116,8 +116,7 @@ def genBitstream(fasmFile: str, specFile: str, bitstreamFile: str):
         vhdl_str += '";\n'
     vhdl_str += "end package emulate_bitstream;"
 
-    # Legacy mode does not output configuration bits
-    # for the first and last row
+    # Legacy mode does not output configuration bit for the first and last row
```
IAmMarcelJung:
```suggestion
    # Legacy mode does not output configuration bits for the first and last row
```
Seems like the 's' got knocked off

IAmMarcelJung:
If the legacy path is taken in [here](https://github.com/mole99/FABulous/blob/fix-bitstream/FABulous/fabric_cad/bit_gen.py#L120-L126), the simulations runs through without errors. Maybe the new approach has to be supported also at another step?

KelvinChung2000:
> Would it be possible to add the FABulous version to the bitstream spec? Then we could easily disable the switch for all new fabrics.
This could also be useful for other features in the future.

Actually, we NEED a version check based toggle, since old "new fabric" is not the same as new "new fabric"

PR #291, opened 2025-01-23
Add the new synthesis options

KelvinChung2000:
Implemented what #284 is looking for. All the available synthesis options are available from the CLI. The options are only available to the synthesis command. 

I have done some testing. Hopefully, all the options work...

PR #313, opened 2025-02-03
Fixing changes after #291

KelvinChung2000:
When entering the shell, now will move the current working directory to the project directory, and when exit will move the user back to the original point. 

This will make all relative paths always work relative to the current FABulous project. 

Some other minor migration to `pathlib`.

Typo fix.

### DIFF
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -125,7 +125,10 @@ def main():
 
     setup_global_env_vars(args)
 
-    projectDir = Path(args.project_dir).absolute()
+    if p := os.getenv("FAB_PROJ_DIR"):
+        projectDir = Path(p).absolute()
+    else:
+        projectDir = Path(args.project_dir).absolute()
 
     args.top = projectDir.stem
 
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -127,21 +127,21 @@ def main():
 
     projectDir = Path(args.project_dir).absolute()
 
-    if not projectDir.exists():
-        logger.error(f"The directory provided does not exist: {projectDir}")
-        exit(-1)
-
     args.top = projectDir.stem
 
     if args.createProject:
         create_project(projectDir, args.writer)
         exit(0)
 
+    if not projectDir.exists():
+        logger.error(f"The directory provided does not exist: {projectDir}")
+        exit(1)
+
     if not (projectDir / ".FABulous").exists():
         logger.error(
             "The directory provided is not a FABulous project as it does not have a .FABulous folder"
         )
-        exit(-1)
+        exit(1)
     else:
         setup_project_env_vars(args)
 
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -125,13 +125,19 @@ def main():
 
     setup_global_env_vars(args)
 
-    args.top = os.getenv("FAB_PROJ_DIR").split("/")[-1]
+    projectDir = Path(args.project_dir).absolute()
+
+    if not projectDir.exists():
+        logger.error(f"The directory provided does not exist: {projectDir}")
+        exit(-1)
+
+    args.top = projectDir.stem
 
     if args.createProject:
-        create_project(os.getenv("FAB_PROJ_DIR"), args.writer)
+        create_project(projectDir, args.writer)
         exit(0)
 
-    if not os.path.exists(f"{os.getenv('FAB_PROJ_DIR')}/.FABulous"):
+    if not (projectDir / ".FABulous").exists():
         logger.error(
             "The directory provided is not a FABulous project as it does not have a .FABulous folder"
         )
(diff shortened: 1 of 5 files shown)


FABulous/FABulous.py
```diff
@@ -125,23 +125,30 @@ def main():
 
     setup_global_env_vars(args)
 
-    args.top = os.getenv("FAB_PROJ_DIR").split("/")[-1]
+    projectDir = Path(args.project_dir).absolute()
```
EverythingElseWasAlreadyTaken: Could we stick to the env var? Otherwise, this would brick the possibility to set the project dir via .env files.

KelvinChung2000: I have updated, so it take from the .env than fall back to args

FABulous/FABulous.py
```diff
@@ -125,23 +125,33 @@ def main():
 
     setup_global_env_vars(args)
 
-    args.top = os.getenv("FAB_PROJ_DIR").split("/")[-1]
+    if p := os.getenv("FAB_PROJ_DIR"):
```
EverythingElseWasAlreadyTaken: You can also provide a default value for os.getenv("FAB_PROJ_DIR", args.projectDir)

But FAB_PROJ_DIR should be always set by the .env bootstrap functions:
https://github.com/FPGA-Research/FABulous/blob/FABulous2.0-development/FABulous/FABulous_CLI/helper.py#L34

FABulous/FABulous.py
```diff
@@ -125,23 +125,33 @@ def main():
...
+    if not projectDir.exists():
+        logger.error(f"The directory provided does not exist: {projectDir}")
+        exit(1)
+
+    if not (projectDir / ".FABulous").exists():
         logger.error(
             "The directory provided is not a FABulous project as it does not have a .FABulous folder"
         )
-        exit(-1)
+        exit(1)
```
EverythingElseWasAlreadyTaken, line 144: Also same here, if it's an error, we should throw one.

FABulous/FABulous_CLI/FABulous_CLI.py
```diff
@@ -271,6 +277,7 @@ def do_load_fabric(self, args):
                 logger.error(
                     "No argument is given and the csv file is set or the file does not exist"
                 )
+                return
```
EverythingElseWasAlreadyTaken, line 280: Should we throw an error here?

FABulous/FABulous_CLI/cmd_synthesis.py
```diff
@@ -246,7 +245,7 @@ def do_synthesis(self, args):
         if resolvePath.exists():
             paths.append(resolvePath)
         else:
-            logger.error(f"{resolvePath} does not exits")
+            logger.error(f"{resolvePath} does not exists")
             return
```
EverythingElseWasAlreadyTaken, line 249: Also here?

FABulous/FABulous.py
```diff
@@ -125,23 +125,33 @@ def main():
...
 
     if args.createProject:
-        create_project(os.getenv("FAB_PROJ_DIR"), args.writer)
+        create_project(projectDir, args.writer)
         exit(0)
 
-    if not os.path.exists(f"{os.getenv('FAB_PROJ_DIR')}/.FABulous"):
+    if not projectDir.exists():
+        logger.error(f"The directory provided does not exist: {projectDir}")
+        exit(1)
```
EverythingElseWasAlreadyTaken, line 138: If this is an error, so exit(-1) would be better or throwing an actual error.

EverythingElseWasAlreadyTaken (commented):
Looks good to me, just a few small comments, mostly regarding errors. :)

FABulous/FABulous.py
KelvinChung2000: I notice that, but this makes the type checker happy.

FABulous/FABulous.py
KelvinChung2000: I think we should use `exit(1)` for error since it is UNIX convention.

FABulous/FABulous_CLI/FABulous_CLI.py
KelvinChung2000: The return will let the user re-enter the command.

FABulous/FABulous.py
EverythingElseWasAlreadyTaken: Sorry, you are right, I've mixed it up.

FABulous/FABulous_CLI/FABulous_CLI.py
EverythingElseWasAlreadyTaken: Ok, then we should keep it that way. :)

FABulous/FABulous.py
EverythingElseWasAlreadyTaken: Okay, then let's keep it that way.

PR #317, opened 2025-02-06
Fix #316

### LINKED ISSUE #316: load_fabric command broken on dev branch
Since #313, the `load_fabric` command seems to be broken. I get the following error:

`ERROR | No argument is given and the csv file is set or the file does not exist`

This should have been caught by the CI testing and I will open a separate issue for this.

KelvinChung2000:
Make `prjoectDir` always absolute after env is loaded

### DIFF
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -125,7 +125,7 @@ def main():
 
     setup_global_env_vars(args)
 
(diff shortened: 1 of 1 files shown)


IAmMarcelJung (approved):
Thanks for this! The issue is now fixed

PR #328, opened 2025-02-18
feat: Add version number and checks

KelvinChung2000:
Add the package version when creating a project. If the project version is less than the starting version, add a warning during package setup. 

It can further improve to raise errors when the project version and the package version have major version differences. But this can be done later.

### DIFF
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -221,7 +221,7 @@ def main():
             )
             exit(1)
 
-    project_version = Version(os.getenv("VERSION", "1.0.0"))
+    project_version = Version(os.getenv("FAB_PROJ_VERSION", "1.0.0"))
     package_version = Version(version("FABulous-FPGA"))
     if package_version < project_version:
         logger.error(
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -1,8 +1,10 @@
 import argparse
 import os
+from importlib.metadata import version
 from pathlib import Path
 
 from loguru import logger
+from packaging.version import Version
 
 from FABulous.FABulous_CLI.FABulous_CLI import FABulous_CLI
 from FABulous.FABulous_CLI.helper import (
@@ -11,6 +13,7 @@ from FABulous.FABulous_CLI.helper import (
     setup_global_env_vars,
     setup_logger,
     setup_project_env_vars,
+    update_project_version,
 )
 
 
@@ -162,6 +165,18 @@ def main():
 
     parser.add_argument("--debug", action="store_true", help="Enable debug mode")
 
+    parser.add_argument(
+        "--version",
+        action="version",
+        version=f"FABulous CLI {Version(version('FABulous-FPGA')).base_version}",
+    )
+
+    parser.add_argument(
+        "--update-project-version",
+        action="store_true",
+        help="Update the project version to match the FABulous package version",
+    )
+
     args = parser.parse_args()
 
     setup_logger(args.verbose, args.debug, log_file=args.log)
@@ -197,6 +212,22 @@ def main():
 
     if not projectDir.exists():
         logger.error(f"The directory provided does not exist: {projectDir}")
+        exit(1)
+
+    if args.update_project_version:
+        if not update_project_version(projectDir):
+            logger.error(
+                "Failed to update project version. Please check the logs for more details."
+            )
+            exit(1)
+
+    project_version = Version(os.getenv("VERSION", "0.0.0"))
+    package_version = Version(version("FABulous-FPGA"))
+    if project_version < package_version:
+        logger.error(
+            f"Version mismatch! FABulous-FPGA version: {package_version}, Project version: {project_version}\n"
+            'Please run "FABulous <project_dir> --update-project-version" to update the project version.'
+        )
         exit(1)
 
     fab_CLI = FABulous_CLI(
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -225,10 +225,15 @@ def main():
     package_version = Version(version("FABulous-FPGA"))
     if project_version < package_version:
         logger.error(
-            f"Version mismatch! FABulous-FPGA version: {package_version}, Project version: {project_version}\n"
-            'Please run "FABulous project_dir --update-project-version" to update the project version.'
+            f"Version incompatible! FABulous-FPGA version: {package_version}, Project version: {project_version}\n"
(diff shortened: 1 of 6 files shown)


IAmMarcelJung (approved):
Cool, thanks for this! :) The fabulous version that is written to the file is a bit weird (it shows `0.0.post754` for me), but I guess this due to the current python setup?

KelvinChung2000:
I am showing `1.0.post157+dirty`. Maybe I am on a more recent install?

Let's delay the merge until we do the release. I don't want to rebase again.

IAmMarcelJung:
Okay then probably I did something wrong although I just cloned your fork and checked out your `version-check` branch and did the usual `pip install -r requirements.txt` and `pip install -e .` installaton process in a fresh `venv`. 

Delaying the merge might make sense as I think you put a lot of effort into the rebase and this should indeed not be done too often :)

KelvinChung2000:
This could be due to the branch checkout. But ultimately, this should not matter.

IAmMarcelJung:
Okay so you also implemented the change in the master? Or what do you mean? As you said it does not matter but I personally would like to understand what happens and if I probably did something wrong.

KelvinChung2000:
I implemented the change in the development. The installation will depend on the `pyproject.toml`, and the version is automatically decided by it.

EverythingElseWasAlreadyTaken:
> Okay so you also implemented the change in the master? Or what do you mean? As you said it does not matter but I personally would like to understand what happens and if I probably did something wrong.

You did nothing wrong, for a clean venv with Kelvins branch, it should be `0.0.post754`, I guess. At least its also the same for me.  This is just how the versioning with seuptools is configured and when you did the `pip install -e .` for your current venv. 

The version Kelvin showed, seems to have some artifacts from whenever he installed FABulous in his venv. 

It's all based on the git history. The first numbers `0.0` shows the last tags in the git history and for me, there were no tags on Kelvins repo, this is 0.0. The  `post754` is the number of commits after the latest tag. The `+dirty` means, that there were uncommitted changes wile the installation. 

Here can you find more information about versioning: https://setuptools-git-versioning.readthedocs.io/en/stable/schemas/tag/dirty_version.html

FABulous/FABulous_CLI/helper.py
```diff
@@ -131,6 +132,12 @@ def setup_project_env_vars(args: argparse.Namespace) -> None:
         )
         os.environ["FAB_PROJ_LANG"] = args.writer
 
+    # string comparison
+    if os.environ["VERSION"] < version("FABulous-FPGA"):
```
EverythingElseWasAlreadyTaken: I think it's a bit too much simplified, if we just check if project version is not newer than your current fabulous version. Older fabulous versions can't handle newer projects as well.

I'm also not sure, if we should check for the whole version string, technically breaking changes should only happen in major releases. 

 I think we should at least not throw warnings for patch, and dirty versions.

EverythingElseWasAlreadyTaken (approved):
This is a really nice improvement! 

We should discuss, how we treat the versioning for the checks.

FABulous/FABulous_CLI/helper.py
KelvinChung2000: I will make the changes once we decide what we want for versioning. Probably using something like [packaging](https://pypi.org/project/packaging/) should allow us to do more fine control.

FABulous/FABulous.py
```diff
@@ -199,6 +214,27 @@ def main():
...
+    if args.update_project_version:
+        if not update_project_version(projectDir):
+            logger.error(
+                "Failed to update project version. Please check the logs for more details."
+            )
+            exit(1)
+
+    project_version = Version(os.getenv("VERSION", "0.0.0"))
+    package_version = Version(version("FABulous-FPGA"))
+    if project_version < package_version:
```
EverythingElseWasAlreadyTaken: Shouldn't it be the other way around? 

If the project version is higher than the package version, it could contain features that are not supported by the older FABulous version. 
But if the package is newer than the package version, we should still have backward compatibility.

FABulous/FABulous.py
```diff
@@ -199,6 +214,27 @@ def main():
...
+        logger.error(
+            f"Version incompatible! FABulous-FPGA version: {package_version}, Project version: {project_version}\n"
+            r'Please run "FABulous \<project_dir\> --update-project-version" to update the project version.'
+        )
+        exit(1)
+    if project_version.major != package_version.major:
```
EverythingElseWasAlreadyTaken, line 232: This will break support for legacy projects, since they have no VERSION set, which defaults to 0.0.0.

FABulous/FABulous_CLI/helper.py
```diff
@@ -500,3 +503,24 @@ def install_oss_cad_suite(destination_folder: Path, update: bool = False):
     os.environ["PATH"] += os.pathsep + str(ocs_folder / "bin")
 
     logger.info("OSS CAD Suite setup completed successfully.")
+
+
+def update_project_version(project_dir: Path) -> bool:
```
EverythingElseWasAlreadyTaken, line 509: Simply updating the project version could cause confusion, since you can't trust the project version anymore and can't identify from which version the project is actually derived from. 

I'd suggest we add another version tag to the .env. 
Something like FAB_PROJ_VERSION_CREATED, which should never be edited, to ensure that we always can track which version of the demo project this was.
And then we have FAB_PROJ_VERSION_EDIT, which is the FABulous Version, the project was edited last.

FABulous/FABulous_CLI/helper.py
```diff
@@ -226,6 +228,7 @@ def create_project(project_dir: Path, lang: Literal["verilog", "vhdl"] = "verilo
 
     with open(os.path.join(project_dir, ".FABulous/.env"), "w") as env_file:
         env_file.write(f"FAB_PROJ_LANG={lang}\n")
+        env_file.write(f"VERSION={version('FABulous-FPGA')}")
```
EverythingElseWasAlreadyTaken: We should name it something like `FAB_PROJ_VERSION`, to identify, that is really about the project.

FABulous/FABulous_CLI/helper.py
KelvinChung2000: My intuition is that later, we will place project migration logic in here. But yes, we will still want created version and edit

FABulous/FABulous.py
KelvinChung2000: I will assume project always start at 1.0.0

PR #334, opened 2025-02-27
feat: Add isort and autoflake pre-commit-hooks 

EverythingElseWasAlreadyTaken:
Add isort pre-commit-hooks to automatically sort imports.
Add autoflake pre-commit-hook to remove unused imports. 

Run isort and autoflake to sort and remove unused imports.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,3 +50,9 @@ repos:
     #     args:
     #         - --use-current-year
     #         - --no-extra-eol                # see below
+
+-   repo: https://github.com/PyCQA/autoflake
+    rev: v2.3.1
+    hooks:
+    -   id: autoflake
+        args: [--remove-all-unused-imports, --ignore-pass-after-docstring, --in-place]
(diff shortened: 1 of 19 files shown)


KelvinChung2000:
If we use `isort` and `autoflake`, should we just replace our whole lint and format tool to `ruff`?

EverythingElseWasAlreadyTaken:
> If we use `isort` and `autoflake`, should we just replace our whole lint and format tool to `ruff`?

Could be also good, and it would be just one tool instead of 3.

KelvinChung2000:
Let's do that then. One thing to consider is, do we want more line width? `black` is 88, and we can set it to 120.

EverythingElseWasAlreadyTaken:
> Let's do that then. One thing to consider is, do we want more line width? `black` is 88, and we can set it to 120.

Yes, I'll set it up tomorrow. 
For me, 88 works well.
If you want to increase the line width, I think 100 is max, to avoid the need of horizontal scrolling when working with vertical split screens.

KelvinChung2000:
I guess that depends on monitor size and the amount of pixels available. You can decide what would be more suitable.

EverythingElseWasAlreadyTaken:
> I guess that depends on monitor size and the amount of pixels available. You can decide what would be more suitable.

Sure, it depends on the monitor size/resolution and your personal settings.

I'll do a new PR for ruff.

PR #335, opened 2025-02-28
feat: Replace Black with Ruff for code linting and formatting 

EverythingElseWasAlreadyTaken:
Replace Black with Ruff for code linting and formatting.
 Run Ruff formatter and apply linting fixes. 
 Remove some unneeded code, that I found while fixing linter issues. 
 Replace the Black workflow with pre-commit workflow. 
Update README.


ruff.toml
```diff
@@ -0,0 +1,77 @@
...
+dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
+
+[format]
+# Like Black, use double quotes for strings.
+quote-style = "double"
+
+# Like Black, indent with spaces, rather than tabs.
+indent-style = "space"
+
+# Like Black, respect magic trailing commas.
```
KelvinChung2000, line 105: Just minor comment mismatch.

ruff.toml
```diff
@@ -0,0 +1,77 @@
...
+skip-magic-trailing-comma = false
+
+# Like Black, automatically detect the appropriate line ending.
+line-ending = "auto"
+
+# Enable auto-formatting of code examples in docstrings. Markdown,
+# reStructuredText code/literal blocks and doctests are all supported.
+#
+# This is currently disabled by default, but it is planned for this
+# to be opt-out in the future.
```
KelvinChung2000, line 115: same here

KelvinChung2000 (approved):

ruff.toml
EverythingElseWasAlreadyTaken: > Just minor comment mismatch.

The naming of this parameter is a bit ambiguous, but skip-magic-trailing-comma=false, means that it's respected.
https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma

ruff.toml
EverythingElseWasAlreadyTaken: Oh, this should be fixed. It is still left true from testing.  I set it to false again. 
We still format docstrings with docformatter, since ruff only can format docstring-code lines and not the whole docsting.

KelvinChung2000:
Something came to my mind: variable naming. Should we also enforce that?

EverythingElseWasAlreadyTaken:
> Something came to my mind: variable naming. Should we also enforce that?

I also thought about this, but it will be a mess to fix in the first place, since there is really no constituency currently.

IAmMarcelJung:
> Something came to my mind: variable naming. Should we also enforce that?

At some point we should do this, but yeah, probably not now.

EverythingElseWasAlreadyTaken:
I've added a lot more rules to the ruff.toml, which were kinda easy to apply and make sense to us. 
I've also added a list of rules we should apply in the future, but these mostly need a lot of manual changes, which just takes time. I'll open an issue for this, after this PR is merged.

EverythingElseWasAlreadyTaken:
Rebase on #384

PR #336, opened 2025-02-28
fix: Unify project templates and remove duplicate files

### LINKED ISSUE #155: Merge Verilog and VHDL project templates
To ease maintainability, it would make sense to merge `FABulous_project_template_verilog` and `FABulous_project_template_vhdl` into a single `FABulous_project_template`.

Many files in both directories are duplicated and some files/filenames already diverge.

My idea is that after merging #154, the logic for reading the tile configuration can be changed a bit:

Instead of specifying file extensions for HDL sources in `BEL` entries inside tile configs, omit them:

```
BEL,./InPass4_frame_config_mux,RAM2FAB_D0_
```

Depending on whether a Verilog/VHDL writer is used, append the correct file ending (`.v`, `.vhdl`). Since both Verilog and VHDL files are in the tile directory, the correct one can be picked up.

What do you think? Am I missing something?

EverythingElseWasAlreadyTaken:
This PR unifys the project templates and adds a common project template, which holds all duplicate files.

- Remove CPU_IO from VHDL project template, since it is unused and mostly a copy of RAM_IO.
- Similar to the Verilog project template, add includes to the .csv and .list files for the VHDL project template.
- Add a common folder for duplicate files in project templates.
-  Add a HDL_SUFFIX parameter to common tile .csv files, which will be replaced while project generation.
-  Rename some of the VHDL files, to match the naming of the Verilog files.
- Fix some file paths and rename some modules to unify the projects templates.
- Update create_project function for new template structure

Fixes #155


KelvinChung2000 (commented):
Looking fine to me. I have been thinking about changing FABulous's workflow recently. The `create_project` does not copy in the files as templates, but generates an empty file structure. Then we provide a CLI command to add Bels and pre-defined tile to the fabric. The current "template" will become a fully assembled example project.  And the user can choose to "start" from the example project. 

But what I am thinking will require a lot of CLI upgrades, so it's probably something for the future. Maybe we can put it out as an enhancement.

EverythingElseWasAlreadyTaken:
> Looking fine to me. I have been thinking about changing FABulous's workflow recently. The `create_project` does not copy in the files as templates, but generates an empty file structure. Then we provide a CLI command to add Bels and pre-defined tile to the fabric. The current "template" will become a fully assembled example project.  And the user can choose to "start" from the example project. 
> 
> But what I am thinking will require a lot of CLI upgrades, so it's probably something for the future. Maybe we can put it out as an enhancement. 

I think this would be a good direction, but I don't like the idea of doing everything through the shell. I would rather like to have something like a configuration file, where you can describe your project and pass to FABulous. FABulous will then assemble your project and just pulls in just the stuff it needs. We could still provide some CLI commands, that helps you build the config file. 

Also I would really like to make the templates an own project/repo with own revisioning, that can be forked and customized by users, without having to deal with FABulous itself.

Something like a FABulous/Tile Library, that  could contain our default bels, as well as the hardened macros for a bunch of open PDKs. 

And you are right, this will require a lot of work, but I think we are going in a good direction. :)

mole99:
I'd love to see what @EverythingElseWasAlreadyTaken is describing: a standalone tile library that can be assembled into the final fabric. But FABulous would be able to generate the tiles without any fabric at all, so you could already harden the tiles for different PDKs and store them in a repository before stitching them together.
Users could host their own tile libraries that supplement the official FABulous tile library 🙌

PR #314, opened 2025-02-04
Fix config bit after #242

KelvinChung2000:
See #248

PR #342, opened 2025-03-19
fix:fabric_gen: Wrong number of config bits in switch matrix

EverythingElseWasAlreadyTaken:
PR #314 fixed the config bit calculation for non power of two multiplexers, but the calculation of 'NoConfigBit' parameter in the switch matrix was not updated accordingly. This resulted in port mismatches in the switch matrix instantiation.


KelvinChung2000 (approved):

EverythingElseWasAlreadyTaken:
@KelvinChung2000 I've updated this PR.

Would be good to merge this, since it fixes an error with non power of two muxes.

PR #344, opened 2025-03-25
doc: minor phrasing change

TaoBi22:
This is a very tiny change but PRd as I don't have write access after the organization migration


TaoBi22:
Closing since master is on ice as per discussion in #345

PR #347, opened 2025-03-30
feat: Install OSS-CAD-Suite

EverythingElseWasAlreadyTaken:
OSS-CAD-Suite can be installed via CLI command or FABulous parameter.

Add OSS-CAD-Suite env variable.


FABulous/FABulous_CLI/FABulous_CLI.py
```diff
@@ -253,6 +253,38 @@ def do_exit(self, *ignored):
...
+    install_oss_cad_suite_parser.add_argument(
+        "destination_folder",
+        type=Path,
+        help="Destination folder for the installation",
+        default="",
+        completer=Cmd.path_complete,
+        nargs=argparse.OPTIONAL,
+    )
+    install_oss_cad_suite_parser.add_argument(
+        "update_existing",
```
KelvinChung2000: I think using `--update` is more natural. Can use `action="store_true"` which make it a flag option.

KelvinChung2000 (commented):
One thing to consider is putting the CLI install command into a separate file. 

We may want to store commands by category, i.e., one file per command category or even a folder per category, then command per file. But this depends on how much we are expanding the CLI.

FABulous/FABulous_CLI/FABulous_CLI.py
EverythingElseWasAlreadyTaken: Good idea, I've implemented it now similar to the create project flag and made both flags mutual exclusive  :)

EverythingElseWasAlreadyTaken:
> One thing to consider is putting the CLI install command into a separate file.
> 
> We may want to store commands by category, i.e., one file per command category or even a folder per category, then command per file. But this depends on how much we are expanding the CLI.

I think for now it's ok how it is.
But if we start to expand the CLI more, then we could think about dividing this down to different files.

Issue #245, opened 2024-10-29
Difference in bitstream generation between master and dev branch


IAmMarcelJung:
Hi everyone,

when playing around with the MPW-5 FABulous chips, I noticed that the bitstream generated by the current version of the `dev` branch differs from the one created by the `master` branch. Also, the bitstream created by the `dev` branch version does not work as expected. I cloned the repo to a new location to make sure I have the most recent version of the branches and no old leftovers or anything. I also used the default user design to make sure I didn't make any changes. Still there could of course be some other configuration issue on my end.

Therefore I would like to know if anyone can either deny or confirm behaviour I'm experiencing.

Thanks in advance!

KelvinChung2000:
I suggest checking the generated bitstream spec file. The master and dev should match. The `bitStreamSpec.bin` is a pickle, so you can load that into Python to compare. 

But this is relatively strange since the simulation is passing.

IAmMarcelJung:
Both `bitStreamSpec.bin` and `bitStreamSpec.csv` do not match on the two branches (which makes sense since they are created from the same data). I looked at the differences in `bitStreamSpec.csv`, but there were a lot of them. The simulation is also executed in the CI pipeline, right?

KelvinChung2000:
The simulation is executed in the CI pipeline. 

If you create two of the same fabrics, one with the master and one with the dev, the key-value pair of the pickled dict should match. The CSV is less important since it is for human reading. The `bitStreamSpec.bin` is used in the simulation to generate the bitstream.

IAmMarcelJung:
>  the key-value pair of the pickled dict should match
So it is okay that the files are not exactly the same? Only the pairs have to match?

KelvinChung2000:
> > the key-value pair of the pickled dict should match
> 
> So it is okay that the files are not exactly the same? Only the pairs have to match? Which pairs do you mean exactly?

Yes, the order of the entry might change, but the key-value pair should match if I remember correctly. 



> I executed the simulation for both cases and also found out that both are working, although the `bitstream.hex` file and therefore the `SelfWriteData` also differs. Maybe the design is simple enough so there is no effect from the difference?

I assume they should have the same configuration logic. Is the order of bitstream different? like little endian and big endian? I am not sure about the details of the bitstream and configuration.

IAmMarcelJung:
>  Is the order of bitstream different?

No the order is the same for the simulation counter, only some bytes differ (20-30). 

[...]

PR #352, opened 2025-04-10
Fix: Bitstream incompatibility between dev and main

KelvinChung2000:
Parsing using yosys leads to changes in the fabric netlist, which leads to incompatable bitstream. 

Updated to use the port name to look for the attribute from the netlist instead of directly finding the attribute from the netlist. 

I have checked against the 1.0 LUT4AB tile. Now `LA_I0_input` mux is starting at config bit 112 instead of 113. And `LA_EN` at bit 121.  

Let me know if this does not fully fix the problem.

### DIFF
--- a/FABulous/fabric_generator/file_parser.py
+++ b/FABulous/fabric_generator/file_parser.py
@@ -871,7 +871,7 @@ def parseBelFile(
             data_dict = json.load(f)
 
         modules = data_dict.get("modules", {})
-        filtered_ports: dict[str, IO] = {}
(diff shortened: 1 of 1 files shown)


IAmMarcelJung:
This is probably related to #245 and could also fix it?

EverythingElseWasAlreadyTaken (approved):
Looks good!
No more changes in bitstream spec, bels.txt and pips.txt compared to current FABulous master.
I've checked Verilog and VHDL.

PR #354, opened 2025-04-11
fix:fabric_gen: Bistream spec gen for non power of 2 MUX

EverythingElseWasAlreadyTaken:
The bitstream spec generation was wrong for non power of 2 MUX. The number of bits to encode a MUX was calculated by (muxsize.bit_length() -1). As an example, this results 3 config bits for a MUX15 and 4 bits for a MUX17, which is not correct.

We changed this to (muxsize - 1).bit_length() for the MUX encoding calculation. This results in the right number of config bits: MUX15: 4 bits, MUX16: 4 bits MUX17: 5 bits...


KelvinChung2000:
With #352 is everything working now?

EverythingElseWasAlreadyTaken:
> With #352 is everything working now?

I hope so. 
Did not run in any issues yet.

Issue #357, opened 2025-05-11
Problems with changing the demo (number of W_IO)


ninachloekassi:
I am trying to adjust the demo project to an eFPGA with 74 to 106 IOs.

## The problem

When running the fabulous_tb testbench, I_top and T_top are marked unitialized, as shown below:

```
fabric(I_top) = 0xxxxxxxxxxxxxxxxxxxxxxxxxxxx gold = 0x000000003fff400000000000000, fabric(T_top) = 0x000ffffffffxxxxXxxxxxxxxxxx gold = 0x0000000000000000000ffffffff
fabric(I_top) = 0xxxxxxxxxxxxxxxxxxxxxxxxxxxx gold = 0x000000003fff800000000000000, fabric(T_top) = 0x000ffffffffxxxxXxxxxxxxxxxx gold = 0x0000000000000000000ffffffff
fabric(I_top) = 0xxxxxxxxxxxxxxxxxxxxxxxxxxxx gold = 0x000000003fffc00000000000000, fabric(T_top) = 0x000ffffffffxxxxXxxxxxxxxxxx gold = 0x0000000000000000000ffffffff
```

## Attempts to solving

- Changed wire/reg sizes in fabulous.tb
- Added new W_IO tiles to top_wrapper (Test + user_design)
- Changed the test designs
- Changed max bitstream length to ~32k
- Change number of rows with NULL instead of W_IO

The problem comes when changing the amount of W_IO tiles.

Thanks in advance for any help

## Fabric

```csv
FabricBegin,,,,,,,,,,,,,,,,,,
NULL,N_term_single,N_term_single,NULL,,,,,,,,#,,,,,,,
W_IO,LUT4AB,LUT4AB,NULL,,,,,#,180,,,,,,
W_IO,LUT4AB,LUT4AB,NULL,,,,,#,180,,,,,,
W_IO,LUT4AB,LUT4AB,NULL,,,,,#,180,,,,,,
W_IO,LUT4AB,LUT4AB,NULL,,,,,#,180,,,,,,
W_IO,LUT4AB,LUT4AB,NULL,,,,,#,180,,,,,,
W_IO,LUT4AB,LUT4AB,NULL,,,,,#,180,,,,,,
W_IO,LUT4AB,LUT4AB,NULL,,,,,#,180,,,,,,
W_IO,LUT4AB,LUT4AB,NULL,,,,,#,180,,,,,,
[...]

PR #359, opened 2025-05-13
Add the current fabric limitations to the documentation

### LINKED ISSUE #358: Docs: Add FABulous limitations / Tiles / Rows / Bels
More information is available here: https://github.com/FPGA-Research/FABulous/discussions/325

IAmMarcelJung:
Since we had the discussion in #325 and this now also showed up in #357, which shows that this is important information for anyone using FABulous, the current limitations of a FABulous fabric are added to the documentation. It containts mostly the information from the discussion. We could also consider using a warning note or something similar instead of dedicating a separate section. Let me know your thoughts on this and if something is missing or wrong.

Resolves #358.

### DIFF
--- a/docs/source/fabric_definition.rst
+++ b/docs/source/fabric_definition.rst
@@ -78,6 +78,28 @@ And the following block provides a tile.csv example (in this case LUT4AB.csv).
    MATRIX,   LUT4AB_switch_matrix.vhdl
    EndTILE
 
+
+Current Fabric Size Limitations
+-------------------------------
+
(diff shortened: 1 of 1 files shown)


EverythingElseWasAlreadyTaken:
Thanks for the PR!

We should also add, that we currently only support a single CLK domain.

mole99 (approved):
LGTM. I think this covers all current limitations (if you add the single clock domain Jonas mentioned).

Thanks for adding this to the documentation!

KelvinChung2000 (approved):

PR #361, opened 2025-05-22
feat: Add User Design Top Wrapper Generation

EverythingElseWasAlreadyTaken:
Add a generateUserDesignTopWrapper to dynamically generate the user design wrapper, based on the Fabric Size.
This can be extended further in the future and allows us to dynamically test different fabric sizes.

Add the generateUserDesignTopWrapper function to CLI and API.
Add User Design Category to CLI.
Add "q" as alias to exit the CLI. 

Add the generateUserDesignTopWrapper to the FABulous.tcl of the default Verilog project. 

Add ports_vectors, module_name and language attributes to Bel Class.
Extend parseBelFile function to add these attributes to Bel objects.
Currently, this works only for Verilog, since it needs the Yosys parsing.

Add a getBelsByTileXY to Fabric Class.


FABulous/fabric_definition/Bel.py
```diff
@@ -37,32 +41,41 @@ class Bel:
...
         The number of config bits of the BEL.
+    language : str
+        Language of the BEL. Currently only VHDL and Verilog are supported.
     belFeatureMap : dict[str, dict]
         The feature map of the BEL.
     withUserCLK : bool
         Whether the BEL has userCLK port. Default is False.
     individually_declared : bool
         Indicates if ports are individually declared. Default is False.
+    ports_vectors: dict[str, dict[str, (IO, int)]]
```
KelvinChung2000: I think should be `dict[str, dict[str, tuple[IO, int]]]`

FABulous/fabric_definition/Fabric.py
```diff
@@ -53,6 +55,8 @@ class Fabric:
         A dictionary of tiles used in the fabric. The key is the name of the tile and the value is the tile.
     superTileDic : dict[str, SuperTile]
         A dictionary of super tiles used in the fabric. The key is the name of the super tile and the value is the super tile.
+    getBelsByTileXY:
```
KelvinChung2000: I don't think we need to include class method here

EverythingElseWasAlreadyTaken: It felt like a cleaner solution, and it can be reused better, especially since the x/y coordinates in the tile map are swapped.
And it keeps the bound checks and so on out of the main logic.

 I think in general, we should make more use of class methods, since our code is often quite complex.
 We are doing bound checks and error handling in many places in our main logic, instead of having class methods that do these things in one place.

KelvinChung2000: I only mean for the documentation

EverythingElseWasAlreadyTaken: Ok, I can kick it out here.

PR #364, opened 2025-05-26
feat: Generate Custom Tile Config

EverythingElseWasAlreadyTaken:
Adds functionality and documentation for the Custom Tile Generation.


KelvinChung2000 (commented):
It looks good to me. I think we should use it and let it go through the testing CI. Maybe we should change the default project to use this to generate the RegFile.

Issue #319, opened 2025-02-07
Add CLI option to pass a list of FABulous commands directly from shell


EverythingElseWasAlreadyTaken:
For CI and more command line centered workflows, it would be great to run FABulous commands directly from the shell, since currently we can only specify a script to run, which is not that flexible.

We should introduce a command line parameter to execute a bunch of FABulous commands just one after another, similar to the [yosys -p / --commands flag](htttps://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd_ref.html).

EverythingElseWasAlreadyTaken:
Done with PR #365

PR #365, opened 2025-05-28
feat: Add force mode, improve exception handling, correct debug log and direct command input with testing

KelvinChung2000:
Introduce a force mode to allow commands to run despite errors, and enhance exception handling by overriding the onecmd method to catch exceptions properly.

Adjust the debug log level to only display debug logs when the debug flag is set. 

add the `-p` option for pass cmd directly into the CLI. 

Add tests for the start-up script.

Make the existing test pytest discoverable.

### DIFF
--- a/FABulous/fabric_generator/file_parser.py
+++ b/FABulous/fabric_generator/file_parser.py
@@ -138,6 +138,9 @@ def parseFabricCSV(fileName: str) -> Fabric:
         if i[0].startswith("Tile"):
             if "GENERATE" in i:
                 # import here to avoid circular import
+                from FABulous.fabric_generator.fabric_automation import (
+                    generate_custom_tile_config,
+                )
 
                 # we generate the tile right before we parse everything
                 i[1] = str(generate_custom_tile_config(filePath.joinpath(i[1])))
--- a/docs/source/Building fabric.rst
+++ b/docs/source/Building fabric.rst
@@ -16,15 +16,15 @@ completed.
 
    .. prompt:: bash
 
-        FABulous -c demo
+        (venv)$ FABulous -c demo
 
 This will create a new project named ``demo`` in the current directory.
 
 #. Running the FABulous shell
 
-   .. prompt:: bash FABulous>
+   .. prompt:: bash
 
-        FABulous demo
+        (venv)$ FABulous demo
 
 And now, we will be in the FABulous shell. After running the above command, the current working directory will be
 moved into the project directory, which is ``demo`` in this case.


KelvinChung2000:
Are you calling by parameter or calling it from inside the CLI? Because if it is called by parameter, the fix will need to be in another PR, since this will be linking to #319

KelvinChung2000:
Halting of the shell is depends on the return value of the command. If it rturn true, then the shell will be closed. We might want to unify how we handle error in the shell. Do we always return true (exit) on error, or raise an exception?

EverythingElseWasAlreadyTaken:
> Halting of the shell is depends on the return value of the command. If it rturn true, then the shell will be closed. We might want to unify how we handle error in the shell. Do we always return true (exit) on error, or raise an exception?

We should raise an exception, returning true in case of an error just feels wrong. 
And also if people start using FABulous as dependency, or it is somehow automated externally, it would cause serious headaches if it always returns true. ^^ 

As a simplification, we could automatically call the `load_fabric` command on entering the shell. 
This would reduce the needed interaction for the user.

KelvinChung2000:
I like to know what the expected behaviour is for the global environment. The current implementation is that the env variable overrides the user-supplied path, which seems a bit odd to me.

EverythingElseWasAlreadyTaken:
> I like to know what the expected behaviour is for the global environment. The current implementation is that the env variable overrides the user-supplied path, which seems a bit odd to me.

You are right, this is a bug then.
The user provided information should always override the global env.
The hierarchy should be like the following (from highest to lowest priority): 
user provided information > shell environment (e.g. `export FAB_ROOT ...`) > project .env > global .env >  default values.

KelvinChung2000:
>  Is it for all environment variables?

At least for `project_dir`. I am not sure about the rest. I only notice this when we are running the tests back to back. 

> How do I use the interactive mode?

This is what we usually have; the CLI is the interactive mode.  This basically enables force mode, so the user will never exit the shell when they encounter an error running any command. 

I will update those "super command" behaviours.

EverythingElseWasAlreadyTaken:
The error handling seems to work now!
 
But it seems, that broke the force mode, which now seems to also exits on errors.

FABulous/fabric_generator/file_parser.py
```diff
@@ -136,9 +138,6 @@ def parseFabricCSV(fileName: str) -> Fabric:
         if i[0].startswith("Tile"):
             if "GENERATE" in i:
                 # import here to avoid circular import
-                from FABulous.fabric_generator.fabric_automation import (
```
EverythingElseWasAlreadyTaken, line 139: Please add this import again, otherwise the custom tile gen will fail.

docs/source/Building fabric.rst
```diff
@@ -14,7 +14,7 @@ completed.
 
 #. Create a new project
 
-   .. prompt:: bash FABulous>
+   .. prompt:: bash
 
         FABulous -c demo
```
EverythingElseWasAlreadyTaken:
```suggestion
        (venv)$ FABulous -c demo
```

EverythingElseWasAlreadyTaken: This must have been slipped through. 

Usually we add a `(venv) $` for shell commands.

It's also wrong in the following `FABulous demo` call, but I can't make a suggestion there, please update also the following one.

EverythingElseWasAlreadyTaken (commented):
Great PR!
Looks like everything is working fine now. 

I've added a few minor comments, if these are fixed, we can merge that PR. :)

PR #372, opened 2025-06-04
fix:fabric_gen Add parseBelFile import

EverythingElseWasAlreadyTaken:
parseBelFile import was missing after #364.

Testworkflow failed silently, but this should be fixed with #366.


KelvinChung2000:
I thought I have add this back in #366, but anyway this looks fine.

PR #380, opened 2025-06-23
Add generative IOs 

EverythingElseWasAlreadyTaken:
This PR adds support for generative IOs and some smaller fixes:

[fix: Custom plib not working with VHDL Makefile](https://github.com/FPGA-Research/FABulous/commit/d2e7880af562c934da242aca3a225885655c49c4)
 - Fix syntax for yosys call with custom plib in Makefile.

[feat: Extend Code Generation](https://github.com/FPGA-Research/FABulous/commit/36686425a00b8eb0558931039a79041c703763a1)
 - Add attributes to addPortVector and addPortSalar, to allow FABulous attributes to be added to the port description.
- Add support for registers.
- Add support for inverted ports, wires and registers.
- Add support to add BelMaps to a module.

[feat: Add generative IOs](https://github.com/FPGA-Research/FABulous/commit/f9bb95d6bdaab2ad04a0b83d73ae686eedc45b40)
- Add Gen_IO feature.
- API and CLI in seperate commit. 
- Docs in separate commit.

[API/CLI: Add support for generative IOs](https://github.com/FPGA-Research/FABulous/commit/c0889c7683859a9594fd0ef9f714c96fd8e1f28e)

[docs: Add generative IO docs](https://github.com/FPGA-Research/FABulous/commit/bd743b510ed01bcbc2ef71e6f45e76272d7ff710)

[feat: Add unused tile dict to Fabric class](https://github.com/FPGA-Research/FABulous/commit/816129ad88c32ca061706cf4616b3d78b7f76eb4)
 - Unused tiles are listed in the fabric csv, but not used in the Fabric itself. It can be useful to have these, if one works tile based, and haven't defined a floorplan yet.

[fix:tile_gen: Add bel comment only if there are bels in the tile](https://github.com/FPGA-Research/FABulous/commit/5a6e0bb2ad2e71f78551e9b15b8e71d14cf21e01)


FABulous/FABulous_API.py
```diff
@@ -293,3 +293,87 @@ def generateUserDesignTopWrapper(self, userDesign: Path, topWrapper: Path):
...
+            In case of an invalid IO type for generative IOs.
+            If the number of config access ports does not match the number of config bits.
+        """
+        tile = self.fabric.getTileByName(tile_name)
+        bels: list[Bel] = []
+        if not tile:
+            logger.error(f"Tile {tile_name} not found in fabric.")
+            raise ValueError
+
+        if self.writer.__class__ is VHDLWriter:
```
KelvinChung2000: Probably use the `isinstance` method will be better.

FABulous/FABulous_API.py
```diff
@@ -293,3 +293,87 @@ def generateUserDesignTopWrapper(self, userDesign: Path, topWrapper: Path):
...
+        ------
+        ValueError
+            If tile not found in fabric.
+            In case of an invalid IO type for generative IOs.
+            If the number of config access ports does not match the number of config bits.
+        """
+        tile = self.fabric.getTileByName(tile_name)
+        bels: list[Bel] = []
+        if not tile:
+            logger.error(f"Tile {tile_name} not found in fabric.")
```
KelvinChung2000, line 324: Can use the `logger.opt(exception=ValueError()).error(....)`, then you don't need to raise bellow

FABulous/FABulous_API.py
```diff
@@ -293,3 +293,87 @@ def generateUserDesignTopWrapper(self, userDesign: Path, topWrapper: Path):
...
+        # update fabric tileDic with generated IO BELs
+        if self.fabric.tileDic.get(tile_name):
+            self.fabric.tileDic[tile_name].bels += bels
+        elif not self.fabric.unusedTileDic[tile_name].bels:
+            logger.warning(
+                f"Tile {tile_name} is not used in fabric, but defined in fabric.csv."
+            )
+            self.fabric.unusedTileDic[tile_name].bels += bels
+        else:
+            logger.error(
```
KelvinChung2000, line 352: logger.opt...

FABulous/fabric_definition/Fabric.py
```diff
@@ -252,10 +260,16 @@ def __repr__(self) -> str:
         return fabric
 
     def getTileByName(self, name: str) -> Tile | None:
-        return self.tileDic.get(name)
+        ret = self.tileDic.get(name)
+        if ret is None:
+            ret = self.unusedTileDic.get(name)
+        return ret
```
KelvinChung2000, line 270: Instead of returning `None`, I think we should raise an error.

FABulous/fabric_definition/Fabric.py
```diff
@@ -252,10 +260,16 @@ def __repr__(self) -> str:
         return fabric
 
     def getTileByName(self, name: str) -> Tile | None:
-        return self.tileDic.get(name)
+        ret = self.tileDic.get(name)
```
KelvinChung2000, line 263: I think you will need the `.get(name, None)`; otherwise, it will raise an error on the first get miss

FABulous/fabric_definition/Gen_IO.py
```diff
@@ -0,0 +1,57 @@
+from dataclasses import dataclass, field
+from FABulous.fabric_definition.define import IO
+
+
+@dataclass
+class Gen_IO:
```
KelvinChung2000, line 6: I think you can make this can inherit the Bel class, which will allow this to treat it as normal bel as well.

FABulous/fabric_definition/Tile.py
```diff
@@ -63,6 +69,10 @@ def __init__(
         for b in self.bels:
             self.globalConfigBits += b.configBit
 
+        for gio in self.gen_ios:
+            if gio.configAccess:
+                self.globalConfigBits += gio.configBit
```
KelvinChung2000: It might lead to a problem. Since if we use Gen IO after tile creation, this will not be updated. We should make the global config bit a class attribute, and compute the result every time this is used. Less efficient, but always right. 

Something like:
```python
@property
def globalConfigBits() -> int:
  configBit = 0
  configBit += sum([...]) #bel sum
  configBit += sum(...) #switch matrix sum
  return configBit
  
```

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -2642,6 +2650,293 @@ def generateBitsStreamSpec(self) -> Dict[str, Dict]:
 
         return specData
 
+    def genIOBel(self, gen_ios: list[Gen_IO], bel_name: str) -> Bel | None:
```
KelvinChung2000: Maybe move this function to a new file, and let `self` be the writer. Or we can do the move later when we are doing the file splitting.

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -2642,6 +2650,293 @@ def generateBitsStreamSpec(self) -> Dict[str, Dict]:
...
+                                ("X", sink),
+                            ]
+                            self.writer.addInstantiation(
+                                compName="cus_mux21",
+                                compInsName=f"inst_cus_mux21_{gio.prefix}{j}",
+                                portsPairs=portsPairs,
+                            )
+                        else:
+                            # generic multiplexer
+                            if self.writer.__class__ == VHDLWriter:
```
KelvinChung2000: use `isinstance()`

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -2642,6 +2650,293 @@ def generateBitsStreamSpec(self) -> Dict[str, Dict]:
...
+                            inverted=gio.inverted,
+                        )
+
+        self.writer.addNewLine()
+        self.writer.addDesignDescriptionEnd()
+        self.writer.addNewLine()
+        self.writer.writeToFile()
+
+        bel: Bel
+        if self.writer.__class__ == VHDLWriter:
```
KelvinChung2000: same here

FABulous/fabric_generator/file_parser.py
```diff
@@ -565,9 +584,77 @@ def parseTiles(fileName: Path) -> tuple[list[Tile], list[tuple[str, str]]]:
                         logger.info(f"Adding bels to custom prims file: {primsFile}")
                         addBelsToPrim(primsFile, [bels[-1]])
                 else:
-                    raise ValueError(
+                    logger.error(
```
KelvinChung2000, line 587: can use `logger.opt`

FABulous/fabric_definition/Fabric.py
```diff
@@ -252,10 +260,16 @@ def __repr__(self) -> str:
...
+        if ret is None:
+            ret = self.unusedTileDic.get(name)
+        return ret
 
     def getSuperTileByName(self, name: str) -> SuperTile | None:
-        return self.superTileDic.get(name)
+        ret = self.superTileDic.get(name)
```
KelvinChung2000, line 273: Same here

FABulous/fabric_generator/fabric_gen.py
```diff
@@ -2642,6 +2650,293 @@ def generateBitsStreamSpec(self) -> Dict[str, Dict]:
...
+                clocked = True
+            if gio.pins <= 0:
+                logger.warning(f"Generative IO {gio.prefix} has no pins, skipping")
+                continue
+            for i in range(gio.pins):
+                # for single pins we kick out the index
+                j = "" if gio.pins == 1 else f"{i}"
+                if gio.IO == IO.INPUT:
+                    if gio.configAccess:
+                        logger.error(
```
KelvinChung2000: `.opt`, there are a few in here as well.

FABulous/fabric_definition/Gen_IO.py
EverythingElseWasAlreadyTaken: I thought about this, but currently all Gen_IO of a tile get assembled together to one Bel.

I could split this, that one Gen_IO creates on Bel, but it felt more robust and easier to handle if you have one big IO Bel instead of a plenty small one.

FABulous/fabric_definition/Tile.py
EverythingElseWasAlreadyTaken: Good idea, I'll implement this. :)

FABulous/fabric_generator/fabric_gen.py
EverythingElseWasAlreadyTaken: I can move it to the fabric automation file, this might be a better fit.

FABulous/fabric_definition/Gen_IO.py
KelvinChung2000: This will still be a single Bel. It will just have a special attribute.  

So rather than:
```
module IO_bidir_1(...);
endmodule
```
and you have lot of them,
become
```
module IO_bidir_3(...);
IO_bidir_1 inst_1(...);
IO_bidir_1 inst_2(...);
IO_bidir_1 inst_3(...);
endmodule
```

but still a single bel.

EverythingElseWasAlreadyTaken: Yes, kind of, but im not talking about the RTL generation, I mean the stored data representation of the GenIO stuff which could get cumbersome if I inherit from BEL, at least in the current implementation. 

Because one Gen_IO object, represents one GEN_IO line in the CSV file, which are altogether stored in the Tile Object and get converted to one Bel later.  So a collection of GenIO currently represents one Bel. So one Gen_IO object is not equal to a Bel object in terms of, the required attributes.

If I make GenIO inherited from Bel, I need to provide the required attributes a Bel has, name, ports etc., but this will split one Bel (RTL) that is later generated, into several Bel objects for one Tile object. So tile would contain several Bels objects that altogether would represent one created BEL, which seems more confusing and could bring more sideffects and corner cases.  

I see two alternative ways here:
1.  I could move the list of GenIOs to the Bel class instead of a Tile, but this would mostly just move the same stuff to a different place.
2. I could extend the GenIO class in a way that it can all information of all GenIOs for a Tile, which can be then inherited from a Bel. But this would make the structure more complex and needs more changes in the code. 

I'm not sure what's brings more benefits here, or if we should just leave it as it is.

KelvinChung2000: If this is how it was designed in the first place, then leave it as it is. Since it will eventually become a Bel, this does not matter then.

FABulous/fabric_definition/Fabric.py
EverythingElseWasAlreadyTaken: None is the default for get() on dict

https://www.w3schools.com/python/ref_dictionary_get.asp

FABulous/fabric_definition/Tile.py
EverythingElseWasAlreadyTaken: https://github.com/FPGA-Research/FABulous/pull/380/commits/50882fa23e9799384fb40b2a2f977b6fb726f364

EverythingElseWasAlreadyTaken: https://github.com/FPGA-Research/FABulous/pull/380/commits/571f0e1a06ac672b208425edd33f9be54bd04c0f

Since we add the IO Bel to the tile after generating it, I don't need to count the genIO config bits manually anymore. :)

KelvinChung2000:
Maybe it is using the wrong `requirement.txt`?

Maybe can try modifying the workflow file and give this a try

```
pre-build-command: "apt-get update -y && apt-get install tk git -y && pip install -r docs/requirements.txt && pip install -e ."
```

PR #385, opened 2025-07-03
fix: Cleanup unused code

EverythingElseWasAlreadyTaken:
While adding the ruff lint, I stumbled upon a lot of unused legacy code.
This PR cleans this.


EverythingElseWasAlreadyTaken:
@KelvinChung2000  
Please check, if you could need any of the removed functions in the future. 

Also, the utility.py contains now just one function, maybe you could move them somewhere else when you do the file splitting.

KelvinChung2000:
I will move it into something like "parser_list.py" later.

PR #386, opened 2025-07-03
chore: Refactor and add test for config mem generation

KelvinChung2000:
Moving the config memory generation into a separate folder.

Moving the config memory CSV parser into a separate file.

Rewrite the genConfigmemInit function to better handle edge cases.

Update Config mem generation to include additional edge case handling, such as preventing negative number indices and ensuring config mem CSV matches the tile configuration memory count. 

Added test fixtures which can be reused for other components.

### DIFF
--- a/.github/workflows/pytest_testing.yml
+++ b/.github/workflows/pytest_testing.yml
@@ -17,5 +17,5 @@ jobs:
           additional_system_packages: "iverilog"
       - name: Run pytest
         run: |
-          export PATH="/usr/bin:$PATH"
+          echo "Running pytest with PATH=$PATH"
           pytest -v -x
--- a/.github/actions/prepare_FABulous_container/action.yaml
+++ b/.github/actions/prepare_FABulous_container/action.yaml
@@ -60,3 +60,8 @@ runs:
       shell: bash
       run: |
         pip3 install -e .
+    - name: Set PATH
+      shell: bash
+      run: |
+        echo "/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin:$PATH" >> "$GITHUB_PATH"
+


.github/workflows/pytest_testing.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with: 
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: ./.github/actions/prepare_FABulous_container
+        with:
+          additional_system_packages: "iverilog"
+      - name: Run pytest
+        run: |
+          export PATH="/usr/bin:$PATH"
```
EverythingElseWasAlreadyTaken: If this is needed because of the installation of iverilog, we should maybe move it to prepare_FABulous_container action.

EverythingElseWasAlreadyTaken (approved):
Looks good to me. Nothing to comment, except one small question.

.github/workflows/pytest_testing.yml
KelvinChung2000: This is for overriding the OSS-CAD setup. The export command does not persist across run steps, this is why I do it here.

EverythingElseWasAlreadyTaken: Yes, I know. I thought about which default behavior we expect.
Should the OSS-CAD-Suite packages have priority over system packages, like currently, or the other way around? 
Since it could be, that the issue you ran in, could also come up somewhere else.

KelvinChung2000: I think we should prioritise the system. 
 
 Regarding taking priority, I notice that the ghdl test run via pytest is noticeably longer than the one run by simply calling 'make'. This might be something related as well.

EverythingElseWasAlreadyTaken: I also think so, could you move the `export` to the prepare_FABulous_container workflow? 
Maybe we should extend it to the Ubuntu $PATH defaults, something like:
 `/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin:$PATH` 
 
But I'm not entirely sure how the PATH handling exactly works in GitHub action, seems like the usual $PATH is not shared between workflows, since there also is a GITHUB_PATH variable, which seems to be the actual global PATH.:
https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#adding-a-system-path


Could be that pytest uses the GHDL LLVM version from the OSS-CAD-suite. 
The GHDL mcode backend we install manually is much faster (~ x3)  for our simulation than the LLVM backend that comes with the OSS-CAD-Suite. Could be, that pytest somehow uses the LLVM backend. 
But maybe the PATH fix, also helps here.

PR #394, opened 2025-07-18
fix: requirements.txt format

EverythingElseWasAlreadyTaken:
Remove some trailing commas in requirements.txt, which could lead to errors in older pip versions and remove duplicate requirements.

Set lower base version of bitarray in pyproject.toml for better compatibility.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
   'fasm>=0.0.2',
   'requests>=2.0.0',
   'cmd2>=2',
-  'bitarray>=3.4.3',
+  'bitarray>=3.0.0',
 ]
 
 [project.urls]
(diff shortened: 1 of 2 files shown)


KelvinChung2000:
The bitarray needs to be 3.3.0 or above. I use a feature which requires that version, https://github.com/ilanschnell/bitarray/issues/230. Technically, when installed via requirements.txt, this will retrieve the latest version; however, I think we should keep this in the requirements.

EverythingElseWasAlreadyTaken:
> The bitarray needs to be 3.3.0 or above. I use a feature which requires that version, [ilanschnell/bitarray#230](https://github.com/ilanschnell/bitarray/issues/230). Technically, when installed via requirements.txt, this will retrieve the latest version; however, I think we should keep this in the requirements.

Sorry, I didn't know that. I'll send a new PR that fixes the pyproject.toml.

Issue #337, opened 2025-02-28
Cleanup and unify Verilog and VHDL bel files


EverythingElseWasAlreadyTaken:
Most of the Verilog and VHDL bel files need some cleanup.

We should: 
- Apply a consistent coding style.
- Remove commented code.
- Unify implementations and remove unneeded module instantiations.

KelvinChung2000:
Coding style probably can be done with (Verible)[https://github.com/chipsalliance/verible]. But we will need to customise it to make it fit for Verilog. 

Some extra resources:
https://github.com/chipsalliance/verible/issues/317

KelvinChung2000:
More or less done, just need some fix up

PR #399, opened 2025-07-22
Verilog and VHDL linting and clean up

KelvinChung2000:
As a tracker

#337

- [ ] Apply a consistent coding style.
  - [ ] Verilog 
  - [x]  VHDL
- [ ] Remove commented code.
  - [ ] Verilog 
  - [ ]  VHDL
- [ ] Unify implementations and remove unneeded module instantiations.
- [ ] Check Bel Map
- [ ] Config Access Bel Map is not working since the config bits have the same name as the ports --> Maybe also others?
- [ ] Fix Latch in [BlockRAM_1KB](https://github.com/FPGA-Research/FABulous/blob/FABulous2.0-development/FABulous/fabric_files/FABulous_project_template_verilog/Fabric/BlockRAM_1KB.v#L45) (Verilog and VHDL)
  - [x] Verilog 
  - [ ]  VHDL 
- [ ] Run Verilator Lint on the whole fabric and fix errors.
- [ ] Check DSP Bel [MULADD](https://github.com/FPGA-Research/FABulous/blob/FABulous2.0-devnelopment/FABulous/fabric_files/FABulous_project_template_verilog/Tile/DSP/DSP_bot/MULADD.v)
  - [x] Verilog
  - [ ] VHDL 
- [x]  Testing Frame data reg
- [ ] Testing Frame select / rename
- [ ] Testing bitbang
- [ ] Testing configFSM
- [ ] testing eFPGA_config
- [ ] testing / rename LUT4AB
- [ ] testing / rename RegFile_32x4
- [ ] Replace the PDK specific SRAM instantiation by a more generic interface
- [ ] Add `default_nettype`


IAmMarcelJung:
We should also consider updating the filenames in e.g. the Fabric folder, since they all use different conventions.

FABulous/fabric_files/FABulous_project_template_verilog/Test/sequential_16bit_en_tb.v
```diff
@@ -1,110 +1,110 @@
...
-        end
+  wire [27:0] I_top;
+  wire [27:0] T_top;
+  reg  [27:0] O_top = 0;
+  wire [55:0] A_cfg, B_cfg;
+
+  reg CLK = 1'b0;
+  reg resetn = 1'b1;
+  reg SelfWriteStrobe = 1'b0;
+  reg [31:0] SelfWriteData = 1'b0;
```
IAmMarcelJung:
```suggestion
  reg [31:0] SelfWriteData = 32'b0;
```

FABulous/fabric_files/FABulous_project_template_verilog/Fabric/eFPGA_Config.v
```diff
@@ -1,102 +1,116 @@
...
+  input s_clk;
+  input s_data;
+  // CPU configuration port
+  input [32-1:0] SelfWriteData;  // configuration data write port
+  input SelfWriteStrobe;  // must decode address and write enable
 
-    output [32-1:0] ConfigWriteData;
-    output ConfigWriteStrobe;
+  output [32-1:0] ConfigWriteData;
+  output ConfigWriteStrobe;
```
IAmMarcelJung, line 35: Below we just write `[31:0]`, we should stick to one notation. I would vote for `[31:0]`. We could also introduce a parameter `CONFIG_DATA_WIDTH` or so and then use `[CONFIG_DATA_WIDTH-1:0]`.

FABulous/fabric_files/FABulous_project_template_verilog/Fabric/eFPGA_Config.v
```diff
@@ -1,102 +1,116 @@
...
-    assign ReceiveLED = UART_LED^BitBangWriteStrobe;
+  assign ComActive = UART_ComActive;
+  assign ReceiveLED = UART_LED ^ BitBangWriteStrobe;
 
-//  wire [FrameBitsPerRow-1:0] FrameAddressRegister;
-//  wire LongFrameStrobe;
-//  wire [RowSelectWidth-1:0] RowSelect;
+  //  wire [FrameBitsPerRow-1:0] FrameAddressRegister;
+  //  wire LongFrameStrobe;
+  //  wire [RowSelectWidth-1:0] RowSelect;
```
IAmMarcelJung, line 97:
```suggestion

```

KelvinChung2000: Stange, I thought the formatter would fix issue like this

FABulous/fabric_files/FABulous_project_template_verilog/Fabric/eFPGA_Config.v
KelvinChung2000: We should probably move to using `CONFIG_DATA_WIDTH`.

FABulous/fabric_files/FABulous_project_template_verilog/Fabric/eFPGA_Config.v
IAmMarcelJung: Maybe it's too hard to distinguish between actual intended comments and just commented out dead code for the formatter?

KelvinChung2000:
The default is snake_case, but the regex is `[a-z_0-9]+`, which makes PascalCase fit within it. I will need to see how to ban PascalCase while keeping snake_case.

IAmMarcelJung:
> Please don't put `timescale` directives in design files. They should only be in testbenches or where delay statements are needed.

Okay I will remove them again. For some reason verilator always complained about them missing and cluttered the linting log with warnings. Maybe this was due to some misconfiguration of verilator on our side.

> While you're at it, it would be great if the signal names could be renamed (if the formatter can do that). I've actually never seen PascalCase used in any other (System)Verilog project. 

I 100% agree, the variable naming is currently all over the place.

> There is a style guide from lowRISC which I quite like: https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md

I also like this style guide, except that I personally prefer an indentation of 4 whitespaces to more emphasize the indentation and also force people to not nest too much logic (in combination with a line length limit), but that's debatable.

KelvinChung2000:
If you don't mind, can you set the configuration of the formatter, and we can slowly clean those up. You can change the `rule.config` file to set the desired behaviour.

mole99:
I'll give this a try after HeiChips, if that's alright.

PR #402, opened 2025-07-24
chore: Split and move file

KelvinChung2000:
Splitting fabric_gen.py into multiple smaller files. 

Moving files into folders.


FABulous/fabric_cad/gen_bitstream_spec.py
```diff
@@ -0,0 +1,166 @@
...
+
+from FABulous.fabric_definition.Fabric import Fabric
+from FABulous.fabric_generator.parser.parse_configmem import parseConfigMem
+from FABulous.fabric_generator.parser.parse_switchmatrix import parseMatrix
+
+if TYPE_CHECKING:
+    from FABulous.fabric_definition.ConfigMem import ConfigMem
+
+
+def generateBitsStreamSpec(fabric: Fabric) -> dict[str, dict]:
```
EverythingElseWasAlreadyTaken:
```suggestion
def generateBitstreamSpec(fabric: Fabric) -> dict[str, dict]:
```

Can rename this to generateBitstreamSpec?

EverythingElseWasAlreadyTaken (approved):
Nice work! This makes the codebase much more clear and maintainable!


It looks fine in general, I have just some minor comments/ideas, we can discuss:

- We should rename the generateBit**s**StreamSpec function, to something like generateBitstreamSpec. 
 
-  We could maybe also rename the model_generation_npnr.py to gen_npnr_model.py to match the naming scheme. 

For the code generator, maybe we should rename the code_generation_VHDL.py 
to code_generator_VHDL.py and also for Verilog to match the base class. 
We should then also rename the Writer classes, to codeGenerator, to match the baseclass.

PR #403, opened 2025-07-24
chore: Adding more ruff rules 

KelvinChung2000:
Adding the following:
* A:  flake8-builtins
* PTH: flake8-use-pathlib
* TRY: tryceratops
* SIM: flake8-simplify
* PT: flake8-pytest-style
* ARG: flake8-unused-aarguments

### DIFF
--- a/.github/actions/prepare_FABulous_container/action.yaml
+++ b/.github/actions/prepare_FABulous_container/action.yaml
@@ -6,6 +6,10 @@ inputs:
     description: "Python version to use"
     required: false
     default: "3.12"
+  oss-cad-suite-nightly:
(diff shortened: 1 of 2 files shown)


EverythingElseWasAlreadyTaken (approved):

EverythingElseWasAlreadyTaken:
Not sure what the problem with the VHDL tests are.
Seems like GHDL is broken in oss-cad-suite?

EverythingElseWasAlreadyTaken:
> It looks like the issue is more specifically with the ghdl plugin. I just checked that Yosys has an update to their API, which might have broke the ghdl plugin. Should we switch from using the latest version to pinning to a known working version?

This could solve the issue. :) 
I can give it a shot.

PR #406, opened 2025-07-25
chore: Full annotation

KelvinChung2000:
Add type annotation for all python files.

Enable ANN rule

### DIFF
--- a/tests/fabric_gen_test/configmem_test/test_configmem_rtl.py
+++ b/tests/fabric_gen_test/configmem_test/test_configmem_rtl.py
@@ -1,20 +1,34 @@
 """RTL behavior validation for generated ConfigMem modules using cocotb."""
 
 import json
+from collections.abc import Callable
 from pathlib import Path
+from typing import Any, Protocol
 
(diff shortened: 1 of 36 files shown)


tests/fabric_gen_test/configmem_test/test_configmem_rtl.py
```diff
@@ -23,30 +37,7 @@ def load_bit_mapping():
     return {}
 
 
-def create_test_pattern(config_info, frame_index, pattern_type="ones"):
```
EverythingElseWasAlreadyTaken, line 26: Was it intended to remove this function?

EverythingElseWasAlreadyTaken (commented):
Nice work, looks good so far. 

You have removed one function and there was no comment/ commit message about it. Please make sure that this was intended.

tests/fabric_gen_test/configmem_test/test_configmem_rtl.py
KelvinChung2000: Yes, this is not used by any test

PR #367, opened 2025-05-28
YosysJson Object

KelvinChung2000:
Add typing to yosys json file to aid development

PR #410, opened 2025-08-04
chore: centralise all settings

KelvinChung2000:
Centralise all the environment variable settings using [Pydantic-settings](https://docs.pydantic.dev/latest/concepts/pydantic_settings/).

Add sub dependency group

### DIFF
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -6,7 +6,6 @@ from pathlib import Path
 from loguru import logger
 from packaging.version import Version
 
-from FABulous.FABulous_CLI.FABulous_CLI import FABulous_CLI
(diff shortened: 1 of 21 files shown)


EverythingElseWasAlreadyTaken (approved):
Thanks for the PR! It looks good now. :) 
I've checked, yosys is not a hard dependency anymore. :)

We could also move the setup_global_env_vars and setup_project_env_vars directly to the FABulous_settings.py, so we have everything settings related in one place. 

Also, in the `FABulous/fabric_generator/parser/parse_hdl.py`  `parseBelFile` function, `yosys` is still hardcoded in the call and not taken from FABulous_settings, but i guess this will be fixed with #367 anways?

PR #428, opened 2025-08-14
chore(CI): Update checkout and repurpose makefile CI

KelvinChung2000:
Updated all action to using the lastest checkout action

Update the makefile CI for checking old project fabric, as now we can run regular check invoking from pytest.

### DIFF
--- a/.github/workflows/fabric_gen.yml
+++ b/.github/workflows/fabric_gen.yml
@@ -19,10 +19,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
+        with:
+          ref: github.base_ref
+      - uses: ./.github/actions/prepare_FABulous_container
+      - name: generate old project
+        run: |
+          FABulous -c demo
+      
+      - uses: actions/checkout@v5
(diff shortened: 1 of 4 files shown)


.github/workflows/fabric_gen.yml
```diff
@@ -1,43 +1,31 @@
...
+            type: vhdl
 
-  run_verilog_simulation_makefile:
-    name: Run Verilog fabric generator flow and simulation with FABulous makefile
-    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
+        with:
+          ref: ${{ github.base_ref }}
```
EverythingElseWasAlreadyTaken: Maybe we need to remove the `run on: push` or hardcode the dev branch, since base_ref is only available for pull request triggers:

```
The base_ref or target branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.
```

https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context

EverythingElseWasAlreadyTaken (approved):
Nice idea!

PR #430, opened 2025-08-14
chore: Expand user design top wrapper gen

EverythingElseWasAlreadyTaken:
Expand the user design top wrapper generation to include the whole fabric.


FABulous/fabric_cad/gen_design_top_wrapper.py
```diff
@@ -62,59 +62,60 @@ def generateUserDesignTopWrapper(
...
+                reversed(bels)
+            ):  # we walk backwards trough the bel list
+                belstr = ""
+                # we only add bels with external ports to the top wrapper.
+                if not bel.externalInput and not bel.externalOutput:
+                    logger.info(
+                        f"Skipping bel {bel.name} in tile X{x}Y{y} since it has no external ports"
+                    )
+                    continue
+                if len(bel.inputs and bel.outputs) == 0:
```
KelvinChung2000:
```suggestion
                if len(bel.inputs + bel.outputs) == 0:
```

KelvinChung2000 (changes requested):
Instead of using `parseBelFile`, you may want to consider updating to using the YosysJson for parsing, and getting the user design port. 

Another thing that you can consider is transitioning to use the HDLGen, which allows you to generate VHDL, too. But ifit  is too much work, I will try to do the rewrite when I have time.

PR #431, opened 2025-08-16
doc: add a lot of doc

KelvinChung2000:
Adding doc to all functions and all files with AI assistance. I have glanced through all the changes and corrected some of the mistakes, but I might still have missed some. 

Format all the doc to a standard of `pydocstyle`.

The repo is now  99.5% documented.

PR #432, opened 2025-08-18
chore: Add header to whole project

KelvinChung2000:
Set a pre-commit for adding a header to every file if it does not already have one. If it got one, it will update it. 

I will set this as a draft for now. Once #431 is done, I will apply the pre-commit to all files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,7 +46,7 @@ repos:
     - id: ruff-format
 
 - repo: https://github.com/KelvinChung2000/license-header-pre-commit
-  rev: 0.4.0
+  rev: v0.4.1
   hooks:
     - id: license-header
       name: License Header


.pre-commit-config.yaml
```diff
@@ -44,3 +44,15 @@ repos:
...
+- repo: https://github.com/KelvinChung2000/license-header-pre-commit
+  rev: v0.4.1
+  hooks:
+    - id: license-header
+      name: License Header
+      language: python
+      files: \.(py|v|sv|vhd|vhdl)$
+      args: [
+        "--template", "license-header",
+        "--copyright-holder", "Novel Computing Technologies Group @ University Heidelberg"
```
IAmMarcelJung, line 57: Did you discuss this with Dirk? This wouldn't include e.g. you, Leo and anyone else outside of our University's Group. I usually use `The FABulous Contributors` (saw something like this in lowRisc files or so) for the projects I'm working on related to FABulous, but I'm also not sure what the best way would be.

KelvinChung2000:
> This is a great addition, thanks for this!
> Please see my comment on the name of the copyright holder.
> Also, we could consider adding an empty line after the header, to separate it a bit from the code itself? Don't know what the convention for this is.
> 
> This might more be for your acutal hook, but I get an error if I modify the header (or add an empty line after it) and run the hook again, although the hook modifies the header correctly (so I would not expect an error to be thrown:
> ```console
> License Header...........................................................Failed
> - hook id: license-header
> - exit code: 1
> - files were modified by this hook
> 
> Updated license header in FABulous/fabric_files/FABulous_project_template_verilog/Tile/W_IO/IO_1_bidirectional_frame_config_pass.v
> 
> Modified 1 files with license headers
> ```

The newline will upset ruff, so it cannot be there. Maybe I can try a commented newline and see does it work. 

I have not talk to Dirk about it yet, but changing that is easy. Maybe we can talk about this in next week's meeting

.pre-commit-config.yaml
mole99: In LibreLane we also use "LibreLane Contributors" from now on with a contributors file in the root.

IAmMarcelJung: Sounds nice, probably we should adopt this!

PR #440, opened 2025-08-22
feat: better context management and new start interface

KelvinChung2000:
Updated the context management to use `get_context` only, and this will only create the model once and automatically resolve all environment dependencies. This can be used as the base for all feature global settings. 

Removed dependency on `FAB_ROOT`. Now the project does not reference the installation dir directly. Project creation is now done via the `importlib.resources`. Global `.env` is now stored at user `$HOME/.fabulous`. And this can be used for all future installations and FABulous related cross-session assets. 

Transition from `argparse` based command to `typer`, which offers a much better looking interface and help message. This also makes creating new commands down the line much easier, such as `FABulous GUI` etc.  Full backwards compatibility in the interface is maintained, but we should drop it soon. 

image

There are other bonuses from `typer` such as auto-complete install, but I have not tested them yet.

### DIFF
--- a/FABulous/FABulous_CLI/helper.py
+++ b/FABulous/FABulous_CLI/helper.py
@@ -5,8 +5,11 @@ import re
 import shutil
 import sys
 import tarfile
+import tempfile
 from collections.abc import Callable, Sequence
+from importlib import resources
 from importlib.metadata import version
+from importlib.resources.abc import Traversable
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal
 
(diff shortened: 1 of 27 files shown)


FABulous/FABulous_CLI/helper.py
```diff
@@ -90,35 +96,63 @@ def create_project(
         The language of project to create ("verilog" or "vhdl"), by default "verilog".
     """
     logger.info(project_dir)
+
+    if lang not in ["verilog", "vhdl"]:
```
EverythingElseWasAlreadyTaken, line 100: lang is already set to verilog as default for the function. 
If something else, not vhdl or verilog is specified, we should raise an error here

PR #444, opened 2025-08-29
test: Add reference tests

EverythingElseWasAlreadyTaken:
A pytest-based framework for testing FABulous against reference projects with regression testing capabilities.

See [tests/reference_tests/README.md](https://github.com/EverythingElseWasAlreadyTaken/FABulous/blob/1aa3e6a6d244599a3eae93812e02164f9aa20dfd/tests/reference_tests/README.md) for more information.

These are the executed tests:
https://github.com/FPGA-Research/FABulous-demo-projects/blob/main/reference_projects_config.yaml

### DIFF
--- a/tests/reference_tests/conftest.py
+++ b/tests/reference_tests/conftest.py
@@ -1,0 +1,118 @@
+"""Pytest configuration and fixtures for reference testing."""
+
+import os
+from pathlib import Path
+
+import pytest
+
+from tests.reference_tests.helpers import download_reference_projects
+from tests.reference_tests.reference_projects_test import load_reference_projects_config
+
+
+# Session-level configuration storage
+class SessionConfig:
+    """Centralized configuration that can be modified during session start."""
+    def __init__(self) -> None:
+        self.projects_dir: Path
+        self.projects_conf: Path
+        self.repo_url: str
+        self.download_projects: bool
+        self.verbose: bool
+
+# globasl session config instance
+_session_config = SessionConfig()
+
+
+@pytest.fixture(autouse=True)
+def test_environment():
+    """Setup test environment for FABulous - reusing existing CLI test env setup."""
+    fabulous_root = str(Path(__file__).resolve().parent.parent.parent / "FABulous")
+    original_env = {}
+
+    # Save original environment
+    for key in ["FAB_ROOT", "FAB_PROJ_DIR", "FABULOUS_TESTING"]:
+        original_env[key] = os.environ.get(key)
+
+    # Set test environment
+    os.environ["FAB_ROOT"] = fabulous_root
+    os.environ["FABULOUS_TESTING"] = "TRUE"
+
+    yield
+
+    # Restore original environment
+    for key, value in original_env.items():
+        if value is None:
+            os.environ.pop(key, None)
+        else:
+            os.environ[key] = value
+
+
+@pytest.fixture(scope="session")
+def config_path() -> Path:
+    """Get the reference projects config path from session config."""
+    if _session_config.projects_conf is None:
+        raise RuntimeError("Session config not initialized. This should be set in pytest_sessionstart.")
+    return _session_config.projects_conf
+
+@pytest.fixture(scope="session")
+def projects_dir() -> Path:
+    """Get the projects directory from session config."""
+    if _session_config.projects_dir is None:
+        raise RuntimeError("Session config not initialized. This should be set in pytest_sessionstart.")
+    return _session_config.projects_dir
+
+
+@pytest.fixture
+def reference_projects_config(config_path: Path) -> list:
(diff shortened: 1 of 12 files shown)


tests/reference_tests/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+        self.repo_url: str
+        self.download_projects: bool
+        self.verbose: bool
+
+# globasl session config instance
+_session_config = SessionConfig()
+
+
+@pytest.fixture(autouse=True)
+def test_environment():
```
KelvinChung2000: I have done something similar in the test_CLI, I recall that one of my PR has moved this out to the toplevel conftest. Do you think we should move this out now, then, and deal with it later when I rebase?

tests/reference_tests/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+    """Setup test environment for FABulous - reusing existing CLI test env setup."""
+    fabulous_root = str(Path(__file__).resolve().parent.parent.parent / "FABulous")
+    original_env = {}
+
+    # Save original environment
+    for key in ["FAB_ROOT", "FAB_PROJ_DIR", "FABULOUS_TESTING"]:
+        original_env[key] = os.environ.get(key)
+
+    # Set test environment
+    os.environ["FAB_ROOT"] = fabulous_root
```
KelvinChung2000: Rather than directly modify the `env`, you can use `monkeypatch.setenv`, which avoids state leaking.

tests/reference_tests/reference_projects_test.py
```diff
@@ -0,0 +1,258 @@
...
+    language: str,
+    caplog: pytest.LogCaptureFixture,
+    commands: list[str] | None = None,
+    skip_on_fail: bool = False,
+) -> tuple[FABulous_CLI, dict[str, Any]]:
+    """Run standard FABulous commands using existing test patterns."""
+
+    setup_logger(0, False)
+
+    os.environ["FAB_PROJ_DIR"] = str(project_path)
```
KelvinChung2000: change to use `monkeypatch`

tests/reference_tests/reference_projects_test.py
```diff
@@ -0,0 +1,258 @@
...
+
+        metafunc.parametrize(
+            "project", active_projects, ids=[p.name for p in active_projects]
+        )
+    else:
+        raise RuntimeError("No 'project' fixture found in test function.")
+
+
+def test_reference_project_execution(
+    project: ReferenceProject, tmp_path: Path, caplog: pytest.LogCaptureFixture) -> None:
```
KelvinChung2000: The `caplog` here will need a special caplog if you are gathering the log from `loguru`, I have a custom `caplog` fixture implemented in `CLI_test`

tests/reference_tests/reference_projects_test.py
```diff
@@ -0,0 +1,258 @@
...
+def run_fabulous_commands_with_logging(
+    project_path: Path,
+    language: str,
+    caplog: pytest.LogCaptureFixture,
+    commands: list[str] | None = None,
+    skip_on_fail: bool = False,
+) -> tuple[FABulous_CLI, dict[str, Any]]:
+    """Run standard FABulous commands using existing test patterns."""
+
+    setup_logger(0, False)
```
KelvinChung2000: This can be in the env setup

tests/reference_tests/reference_projects_test.py
```diff
@@ -0,0 +1,258 @@
...
+            projects.append(project)
+        except KeyError as e:
+            logger.warning(f"Invalid project config, missing key {e}: {project_data}")
+        except Exception as e:  # noqa: BLE001
+            logger.warning(f"Failed to load project config: {e}")
+
+    return projects
+
+
+def run_fabulous_commands_with_logging(
```
KelvinChung2000: I think this in helper make more sense.

Maybe we can merge this with the `cli` fixture such that the `cli` fixture is a generator fixture which return a `cli` factory, and we can pass argument into the to the factory to get a configured CLI.

KelvinChung2000: After a bit more thinking, this also means the `CLI_test` will need to be updated as well. Could be done in  the future

KelvinChung2000 (changes requested):

tests/reference_tests/reference_projects_test.py
EverythingElseWasAlreadyTaken: I've moved it to helpers, we can change this in a dedicated PR

EverythingElseWasAlreadyTaken:
I've created a global conftest.py, where I moved all shared fixtures and functions.

KelvinChung2000 (approved):

PR #447, opened 2025-09-02
chore: testing exsting construct

KelvinChung2000:
Spliting #399 written construct into this PR

### DIFF
--- a/fabulous/fabric_files/FABulous_project_template_vhdl/Fabric/models_pack.vhdl
+++ b/fabulous/fabric_files/FABulous_project_template_vhdl/Fabric/models_pack.vhdl
@@ -354,6 +354,125 @@ end architecture Behavior;
 
 library ieee;
   use ieee.std_logic_1164.all;
+  use ieee.numeric_std.all;
+
+-- Sky130 SRAM 1RW1R 32x256x8 simulation model
+--   ADDR_WIDTH = 8
+--   DATA_WIDTH = 32
+--   DELAY = 3
+--   NUM_WMASKS = 4
+--   RAM_DEPTH = 256
+
+entity sram_1rw1r_32_256_8_sky130 is
+  port (
+    addr0  : in    unsigned(7 downto 0);
+    addr1  : in    unsigned(7 downto 0);
+    clk0   : in    std_logic;
+    clk1   : in    std_logic;
+    csb0   : in    std_logic;
+    csb1   : in    std_logic;
+    din0   : in    unsigned(31 downto 0);
+    dout0  : out   unsigned(31 downto 0);
+    dout1  : out   unsigned(31 downto 0);
+    web0   : in    std_logic;
+    wmask0 : in    unsigned(3 downto 0)
+  );
+end entity sram_1rw1r_32_256_8_sky130;
+
+architecture from_verilog of sram_1rw1r_32_256_8_sky130 is
+
+  type mem_type is array (255 downto 0) of unsigned(31 downto 0);
+
+  signal dout0_reg  : unsigned(31 downto 0);
+  signal dout1_reg  : unsigned(31 downto 0);
+  signal addr0_reg  : unsigned(7 downto 0);
+  signal addr1_reg  : unsigned(7 downto 0);
+  signal csb0_reg   : std_logic;
+  signal csb1_reg   : std_logic;
+  signal din0_reg   : unsigned(31 downto 0);
+  signal mem        : mem_type;
+  signal web0_reg   : std_logic;
+  signal wmask0_reg : unsigned(3 downto 0);
+
+begin
+
+  dout0 <= dout0_reg;
+  dout1 <= dout1_reg;
+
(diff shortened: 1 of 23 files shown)


fabulous/fabric_files/FABulous_project_template_vhdl/Fabric/models_pack.vhdl
```diff
@@ -752,209 +781,339 @@ use ieee.std_logic_1164.all;
...
+  end component;
+
+end package my_package;
+
+-- Sky130 SRAM 1RW1R 32x256x8 simulation model
+library ieee;
+use ieee.std_logic_1164.all;
+use ieee.numeric_std.all;
+
+-- Generated from Verilog module sram_1rw1r_32_256_8_sky130 (/home/kelvin/FABulous_fork.worktrees/FABulous2.0-development/FABulous/fabric_files/FABulous_project_template_verilog/Fabric/models_pack.v:508)
```
IAmMarcelJung: Probably remove the absoulte path here

tests/fabric_construct_test/test_BlockRAM_1KB_rtl.py
```diff
@@ -0,0 +1,386 @@
...
+        (0x10, 0x33333333),
+        (0x20, 0x44444444),
+        (0xFF, 0x55555555),
+    ]
+
+    # Write all test patterns
+    for addr, data in test_patterns:
+        dut.wr_addr.value = addr
+        dut.wr_data.value = data
+        await RisingEdge(dut.clk)
```
IAmMarcelJung, line 160: Do we even need to test different values if we only do the smoke test and check if an int is returned?

tests/fabric_construct_test/test_BlockRAM_1KB_rtl.py
```diff
@@ -0,0 +1,386 @@
...
+    dut.wr_data.value = new_data  # Try to write different data
+    # Don't set the dynamic write enable bit (bit 20 in wr_data)
+    await RisingEdge(dut.clk)
+    await Timer(Decimal(10), units="ps")
+
+    # Read back - smoke test that read completes
+    dut.rd_addr.value = test_addr
+    await RisingEdge(dut.clk)
+    await Timer(Decimal(10), units="ps")
+    # Note: Cannot verify write was prevented due to SRAM timing variability
```
IAmMarcelJung, line 279: Does this even test anything if there is no assertion? And even if there was an assertion, a smoke test probably also would not test the behavior, since the first write would already write a valid value and therefore the read will always work?

tests/fabric_construct_test/test_BlockRAM_1KB_rtl.py
```diff
@@ -0,0 +1,386 @@
...
+    dut.C1.value = 0
+    dut.C2.value = 0
+    dut.C3.value = 0
+    dut.C4.value = 1  # alwaysWriteEnable = 1
+    dut.C5.value = 0  # no register bypass
+
+    await RisingEdge(dut.clk)
+
+    # Test address extension via data bits
+    # Default parameters: ReadAddressMSBFromDataLSB = 24, WriteAddressMSBFromDataLSB = 16
```
IAmMarcelJung, line 364:
```suggestion
    # Default parameters: READ_ADDRESS_MSB_FROM_DATALSB = 24, WRITE_ADDRESS_MSB_FROM_DATALSB = 16
```
It helps to match the actual parameter name to know where this comes from.

tests/fabric_construct_test/test_BlockRAM_1KB_rtl.py
```diff
@@ -0,0 +1,386 @@
...
+    # Bits [25:24] of wr_data become bits [9:8] of read address for next read
+    extended_read_data = base_data | (0x3 << 24)  # Set bits [25:24] = 11
+    dut.wr_data.value = extended_read_data  # This affects read address mapping
+    dut.rd_addr.value = base_addr
+    await RisingEdge(dut.clk)
+    await Timer(Decimal(10), units="ps")
+
+    # The exact behavior depends on how the address extension is implemented
+    # This smoke test verifies the module accepts the extended addressing format
+    # without simulation errors
```
IAmMarcelJung, line 386: Does this even test anything if there is no assertion? Will this just fail if the module does not accept the extended addressing format?

IAmMarcelJung (commented):
A first small review, basically only of the BRAM test. In general LGTM, just some questions.

tests/fabric_construct_test/test_BlockRAM_1KB_rtl.py
KelvinChung2000: This is probably something I missed during my review. I will need to see what should be done

KelvinChung2000:
I don't know what the right thing to do is, since they need to be there for complete coverage. Unless we decide we will never "retire" them, then we can add keep them, and add in the flexibility as we need to.

PR #458, opened 2025-09-04
chore[fabric_files]: Remove LUT4AB_ConfigMem.csv since it will be generated during flow

EverythingElseWasAlreadyTaken:


KelvinChung2000:
This is a shuffled bit mapping that is physically optimised to minimise area for SKY130. The remapping optimisation feeds back the bitstream remapping. It works like a loop. It generates a remapping, which generates a new fabric RTL, then go through to the backend to generate the `def`, then some script which performs optimisation based on the `def` information, and generates a new configmem csv, which feeds back into FABulous for new fabric RTL, and this repeats until the tiles cannot get smaller. And that remapping is the result of the process described.

EverythingElseWasAlreadyTaken:
> This is a shuffled bit mapping that is physically optimised to minimise area for SKY130. The remapping optimisation feeds back the bitstream remapping. It works like a loop. It generates a remapping, which generates a new fabric RTL, then go through to the backend to generate the `def`, then some script which performs optimisation based on the `def` information, and generates a new configmem csv, which feeds back into FABulous for new fabric RTL, and this repeats until the tiles cannot get smaller. And that remapping is the result of the process described.

If this is an optimized version for SKY130, I am not sure if we should keep it in the general demo project.

Whenever the automated physical flow is done, we should test if this brings any benefits for other PDKs.

PR #459, opened 2025-09-04
fix: Fix typer run command

KelvinChung2000:
I need to do the fix for the multi-line fstring together, as it seems like this is not supported at python3.10, which is what my system python is using.


EverythingElseWasAlreadyTaken:
> It is looking like a pre-commit issue, as it refers back to my system version even I am running under venv

pre commit handles its own vens. 

You can specify a required python version for hooks:
https://pre-commit.com/#overriding-language-version

PR #462, opened 2025-09-08
feat: Automate FABulator installation

EverythingElseWasAlreadyTaken:
Automate the installation of FABulator.

Add CLI commands to install FABulator.

Move git repo download helper from test to FABulous_CLI.helpers.

Add function to add/update values in global .env file.


FABulous/FABulous_CLI/FABulous_CLI.py
```diff
@@ -326,6 +327,34 @@ def do_install_oss_cad_suite(self, args: argparse.Namespace) -> None:
 
         install_oss_cad_suite(dest_dir, args.update_existing)
 
+    install_FABulator_parser = Cmd2ArgumentParser()
```
KelvinChung2000, line 331: Maybe at some point we should move all the parsers to a different file,  and maybe into some form of factory to compose the common parser... Perhaps a new issue

FABulous/FABulous_CLI/FABulous_CLI.py
```diff
@@ -326,6 +327,34 @@ def do_install_oss_cad_suite(self, args: argparse.Namespace) -> None:
...
+        "destination_folder",
+        type=Path,
+        help="Destination folder for the installation",
+        default="",
+        completer=Cmd.path_complete,
+        nargs=argparse.OPTIONAL,
+    )
+
+    @with_category(CMD_SETUP)
+    @allow_blank
```
KelvinChung2000, line 342: You can keep this for now, but we can stop using it if we make sure all the calls are done via the `onecmd_plus_hooks`, which can avoid the blank issue. If I recall correctly, the upstream `cmd2` has also fixed this problem.

FABulous/FABulous_CLI/helper.py
```diff
@@ -447,6 +443,7 @@ def execute(self) -> bool:
         """Execute all steps in the pipeline.
 
         Returns:
+
```
KelvinChung2000:
```suggestion
```

FABulous/FABulous_CLI/helper.py
```diff
@@ -470,3 +467,117 @@ def execute(self) -> bool:
...
+    Returns
+    -------
+        True if successful, False otherwise
+
+    Raises
+    ------
+    """
+    try:
+        logger.info(f"Downloading reference projects from {repo_url}")
```
KelvinChung2000:
```suggestion
        logger.info(f"Cloning repo from {repo_url}")
```

FABulous/FABulous_CLI/helper.py
```diff
@@ -470,3 +467,117 @@ def execute(self) -> bool:
...
+                return False
+
+            logger.info("✓ Repository cloned successfully")
+            return True
+
+    except subprocess.TimeoutExpired:
+        logger.error("Git operation timed out")
+        return False
+    except FileNotFoundError:
+        logger.error("Git command not found. Please install git.")
```
KelvinChung2000: This is likely raised from a subprocess error?

If we do the check in the first place, then we can skip this section.

FABulous/FABulous_settings.py
```diff
@@ -310,3 +314,34 @@ def reset_context() -> None:
...
+    Raises
+    ------
+        RuntimeError
+            If the user config directory is not set.
+    """
+
+    # Use user config directory for global .env file
+    user_config_dir = FAB_USER_CONFIG_DIR
+
+    if user_config_dir is None:
```
KelvinChung2000: We should not need this check, since this is "staticly" avaliable

FABulous/FABulous_CLI/helper.py
EverythingElseWasAlreadyTaken: But this would mean, that git is a hard dependency. 
But we actually just need it, if someone wants to install FABulator through the command line.

EverythingElseWasAlreadyTaken:
> Additionally, I noticed that the `run_FABulator` command currently does not check for `mvn`, which may also need adjustment.

Worst case, it would raise an error. 

We could either add a check and make it a hard dependency or just update the error message, to point to the FABulator docs for manual installation.

FABulous/FABulous_CLI/helper.py
KelvinChung2000: Let's just leave it as is.

FABulous/FABulous_CLI/helper.py
```diff
@@ -470,3 +466,147 @@ def execute(self) -> bool:
...
+    if shutil.which("mvn") is None:
+        logger.warning(
+            "Application mvn (Java Maven) not found in PATH."
+            "FABulator may not work correctly."
+        )
+
+    add_var_to_global_env("FAB_FABULATOR_ROOT", str(fabulator_dir.absolute()))
+
+
+def is_application_installed(app_name: str) -> bool:
```
KelvinChung2000: I don't think this is used

KelvinChung2000 (approved):
Just an unused function, we can merge after the clean up

FABulous/FABulous_CLI/helper.py
EverythingElseWasAlreadyTaken: Yes, this was accidentally pushed. Should be fixed now.

PR #473, opened 2025-09-10
feat: add debug env variable

KelvinChung2000:
Enable debug mode if the the `FAB_DEBUG` is set in the environment.

User input in the CLI override the settings.

### DIFF
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -94,7 +94,10 @@ def common_options(
             "--verbose", "-v", count=True, help="Show detailed log information"
         ),
     ] = 0,
-    debug: Annotated[bool, typer.Option("--debug", help="Enable debug mode")] = False,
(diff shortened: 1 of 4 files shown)


FABulous/FABulous.py
```diff
@@ -123,7 +126,7 @@ def common_options(
     """Common options for all FABulous commands."""
 
     shared_state.verbose = verbose
-    shared_state.debug = debug
+    shared_state.debug = os.getenv("FAB_DEBUG") is not None if debug is None else debug
```
KelvinChung2000, line 129: I `os.getenv` here since we cannot init the full context here. If we have enough "luncher" level settings, I will split it out from the `FABulousSettings`, and we might want to rename stuff, like "FABulous.py" to "FABulous_lucher.py" etc.

PR #475, opened 2025-09-11
feat: Improve typer Usage and testing

KelvinChung2000:
Added a lot of testing for Typer front end

Impoving the UX of the interface. 

Instead of calling 
```bash
FABulous run <project> "cmd"
```

Now become if you are not starting in the project dir
```bash
FABulous -p <project> run "cmd"
```

And if is in the project dir becomet;
```bash
FABulous run "cmd"
```

Which is much closer to what typical CLI tool expect and work with.


docs/source/Usage.rst
```diff
@@ -115,8 +121,17 @@ To use the FABulous Shell, you can run the following command:
...
-  FABulous <name_of_project>
+  FABulous create-project <name_of_project>
+
+  # to start the shell
+  cd <name_of_project>
+  FABulous start
+
+  # or
+
+  FABulous --project-dir start <name_of_project>
```
EverythingElseWasAlreadyTaken: This feels really unintuitive to specify like this. 


I would rather expect something like:
` FABulous --project-dir <name_of_project> start` 
` FABulous <name_of_project> start`
or my favorlike it was before:
` FABulous <name_of_project>`

Also for the run command in the issue description it was like this
` FABulous -p <project> run "cmd"`

docs/source/Usage.rst
```diff
@@ -115,8 +121,17 @@ To use the FABulous Shell, you can run the following command:
...
+
+If you want to use a specific RTL language (Verilog or VHDL) for your project, you can use the ``--writer`` flag when creating the project.
```
EverythingElseWasAlreadyTaken, line 137: This should be explained a bit further, that the dealt language is Verilog, and it can be switched to VHDL with the --writer. Also maybe add an example.

FABulous/FABulous_CLI/helper.py
```diff
@@ -158,28 +159,24 @@ def create_project(
...
-                f"Language template ({lang}) not found in package resources"
+                f"Language template ({lang!s}) not found in package resources"
             )
 
     except (ImportError, AttributeError) as e:
         raise FileNotFoundError(
             f"Unable to access fabric templates from package: {e}"
         ) from e
 
-    if project_dir.exists():
```
EverythingElseWasAlreadyTaken, line 177: Why is this check removed? 
Wouldn't the whole project then not just be overwritten?

FABulous/FABulous.py
```diff
@@ -115,32 +192,26 @@ def common_options(
         Path | None,
         typer.Option("--project-dot-env", "-pde", help="Set project .env file path"),
     ] = None,
-    force: Annotated[
```
EverythingElseWasAlreadyTaken, line 118: Do we kick out the force mode entirely?

FABulous/FABulous.py
```diff
@@ -74,17 +61,102 @@ def validate_project_directory(value: str) -> Path | None:
...
         exists=True,
     ),
 ]
 
+WriterType = Annotated[
+    HDLType,
+    typer.Option(
+        "--writer",
+        "-w",
+        help="Set type of HDL code generated. System Verilog is not supported yet.",
```
EverythingElseWasAlreadyTaken, line 79:
```suggestion
        help="[verilog|vhdl] Set type of HDL code generated. System Verilog is not supported yet.",
```

docs/source/Usage.rst
KelvinChung2000: This is probably a typo, the intended way is ` FABulous -p <project> start`. I can see how do I set up such that it can do `FABulous <project>` as an alias with start

FABulous/FABulous_CLI/helper.py
KelvinChung2000: I recall adding a feature that would prompt the user and ask, 'Do you want to overwrite?' However, this feature has been lost. I will need to double-check.

FABulous/FABulous.py
KelvinChung2000: If I remember correctly, the help message will include the available options, as I am giving it an enum type

FABulous/FABulous.py
KelvinChung2000: The force mode is now command specific, only for run and start command

FABulous/FABulous.py
KelvinChung2000: If you do `FABulous c --help` you will be able to see the options

FABulous/FABulous_CLI/helper.py
KelvinChung2000: The feature is actually there, on line 261 of `create_project_cmd`. This will keep the code cleaner and more general. As a user, you can now directly control what to do with the project. Like us, we are constantly creating new demos, which most of the time is fine just to get overwritten, and we can do that without needing an extra rm command before doing it again.

PR #478, opened 2025-09-12
chore: eliminate `os.chdir`

KelvinChung2000:
Make the whole application no longer rely on `os.chdir` to get correct behaviour; all relative paths within the project should be fully resolvable with the help of `get_context().proj_dir`. 

stacked on #475

### DIFF
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -250,11 +250,19 @@ def check_version_compatibility(_: Path) -> None:
 def create_project_cmd(
     project_dir: Annotated[Path, typer.Argument(help="Directory to create a project")],
     writer: WriterType = HDLType.VERILOG,
+    overwrite: bool = False,
 ) -> None:
     """Create a new FABulous project.
 
     Alias: c
     """
+    if project_dir.exists() and not overwrite:
+        delete_existing = typer.confirm(
+            f"Project directory {project_dir} already exists. Overwrite?"
+        )
+        if not delete_existing:
+            logger.info("Project creation cancelled.")
+            raise typer.Exit(0)
     create_project(project_dir, writer)
     logger.info(f"FABulous project created successfully at {project_dir}")
 


docs/source/Usage.rst
```diff
@@ -85,6 +85,12 @@ If you are using yosys oss-cad-suite, no need to install the following packages.
...
 
+.. prompt:: bash $
+
+   FABulous install-cad-tools
+
+We have also provided a command to install the above packages automatically.
+
 .. prompt:: bash $
 
    FABulous install-cad-tools
```
IAmMarcelJung: This looks doubled, is this intended?

docs/source/Usage.rst
```diff
@@ -155,12 +161,17 @@ To run the FABulous flow directly from the command line, you can use the followi
...
 
+  # -p is short for --project-dir
+  FABulous -p <name_of_project> run "run_FABulous_fabric; run_FABulous_bitstream user_design/sequential_16bit_en.v"
   # -p is short for --project-dir
   FABulous -p <name_of_project> run "run_FABulous_fabric; run_FABulous_bitstream user_design/sequential_16bit_en.v"
 
   # or r is short for run
   FABulous -p <name_of_project> r "run_FABulous_fabric; run_FABulous_bitstream user_design/sequential_16bit_en.v"
+  # or r is short for run
+  FABulous -p <name_of_project> r "run_FABulous_fabric; run_FABulous_bitstream user_design/sequential_16bit_en.v"
```
IAmMarcelJung: Same as above

docs/source/Usage.rst
```diff
@@ -89,6 +89,8 @@ We have also provided a command to install the above packages automatically.
 
    FABulous install-cad-tools
 
+We have also provided a command to install the above packages automatically.
```
IAmMarcelJung: This still looks like a duplicate

docs/source/Usage.rst
EverythingElseWasAlreadyTaken, line 68: we don't have a requirements file anymore

docs/source/Usage.rst
```diff
@@ -83,12 +83,12 @@ If you are using yosys oss-cad-suite, no need to install the following packages.
...
  Only needed for VHDL synthesis
 
-We have also provided a command to install the above packages automatically.
-
 .. prompt:: bash $
 
    FABulous install-cad-tools
 
+We have also provided a command to install the above packages automatically.
+
```
EverythingElseWasAlreadyTaken, line 91: This should be moved above the command, and the note about the oss_cad_suite should relate a bit more to the automated installation.

docs/source/Usage.rst
```diff
@@ -158,9 +158,8 @@ To run the FABulous flow directly from the command line, you can use the followi
 
   # -p is short for --project-dir
   FABulous -p <name_of_project> run "run_FABulous_fabric; run_FABulous_bitstream user_design/sequential_16bit_en.v"
-
-  # or r is short for run
-  FABulous -p <name_of_project> r "run_FABulous_fabric; run_FABulous_bitstream user_design/sequential_16bit_en.v"
+  # -p is short for --project-dir
```
EverythingElseWasAlreadyTaken: This is a duplicate from above.

docs/source/Usage.rst
```diff
@@ -83,12 +83,12 @@ If you are using yosys oss-cad-suite, no need to install the following packages.
 :`Yosys ghdl plugin <https://github.com/ghdl/ghdl-yosys-plugin>`_:
  Only needed for VHDL synthesis
 
-We have also provided a command to install the above packages automatically.
-
 .. prompt:: bash $
 
    FABulous install-cad-tools
```
EverythingElseWasAlreadyTaken, line 89: This command is not working

FABulous/FABulous.py
```diff
@@ -366,7 +348,6 @@ def script_cmd(
     # Check if script file exists before trying to execute
     if not script_file.exists():
         logger.error(f"Script file {script_file} does not exist")
-        os.chdir(entering_dir)
```
EverythingElseWasAlreadyTaken, line 369: have you checked, that we don't use any relative paths anymore in FABulous? I guess this has the implication, that we have to use project_dir as base path everywhere?

docs/source/Usage.rst
KelvinChung2000: This is from the previous CLI update... Wonder why the CI is not catching it. This should be a new issue

FABulous/FABulous.py
KelvinChung2000: I have kept for script to use the `chdir` as some existing scripts can still have relative path with the project dir implicitly. But all path should either be relative to project dir or is absolute at this point. I have check and added `absolute` to all the input for all parsing. Input from the CLI is all relative to `self.projectDir`.

docs/source/Usage.rst
EverythingElseWasAlreadyTaken: I guess the CI is only testing the `install-oss-cad-suite` command, which is working. 
I've opened an issue #488

docs/source/Usage.rst
KelvinChung2000: will do this in a new pr.

PR #480, opened 2025-10-06
fix!: Make VHDL attribute handling and belMap processing more robust …

EverythingElseWasAlreadyTaken:
…and fix broken BELs

Update Yosys_obj to keep BelMap dicts in sync with legacy bel parsing and updated yosys json parsing.

Refactor belMapProcessing and remove hardcoded handling of INIT vector. Allowing now defining BelMaps for all parameters and vector. Use the actual bit mapping value of the BelMaps to define the mapping of the config bits, instead of relying on the order of the belMap definition.

Add checks for right order of BelMap config bit matching.

BREAKING: The lowest value of a belMap vector is now defined as <parameter>[0] which was previously just <parameter> This will make the bitstream spec and bel.v2.txt more consistent.

Add a "FAB_ATTR_" prefix for BelMap values, which gets removed in the further internal processing.
This allows to define attributes with names,
which usually reserved for internal use of the synthesis tool. Like the "FF" attribute, which is internally reserved for GHDL.

Fix errors in belMaps of:
LUT4c_frame_config_dffesr.vhdl
Config_access.vhdl
OutPass4_frame_config_mux.vhdl


FABulous/fabric_definition/Yosys_obj.py
```diff
@@ -272,7 +272,7 @@ def __init__(self, path: Path) -> None:
                 "--std=08",
                 "--out=verilog",
                 str(temp),
-                f"{get_context().model_pack!s}",
+                f"{get_context().models_pack!s}",
```
KelvinChung2000: Will need to update `FABulousSettings` as well

FABulous/fabric_generator/parser/parse_hdl.py
```diff
@@ -24,10 +25,23 @@ def belMapProcessing(module_info: YosysModule) -> dict:
...
+
+    Raises
+    ------
+    ValueError
+        If any BEL mapping attribute has an invalid format or index.
     """
     belMapDic = {}
+
+    # update attributes to remove the fab_attr_ prefix, ignoring case
+    fab_attrs = [key for key in module_info.attributes if "fab_attr_" in key.casefold()]
```
KelvinChung2000, line 37: Will need to add documentation about this new "feature"

FABulous/fabric_definition/Yosys_obj.py
EverythingElseWasAlreadyTaken: Remved here.
This was added by accident. The whole model_pack handling is done in #481

FABulous/fabric_generator/parser/parse_hdl.py
EverythingElseWasAlreadyTaken: Yes, we need to document the whole VHDL attribute handling in general.

EverythingElseWasAlreadyTaken: #482

KelvinChung2000 (approved):

PR #483, opened 2025-10-09
fix: Set max version for pydantic and add warning for unresolved tool path

EverythingElseWasAlreadyTaken:
In pydantic version 2.12 the field_name property is broken, so we stick to lower versions.

Add a warning when a tool path could not be resolved.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,8 @@ dependencies = [
   'requests>=2.0.0',
   'cmd2>=2',
   'bitarray>=3.3.0',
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -26,6 +26,8 @@ dependencies = [
   'requests>=2.0.0',
   'cmd2>=2',
   'bitarray>=3.3.0',
+  # in pydantic v2.12 the fieldvlidators info.field_name property is broken
```
IAmMarcelJung:
```suggestion
  # in pydantic v2.12 the fieldvalidators info.field_name property is broken
```

PR #484, opened 2025-10-09
fix: Add check if models_pack path is absolute path

EverythingElseWasAlreadyTaken:
If its an absolute path, we should not touche it and directly throw an error if its not a file.

### DIFF
--- a/FABulous/FABulous_settings.py
+++ b/FABulous/FABulous_settings.py
@@ -84,17 +84,17 @@ class FABulousSettings(BaseSettings):
                 mp = p / "Fabric" / "my_lib.vhdl"
                 if mp.exists():
                     logger.warning(
-                        f"Model pack path is not set. Guessing model pack as: {mp}"
+                        f"Models pack path is not set. Guessing models pack as: {mp}"
                     )
                     return mp
                 mp = p / "Fabric" / "models_pack.vhdl"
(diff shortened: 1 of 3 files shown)


FABulous/FABulous_settings.py
```diff
@@ -123,9 +123,10 @@ def parse_models_pack(cls, value: Path | None, info: ValidationInfo) -> Path | N
             else:
                 raise ValueError("Project directory is not set.")
 
-            # check if project dir is in model pack path, since in the default it is
+            # check if project dir is  not absolute and
+            # in model pack path, since in the default it is
```
IAmMarcelJung:
```suggestion
            # in models pack path, since in the default it is
```

IAmMarcelJung:
```suggestion
            # Check if 'project_dir' is not an absolute path and
            # in the models pack path, since in the default it is
```

FABulous/FABulous_settings.py
```diff
@@ -123,9 +123,10 @@ def parse_models_pack(cls, value: Path | None, info: ValidationInfo) -> Path | N
...
             # put there as folder
```
IAmMarcelJung:
```suggestion
            # put there as folder.
```

PR #485, opened 2025-10-10
feat: add editor settings

KelvinChung2000:
`cmd2` come with the `edit` by default command and now I have enable it to set it from .env. 

The usage is `edit <file_name>`.

### DIFF
--- a/FABulous/FABulous_CLI/helper.py
+++ b/FABulous/FABulous_CLI/helper.py
@@ -218,9 +218,6 @@ def create_project(project_dir: Path, lang: HDLType = HDLType.VERILOG) -> None:
         str(project_dir.absolute() / "Fabric" / f"models_pack.{new_suffix}"),
     )
 
-    if e := os.getenv("EDITOR", None):
-        set_key(env_file, "FAB_EDITOR", e)
-
     logger.info(
         f"New FABulous project created in {project_dir} with {lang!s} language."
     )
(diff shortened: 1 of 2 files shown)


FABulous/FABulous_CLI/helper.py
```diff
@@ -218,6 +218,11 @@ def _copy_template_safely(template_ref: Traversable, target_dir: Path) -> None:
         str(project_dir.absolute() / "Fabric" / f"models_pack.{new_suffix}"),
     )
 
+    if e := os.getenv("EDITOR", None):
```
EverythingElseWasAlreadyTaken: We should not set a default editor here, people can set it explicitly if they want.
Instead we should rather fall back to the global $EDITOR or $VISUAL environment variable. 

In case these are not set, we should just print a warning.

FABulous/FABulous_settings.py
```diff
@@ -55,8 +55,10 @@ class FABulousSettings(BaseSettings):
         description="Deprecated, use proj_version instead",
     )
 
+    # CLI options
     debug: bool = False
     verbose: int = 0
+    editor: str = "gedit"
```
EverythingElseWasAlreadyTaken: As mentioned above, we should have use $EDITOR or $VISUAL as default.

FABulous/FABulous_CLI/helper.py
KelvinChung2000: Looks like cmd2 works even without the `$EDITOR` variable set.

KelvinChung2000: https://github.com/python-cmd2/cmd2/blob/main/cmd2/utils.py#L307

It tries very hard to find something that it can use

EverythingElseWasAlreadyTaken: ok, this is good, they even check the $EDITOR variable. But we should not set this by default in the .env file for the user while project creation. 

Either users set it themselves, then they might rather want to set it global .env, and if none is set, cmd2 should do its thing and look for something it can use.

EverythingElseWasAlreadyTaken: Also please add FAB_EDITOR to the docs.

PR #495, opened 2025-10-15
feat: new project structure

KelvinChung2000:
Part of #492


KelvinChung2000:
I want to wait for a bit to let Asma and Marcel fill in the PDK-specific tile size. But sure, just do it in another PR.

PR #498, opened 2025-10-15
feat: All the steps scripts for GDS flow

KelvinChung2000:
Part of #492

All the scripts for librelane steps.

Update to pyproject.toml for excluding `Protocol` classes which is in the form of `.*Like` as class name.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,6 +96,7 @@ dev = [
   "taskipy>=1.14.0",
   "pyyaml>=6.0.0",
   "interrogate>=1.7.0",
+  "librelane>=3.0.0.dev43",
 ]
 
 [tool.coverage.run]
--- a/uv.lock
+++ b/uv.lock
@@ -12,6 +12,20 @@ wheels = [
 ]
 
 [[package]]
+name = "anyio"
+version = "4.11.0"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "idna" },
+    { name = "sniffio" },
+    { name = "typing-extensions", marker = "python_full_version < '3.13'" },
+]
+sdist = { url = "https://files.pythonhosted.org/packages/c6/78/7d432127c41b50bccba979505f272c16cbcadcc33645d5fa3a738110ae75/anyio-4.11.0.tar.gz", hash = "sha256:82a8d0b81e318cc5ce71a5f1f8b5c4e63619620b63141ef8c995fa0db95a57c4", size = 219094, upload-time = "2025-09-23T09:19:12.58Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/15/b3/9b1a8074496371342ec1e796a96f99c82c945a339cd81a8e73de28b4cf9e/anyio-4.11.0-py3-none-any.whl", hash = "sha256:0287e96f4d26d4149305414d4e3bc32f0dcd0862365a4bddea19d7a1ec38c4fc", size = 109097, upload-time = "2025-09-23T09:19:10.601Z" },
+]
+
+[[package]]
 name = "arpeggio"
 version = "2.0.3"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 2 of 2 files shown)


FABulous/fabric_generator/gds_generator/script/add_buffers.tcl
```diff
@@ -0,0 +1,55 @@
+# Copyright 2020-2023 Efabless Corporation
```
EverythingElseWasAlreadyTaken: If this file is copied from Efabless, we should give credits and ad a link or mention it in the docs

EverythingElseWasAlreadyTaken:
This seems to be mostly stuff from Leo/Efabless.
I'm not sure how much was changed from this stuff, but it feels like we should rather import this or fork it and make it a separate module, instead of just copying the stuff.  This would allow a better maintainability, if the development moves on and also allow us to also send patches back to the initial repo, without extracting the changes manually. 
Also, this would give the credits to the original author.

KelvinChung2000:
Another problem with script importing is that it might make sourcing the script slightly more difficult, as the script is referenced by the step as a path. If scripts need importing, we will need to make sure they are located in the same file directory.

FABulous/fabric_generator/gds_generator/script/tile_io_place.py
```diff
@@ -0,0 +1,1133 @@
...
+        ]  # Remove the virtual pins from the side_pin_placement list
+
+    info(
+        f"Placement details {side} | "
+        f"{virtual_pin_count=} {actual_pin_count=} {total_pin_count=} "
+        f"possible_locations={len(possible_locations)} "
+        f"{tracks_per_pin=} {used_tracks=} {unused_tracks=} {starting_track_index=}",
+    )
+
+    VISUALIZE_PLACEMENT = False
```
EverythingElseWasAlreadyTaken: This is hardcoded to false, so the following code is unreachable

FABulous/fabric_generator/gds_generator/script/tile_io_place.py
```diff
@@ -0,0 +1,1133 @@
...
+        map_str.append("]")
+        debug("".join(map_str))
+        debug(f"Used track indices: {used_track_indices}")
+
+    return result, side_pin_placement
+
+
+identifiers = re.compile(r"\b[A-Za-z_][A-Za-z_0-9]*\b")
+standalone_numbers = re.compile(r"\b\d+\b")
+trash = re.compile(r"^[^\w\d]+$")
```
EverythingElseWasAlreadyTaken: these expressions can be private, since they are only used in sorter

FABulous/fabric_generator/gds_generator/script/tile_io_place.py
```diff
@@ -0,0 +1,1133 @@
...
+                else:
+                    rect = odb.Rect(0, 0, h_length + h_extension, h_width)
+                    if side is Side.EAST:
+                        x = BLOCK_UR_X - h_length
+                    else:
+                        x = BLOCK_LL_X - h_extension
+                    rect.moveTo(x, slot - h_width // 2)
+                    odb.dbBox_create(pin_bpin, h_layer, *rect.ll(), *rect.ur())
+
+
```
EverythingElseWasAlreadyTaken, line 1036: This file is really big and contains multiple classes, etc. it should be maybe split up.

KelvinChung2000: I have spent a bit of time doing some cleanup/inline, etc. And I don't think splitting is offering much. As most of them are like a "single" use struct for organisation. I could look into ways to simplify some of the logic if you think that will help.

EverythingElseWasAlreadyTaken: I guess we can leave it for now, but I guess this will need some kind of refactor at some point.

FABulous/fabric_generator/gds_generator/script/tile_io_place.py
```diff
@@ -40,9 +40,25 @@ def grid_to_tracks(origin: float, count: int, step: float) -> list[float]:
...
+    """Equally space pins along possible locations on a side.
+
+    Parameters
+    ----------
+    sdie: str
+        side of the floorplan being processed.
+    side_pin_placement: list[int | str]
+        the actual pin placement list for the side, including virtual pins as integers.
+    possible_locations: list[float]
+        the possible locations on the side to place pins.
```
IAmMarcelJung:
```suggestion
    side_pin_placement: list[int | str]
        The actual pin placement list for the side, including virtual pins as integers.
    possible_locations: list[float]
        The possible locations on the side to place pins.
```

tests/gds_flow_test/test_tile_io_place.py
```diff
@@ -0,0 +1,794 @@
+"""Tests for tile_io_place module."""
+# ruff: noqa: E402, SLF001, E501, F841
```
EverythingElseWasAlreadyTaken: Please don't deactivate these for the whole file.

EverythingElseWasAlreadyTaken (approved):
LGTM

PR #521, opened 2025-11-05
ci: Add workflow to publish FABulous to PyPI and cleanup pyproject.toml 

EverythingElseWasAlreadyTaken:
[ci: add publish to PyPI workflow](https://github.com/FPGA-Research/FABulous/commit/bc874a0fc0bb0ef06e70f43f76e2369864ef9c62)

[chore(pyproject.toml): Cleanup and prepare for release](https://github.com/FPGA-Research/FABulous/commit/3d0f337e1e2e1b9b1fe50f50d98610ebc8d92286)

### DIFF
--- a/.github/workflows/publish_to_pypi.yml
+++ b/.github/workflows/publish_to_pypi.yml
@@ -15,9 +15,9 @@ jobs:
         fetch-depth: 0
         persist-credentials: false
     - name: Install uv
-      uses: astral-sh/setup-uv@v6
+      uses: astral-sh/setup-uv@v7.2.1
     - name: Set up Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: "3.12"
     - name: Build a binary wheel and a source tarball
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ description = "FABulous FPGA Fabric generator"
 readme = "README.md"
 requires-python = ">=3.12"
 dynamic = ["version"]
-license = {text = "Apache-2.0"}
+license = { text = "Apache-2.0" }
 
 classifiers = [
   "Programming Language :: Python :: 3",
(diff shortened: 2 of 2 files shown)


.github/workflows/publish_to_pypi.yml
```diff
@@ -0,0 +1,62 @@
...
+    name: Build distribution
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v5
+      with:
+        fetch-depth: 0
+        persist-credentials: false
+    - name: Install uv
+      uses: astral-sh/setup-uv@v6
```
KelvinChung2000: There is a newer version of `setup-uv`  (v.7.1.2)
```suggestion
      uses: astral-sh/setup-uv@v7.1.2
```

pyproject.toml
```diff
@@ -54,11 +56,15 @@ local_scheme = "node-and-date"
 
 [tool.setuptools.packages.find]
 exclude = [
-  "docs, demo",
+  "docs",
+  "demo",
```
KelvinChung2000: We probably don't need to keep `demo` in here anymore

.github/workflows/publish_to_pypi.yml
```diff
@@ -0,0 +1,62 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v5
```
KelvinChung2000:
```suggestion
      uses: actions/setup-python@v6
```

EverythingElseWasAlreadyTaken:
> One thing to consider is that we should perform a pip install test at the end of the publish as a smoke test. Creating a project and loading Fabric should already exercise most of the package-dependent functions. 

This is not that trivial, since publishing will happen 15 minutes after the merge and will need a manual accept of one of the maintainers, for security reasons.

KelvinChung2000:
Sure, we can add a seperate workflow for testing that later

PR #526, opened 2025-11-05
chore: migration to latest software and project upgrade task

KelvinChung2000:
Update all our dependencies to ensure compatibility with the latest software. 

Solves #486  dependency issue

### DIFF
--- a/tests/fabric_gen_test/configmem_test/test_configmem_rtl.py
+++ b/tests/fabric_gen_test/configmem_test/test_configmem_rtl.py
@@ -60,7 +60,7 @@ async def initialize_configmem(dut: ConfigMemDUT) -> None:
 
 
 @cocotb.test
-async def configmem_settings_test(dut: ConfigMemDUT) -> None:
+async def cocotb_test_configmem_settings(dut: ConfigMemDUT) -> None:
     """Test exact bit mapping from FrameData to ConfigBits using direct mapping."""
     await initialize_configmem(dut)
 


tests/fabric_gen_test/configmem_test/test_configmem_rtl.py
```diff
@@ -59,9 +59,8 @@ async def initialize_configmem(dut: ConfigMemDUT) -> None:
     await Timer(10, units="ps")
 
 
-@pytest.mark.skip(reason="Cocotb test - run by simulation, not pytest")
 @cocotb.test
-async def test_configmem_settings(dut: ConfigMemDUT) -> None:
```
EverythingElseWasAlreadyTaken, line 64: why did you rename the test, I thought it is best practice to name tests with test_<>?

pyproject.toml
```diff
@@ -29,10 +29,10 @@ dependencies = [
   'cmd2>=2',
   'bitarray>=3.3.0',
   # in pydantic v2.12 the fieldvalidators info.field_name property is broken
-  "pydantic<2.12",
+  "pydantic>=2.12",
```
EverythingElseWasAlreadyTaken: in pydantic v2.12 the fieldvalidators info.field_name property is broken

make it at least 2.12.1
https://github.com/pydantic/pydantic-settings/issues/691

tests/CLI_test/test_arguments.py
```diff
@@ -771,7 +771,7 @@ def test_default_writer_is_verilog(
                 "run",
             ],
             False,
-            1,
+            0,
```
EverythingElseWasAlreadyTaken, line 774: we should at least add a warning, when run with no argument is executed, if it does not throw an error anymore.

tests/fabric_gen_test/configmem_test/test_configmem_rtl.py
KelvinChung2000: This is actually what led to the test failure. It is likely because the `pytest.mark.skip` wrapped the function, which led to `cocotb` not being able to find the test. However, I think it is better not to name them as `test_*` because we do not want them to be discovered by `pytest` anyway.

EverythingElseWasAlreadyTaken: Let us then just define something like  `cocotb_test_`  as prefix for cocotb tests

EverythingElseWasAlreadyTaken (approved):

PR #528, opened 2025-11-06
chore: Add missing dependencies and remove FASM

EverythingElseWasAlreadyTaken:
Add missing dependencies.

Remove FASM dependency, since it is only used by FABulous-bit-gen, which has it as dependency.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -325,9 +325,9 @@ name = "fabulous-fpga"
 source = { editable = "." }
 dependencies = [
     { name = "bitarray" },
+    { name = "click" },
     { name = "cmd2" },
     { name = "fabulous-bit-gen" },
-    { name = "fasm" },
     { name = "gnureadline", marker = "sys_platform != 'win32'" },
     { name = "loguru" },
     { name = "packaging" },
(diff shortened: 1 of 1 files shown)


EverythingElseWasAlreadyTaken:
> Maybe stack this on top of #526, since this lock will no longer be valid after that is merged in

We can merge it after #526.

EverythingElseWasAlreadyTaken:
> I think you might also need to update the flake lock. Lets see if it fails.

I don't think so,  I haven't seen any references to our python dependencies in the flake.lock

pyproject.toml
```diff
@@ -33,6 +32,8 @@ dependencies = [
   "packaging>=25.0",
   "typer>=0.20.0",
   "FABulous-bit-gen>=0.1.0",
+  'PyYAML>=6.0.0',
+  'click>=8.0.0',
```
KelvinChung2000, line 36: Actually, are these sub-dependencies from other packages?

EverythingElseWasAlreadyTaken: No, we use them in our code somewhere.

PR #536, opened 2025-11-10
FABulous timing model based on GDS flow

hausdinge:
Timing Model for FABulous (calculating delay times between PIPS)

Its a stacked PR and the merging should happen after the integration of the GDS flow for FABulous 2.

A better version of this model will follow which will do timing on the actual swm-mux topology extracted
from the physical netlist - its also way faster.

This model relies on synth mappings and in general it needs a huge search space making it very slow for some tiles.
I still must add some things like for some cases if files already exist or throw exceptions for such cases (error handling).

Also the code should be a bit cleaned up. This model will get frequent updated and changes.

### DIFF
--- a/FABulous/FABulous.py
+++ b/FABulous/FABulous.py
@@ -14,6 +14,8 @@ import typer
 from loguru import logger
 from packaging.version import Version
 
+import timing_model.FABulousTimingModel as FAB_TimingModel
+
 from FABulous.custom_exception import PipelineCommandError
 from FABulous.fabric_definition.define import HDLType
 from FABulous.FABulous_CLI import FABulous_CLI
@@ -207,7 +209,7 @@ def common_options(
     )
 
     if (s := ctx.invoked_subcommand) and (
-        s.startswith("install") or s == "create-project" or s == "c"
+        s.startswith("install") or s.startswith("timing") or s == "create-project" or s == "c"
     ):
         return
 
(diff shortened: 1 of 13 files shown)


FABulous/FABulous.py
```diff
@@ -207,7 +209,7 @@ def common_options(
     )
 
     if (s := ctx.invoked_subcommand) and (
-        s.startswith("install") or s == "create-project" or s == "c"
+        s.startswith("install") or s.startswith("timing") or s == "create-project" or s == "c"
```
KelvinChung2000, line 212: It is better to allow context setup, so for example, the user can override the `sta` path or log the result to somewhere else.

FABulous/FABulous.py
```diff
@@ -244,6 +246,46 @@ def check_version_compatibility(_: Path) -> None:
...
+@app.command("timing-model")
+def timing_model_cmd(
+    project_dir:            Annotated[Path,  typer.Argument(                          help="Directory to create a project")],
+    out_file:               Annotated[Path,  typer.Option("--out_file",               help="normally project_dir/.Fabulous/pips.txt")],
+    lib_corner_file:        Annotated[Path,  typer.Option("--lib_corner_file",        help="path to PDK stdcell lib file")],
+    techmaps:               Annotated[str,   typer.Option("--techmaps",               help="PDK techmaps like tribuf mux...")],
+    tiehi_cell_and_port:    Annotated[str,   typer.Option("--tiehi_cell_and_port",    help="e.g sg13g2_tiehi L_HI")],
+    tielo_cell_and_port:    Annotated[str,   typer.Option("--tielo_cell_and_port",    help="e.g sg13g2_tielo L_LO")],
+    min_buf_cell_and_ports: Annotated[str,   typer.Option("--min_buf_cell_and_ports", help="e.g sg13g2_buf_1 A X")],
+    clk_freq_mhz:           Annotated[str,   typer.Option("--clk_freq_mhz",           help="e.g 40")],
```
KelvinChung2000, line 258: I think is better to use something like `int` or `float` so you don't need to do parsing checks. But in general, we don't need to have a separate command at the top level for invoking this. The `FABulous` acts more like a "luncher".

FABulous/FABulous_API.py
```diff
@@ -125,6 +130,20 @@ def loadFabric(self, fabric_dir: Path) -> None:
...
+            path_dir = os.path.dirname(fabric_dir)
+            path = Path(f"{path_dir}/fabric.pkl")
+            obj = self.fabric
+            
+            # Use a temporary file to ensure atomic write
+            with tempfile.NamedTemporaryFile(dir=path_dir, delete=False) as tmp:
+                tmp_path = Path(tmp.name)
+                
+                pickle.dump(obj, tmp, protocol=pickle.HIGHEST_PROTOCOL)
+            os.replace(tmp_path, path)
```
KelvinChung2000, line 145: I think can simplify to `pickle.dump(self.fabric,fabric_dir / "fabric.pkl", prtocol=pickle.HIGHEST_PROTOCOL)`

timing_model/FABulousTimingModel.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+# Utility functions for serializing and deserializing objects
+# Save an object to a file using pickle
+def load_obj(path: Path):
+    with open(path, "rb") as f:
+        return pickle.load(f)
+
+
+def genNextpnrModel(fabric: Fabric, tile_names: str = "", repeat: bool = False,
```
KelvinChung2000, line 30: It feels like you can merge this into the existing `genNextpnrModel`

timing_model/Makefile
```diff
@@ -0,0 +1,9 @@
+# This Makefile runs the timing model:
```
KelvinChung2000, line 1: Can be removed

timing_model/StaConnInterface.py
```diff
@@ -0,0 +1,468 @@
...
+    clk_freq_mhz = ""         
+    tiles_dir = ""             
+    final_nl = ""               
+    nom_spef = ""               
+    output_dir = ""             
+    pips_file = ""              
+    top_name = ""               
+    extra_verilog  = ""         
+    lib_corner_file = ""        
+    techmaps = ""
```
KelvinChung2000, line 72: can be `Path` or `list[Path]`

timing_model/StaConnInterface.py
```diff
@@ -0,0 +1,468 @@
...
+import subprocess
+import sys
+from pathlib import Path
+
+# Determine the directory of this script
+# Handle cases where __file__ may not be defined
+if '__file__' in globals():
+    SCRIPT_DIR = Path(__file__).resolve().parent
+else:
+    SCRIPT_DIR = Path(sys.argv[0]).resolve().parent
```
KelvinChung2000, line 22: If you really need the script dir, access it via `importlib.resources`

timing_model/StaConnInterface.py
```diff
@@ -0,0 +1,468 @@
...
+
+class TileTimingModelGenerator:
```
KelvinChung2000, line 24: Try to use `Path` if it is a `Path`

timing_model/StaConnInterface.py
```diff
@@ -0,0 +1,468 @@
...
+
+    args = p.parse_args()
+
+    # Resolve tclsh
+    tclsh = args.tclsh
+    if os.path.sep not in tclsh:
+        found = shutil.which(tclsh)
+        if not found:
+            p.error(f"tclsh not found in PATH: {tclsh}")
+        tclsh = found
```
KelvinChung2000, line 435: If set in `FABulous_settings` , you can add a check in there easily.

timing_model/flow.tcl
```diff
@@ -0,0 +1,120 @@
+# Current script location
```
KelvinChung2000, line 1: This feels like can be a rewite as a python file, which then can be tested using `pytest`

KelvinChung2000 (changes requested):

PR #538, opened 2025-11-11
chore: Add warning for Windows systems 

EverythingElseWasAlreadyTaken:


FABulous/FABulous.py
```diff
@@ -523,6 +524,19 @@ def run_cmd(
...
+            "FABulous is not officially supported on native Windows. "
+            "You are using it at your own risk - errors may occur.\n"
+            "For the best experience on Windows-based system, "
+            "please use Windows Subsystem for Linux (WSL).\n"
+            "For more information, visit: "
+            "https://docs.microsoft.com/en-us/windows/wsl/install"
+            "For all FABulous installation options, see:"
+            "https://fabulous.readthedocs.io/en/latest/getting_started/installation/index.html"
+        )
+
```
KelvinChung2000, line 538: I think we should exit directly if this is considered an error. Alternatively, we can prompt the user to confirm they really want to continue. You can use typer https://typer.tiangolo.com/tutorial/prompt/ 

This might be annoying for users who really decide to use Windows. We could consider adding a save for future, but I think this "save for future" can be done in another PR if there is real demand.

KelvinChung2000 (changes requested):

FABulous/FABulous.py
EverythingElseWasAlreadyTaken: I didn't want to exit directly, that people can still try on their own, if they want to take the risk. 

Adding a prompt, is a good idea, will do this. 

Technically we can just save the confirmation in the .env for the project. 
This way we make sure, people confirm this for every project or even set it globally for themselves if they don't care.

PR #543, opened 2025-11-14
fix: allow standalone API use case working again

### LINKED ISSUE #542: FABulous requires `.FABulous` file
The latest version of `FABulous2.0-development` requires you to run `init_context()`, which then fails with `Value error, . is not a FABulous project [type=value_error, input_value='.', input_type=str]`.

Generally, I'm against the idea of FABulous managing environment variables and tool paths. There are already mechanisms for that, in the case of tool paths it's the `PATH` environment variable.

In any case, FABulous should not require a `.FABulous` file. If none exists, it should simply use the default tool binaries which can be set through `PATH`, for example, by Nix.

KelvinChung2000:
fixes #542

### DIFF
--- a/FABulous/FABulous_settings.py
+++ b/FABulous/FABulous_settings.py
@@ -317,32 +317,39 @@ def init_context(
     user_config_env = FAB_USER_CONFIG_DIR / ".env"
     if user_config_env.exists():
         env_files.append(user_config_env)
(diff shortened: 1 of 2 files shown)


EverythingElseWasAlreadyTaken (approved):
LGTM

PR #546, opened 2025-11-18
chore: Write models_pack path as relative in .env

EverythingElseWasAlreadyTaken:
Write the models_pack path als relative path in the project .env to allow the FABuluos project to be moved, without an error of models_pack not found. This is already supported in the models_pack field validator and need no further changes. The field validator still publishes the path as absolute path for internal usage.

Add a log message with the models pack path.

### DIFF
--- a/FABulous/FABulous_CLI/helper.py
+++ b/FABulous/FABulous_CLI/helper.py
@@ -216,7 +216,7 @@ def create_project(project_dir: Path, lang: HDLType = HDLType.VERILOG) -> None:
     set_key(
         env_file,
         "FAB_MODELS_PACK",
-        str(project_dir.absolute() / "Fabric" / f"models_pack.{new_suffix}"),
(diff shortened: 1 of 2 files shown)


FABulous/FABulous_CLI/helper.py
```diff
@@ -216,7 +216,7 @@ def _copy_template_safely(template_ref: Traversable, target_dir: Path) -> None:
     set_key(
         env_file,
         "FAB_MODELS_PACK",
-        str(project_dir.absolute() / "Fabric" / f"models_pack.{new_suffix}"),
+        str(f"{project_dir.name}/Fabric/models_pack.{new_suffix}"),
```
KelvinChung2000:
```suggestion
        str(Path(project_dir.name) / "Fabric" / f"models_pack.{new_suffix}"),
```
Maybe this is better? Since this will auto adapt to the systems seperator?

EverythingElseWasAlreadyTaken: Good point! 
I've added it.

PR #549, opened 2025-11-27
chore: gds flow doc

KelvinChung2000:
A redo cherry pick of #547 since I think I have broken something during rebase


KelvinChung2000:
Only checking the doc folder should do

PR #553, opened 2025-12-02
feat: nix based docker image

KelvinChung2000:
Adding workflow to generate Docker images based on the Nix environment. This will allow people who have no access to Nix but do have access to Docker to use the flow. There are a bit more caveats to using Docker than using Nix directly, but this is still better than nothing. This will also open up the possibility to use [devcontainer](https://github.com/devcontainers), which in turn enables running FABulous under vscode web (GitHub Codespace), which means we should be able to run FABulous anywhere in the world*. 

The package is now private, and we can set it to public once everything is good.

### DIFF
--- a/.github/workflows/nix-image.yml
+++ b/.github/workflows/nix-image.yml
@@ -74,7 +74,7 @@ jobs:
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
-          username: fpga-research
+          username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Tag and push dev image
@@ -98,7 +98,7 @@ jobs:
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
-          username: fpga-research
+          username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Push pre-release images
@@ -106,6 +106,8 @@ jobs:
         run: |
           VERSION="${GITHUB_REF_NAME#v}"
           # Push release image with pre-release version tag
+          docker tag "fabulous:latest" "${{ env.DOCKER_IMAGE }}:pre-release"
+          docker push "${{ env.DOCKER_IMAGE }}:pre-release"
           docker tag "fabulous:latest" "${{ env.DOCKER_IMAGE }}:${VERSION}"
           docker push "${{ env.DOCKER_IMAGE }}:${VERSION}"
 


.github/workflows/nix-image.yml
```diff
@@ -0,0 +1,100 @@
+name: Build and Push Docker Image
+
+on:
+  release:
```
EverythingElseWasAlreadyTaken, line 4: How are pre-releases/betas handled? 

For example, on pypi, our pre-releases/beta are automatically also tagged as pre-release, once there is an official release, you can install a pre-release version only by adding the  `--pre` flag to pip install. 

We should avoid people fetching 'latest' and get automatically dev/beta version.

.github/workflows/nix-image.yml
```diff
@@ -0,0 +1,130 @@
...
+    name: Push Dev Image
+    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: fpga-research
```
EverythingElseWasAlreadyTaken: The username should not be hardcoded here. 
Please make it a gh variable.

.github/workflows/nix-image.yml
```diff
@@ -0,0 +1,130 @@
...
+    name: Push Release Images
+    if: github.event_name == 'release'
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: fpga-research
```
EverythingElseWasAlreadyTaken: same here

.github/workflows/nix-image.yml
KelvinChung2000: As a scrate or as an evn variable?

.github/workflows/nix-image.yml
```diff
@@ -0,0 +1,130 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push pre-release images
+        if: github.event.release.prerelease == true
+        run: |
+          VERSION="${GITHUB_REF_NAME#v}"
+          # Push release image with pre-release version tag
+          docker tag "fabulous:latest" "${{ env.DOCKER_IMAGE }}:${VERSION}"
```
EverythingElseWasAlreadyTaken, line 111: This will set the fabulous latest to the pre-release version. 
Is there any way to indicate this is a pre-release? 
In the worst case, people will get a beta with bugs, but think it's stable.

.github/workflows/nix-image.yml
EverythingElseWasAlreadyTaken: I think env var is OK here. 
As it is common practice also here:
https://github.com/docker/login-action

.github/workflows/nix-image.yml
KelvinChung2000: The `fabulous:latest` is the image tag from the build. When it is pushed, the pre-release tag will be named as `2.0.0b4` etc. I could also just call it `pre-release`.

EverythingElseWasAlreadyTaken: Thats what I meant. If someone pulls fabulous:latest, they might get a beta without knowing it. 
I guess pre-release will do It.

KelvinChung2000: That `latest` tag only sits within the workflow. So the `latest` tag pulled from the ghrc will only be the stable release, and until we do a stable release, there won't be any `latest` tag getting pushed to the registry.  A thing to check is whether the pulled-down image will call the right name (I assume it should).

EverythingElseWasAlreadyTaken: ok, then it should be fine

PR #556, opened 2025-12-06
feat: add fabulator to nix

KelvinChung2000:
Including FABulator to the nix image

### DIFF
--- a/.github/workflows/nix-env.yml
+++ b/.github/workflows/nix-env.yml
@@ -42,11 +42,16 @@ jobs:
             substituters = https://cache.nixos.org https://nix-cache.fossi-foundation.org
             extra-trusted-public-keys = nix-cache.fossi-foundation.org:3+K59iFwXqKsL7BNu6Guy0v+uTlwsxYQxjspXzqLYQs=
 
-      # - name: Enable magic Nix cache
-      #   uses: DeterminateSystems/magic-nix-cache-action@v8
-
-      - name: Print Nix version
-        run: nix --version
+      - name: Free up disk space
+        run: |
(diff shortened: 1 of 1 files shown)


EverythingElseWasAlreadyTaken:
LGTM. 
Is the failing CI related to this change ?

.github/workflows/nix-env.yml
```diff
@@ -42,11 +42,16 @@ jobs:
...
+      - name: Free up disk space
+        run: |
+          # Remove unnecessary tools and libraries
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /opt/hostedtoolcache
+          sudo rm -rf /usr/local/share/boost
+          sudo rm -rf /usr/lib/jvm
+          sudo rm -rf /usr/share/swift
```
IAmMarcelJung, line 54: Do we actually need all of this here? E.g. the android and swift stuff we shouldn't have anyways? I know that if it's not there it doesn't really hurt and will be skipped, but keeping it in the minimum required setup should still be the way to go.

KelvinChung2000: This comes with the container in the first place, since those are the things other people tend to need, and GitHub packages them into the runner image. The problem is that having them, the runner will run out of storage, and we will not be able to build the image. A better solution would be to use a runner with more storage, but that would mean paying.

IAmMarcelJung: Ah okay didn't know that these already ship with the container! Then ofc we have to remove them too.

PR #559, opened 2025-12-10
feat: Devcontainer and codespaces

KelvinChung2000:
Setting up a devcontainer from Nix is non-trivial. I have updated the implementation, which now has a base Ubuntu Docker image with Nix installed and set up the Nix environment. The result is that the image is larger, but this offers much more usability as a general image, as users can do things like `apt install` in the container to set it up the way they like.

With the base image-based approach, we can also provide GitHub Codespaces, enabling users to try out FABulous on the web. I have put some effort into getting the GUI working as well. 

I have added some documentation. The codespace documentation is largely commented, and once the feature is merged, I will include figures to guide users in using it, since it is not as simple as clicking one button and everything works.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -38,15 +38,18 @@ RUN nix develop --profile /nix/var/nix/profiles/devshell --command true && \
 # Switch back to root for system-level setup
 USER root
 
-# Set up workspace, runtime dir, and devshell profile script
+# Set up workspace and runtime dir
 RUN mkdir -p /workspaces /run/user/1000 && \
     chown nixuser:nixuser /workspaces /run/user/1000 && \
-    chmod 700 /run/user/1000 && \
-    { echo '#!/bin/bash'; \
-    echo 'DEVSHELL_DIR=${DEVSHELL_DIR:-$(find /nix/store -maxdepth 1 -type d -name "*-devshell-dir" 2>/dev/null | head -1)}'; \
-    echo '[ -n "$DEVSHELL_DIR" ] && [ -f "$DEVSHELL_DIR/env.bash" ] && source "$DEVSHELL_DIR/env.bash"'; \
-    echo 'export PATH="/nix/var/nix/profiles/devshell/bin:/home/nixuser/.nix-profile/bin:$PATH"'; \
-    } > /etc/profile.d/devshell.sh && \
+    chmod 700 /run/user/1000
+
+# Install devshell profile script, resolving DEVSHELL_DIR at build time.
+# Sourced via BASH_ENV (non-interactive), /etc/bash.bashrc (system interactive),
+# and ~/.bashrc (user interactive) to cover all shell invocation modes.
+COPY .devcontainer/devshell.sh.tpl /tmp/devshell.sh.tpl
+RUN DEVSHELL_DIR=$(find /nix/store -maxdepth 1 -type d -name "*-devshell-dir" | head -1) && \
+    sed "s|@@DEVSHELL_DIR@@|$DEVSHELL_DIR|" /tmp/devshell.sh.tpl > /etc/profile.d/devshell.sh && \
+    rm /tmp/devshell.sh.tpl && \
     chmod +x /etc/profile.d/devshell.sh && \
     echo 'source /etc/profile.d/devshell.sh' >> /home/nixuser/.bashrc && \
     echo 'source /etc/profile.d/devshell.sh' >> /etc/bash.bashrc && \
@@ -59,14 +62,8 @@ RUN rm -f /etc/apt/apt.conf.d/docker-gzip-indexes && \
     rm -rf /var/lib/apt/lists/* && \
     apt-get update
 
-# Initialize git repo in workspace as nixuser
+# Run as nixuser by default and set environment variables
 USER nixuser
-RUN cd /workspaces && \
-    git init && \
-    git config user.email "devcontainer@fabulous.local" && \
-    git config user.name "Dev Container"
-
-# Set environment variables
 ENV BASH_ENV=/etc/profile.d/devshell.sh \
     XDG_RUNTIME_DIR=/run/user/1000 \
     QT_X11_NO_MITSHM=1 \
--- a/.devcontainer/codespace/devcontainer.json
+++ b/.devcontainer/codespace/devcontainer.json
@@ -15,7 +15,7 @@
     "customizations": {
         "vscode": {
             "settings": {
-                "python.defaultInterpreterPath": "/nix/var/nix/profiles/default/bin/python",
+                "python.defaultInterpreterPath": "python",
                 "python.analysis.typeCheckingMode": "basic",
                 "editor.formatOnSave": true,
                 "[python]": {
--- a/.devcontainer/local/devcontainer.json
+++ b/.devcontainer/local/devcontainer.json
@@ -29,7 +29,7 @@
     "customizations": {
         "vscode": {
             "settings": {
-                "python.defaultInterpreterPath": "/nix/var/nix/profiles/default/bin/python",
+                "python.defaultInterpreterPath": "python",
                 "python.analysis.typeCheckingMode": "basic",
                 "editor.formatOnSave": true,
                 "[python]": {
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -72,6 +72,7 @@ jobs:
           file: .devcontainer/Dockerfile
           tags: ${{ steps.tags.outputs.tags }}
           push: ${{ github.event_name != 'pull_request' }}
+          load: true
 
       - name: Test Docker image
         if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
--- a/.devcontainer/devshell.sh.tpl
+++ b/.devcontainer/devshell.sh.tpl
@@ -1,0 +1,8 @@
+# Sourced by BASH_ENV, /etc/bash.bashrc, and ~/.bashrc
+# _DEVSHELL_SOURCED guard prevents re-sourcing in nested shells
+[ -n "$_DEVSHELL_SOURCED" ] && return
+export _DEVSHELL_SOURCED=1
+
+export DEVSHELL_DIR="@@DEVSHELL_DIR@@"
+[ -f "$DEVSHELL_DIR/env.bash" ] && source "$DEVSHELL_DIR/env.bash"
+export PATH="/nix/var/nix/profiles/devshell/bin:/home/nixuser/.nix-profile/bin:$PATH"
--- a/docs/source/developer_guide/development.md
+++ b/docs/source/developer_guide/development.md
@@ -99,7 +99,7 @@ For contributors who prefer a containerized development environment, FABulous pr
 
 ### Working in the Container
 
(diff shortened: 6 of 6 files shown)


.devcontainer/codespace/devcontainer.json
```diff
@@ -0,0 +1,48 @@
...
+            ]
+        },
+        "codespaces": {
+            "openFiles": []
+        }
+    },
+    "postCreateCommand": "bash .devcontainer/post-create.sh",
+    "postAttachCommand": "bash .devcontainer/display-switch.sh",
+    "initializeCommand": "xhost +local:docker 2>/dev/null || true",
+    "remoteUser": "root",
```
EverythingElseWasAlreadyTaken: Is it needed that this runs with root rights?

Otherwise, ruse a non-root user and create one in the Dockerfile if needed.

.devcontainer/codespace/devcontainer.json
```diff
@@ -0,0 +1,48 @@
+{
+    "name": "Remote (codespace)",
+    "image": "ghcr.io/fpga-research/fabulous:dev",
+    "features": {
+        "ghcr.io/devcontainers/features/desktop-lite:1": {
+            "password": "fabulous",
```
EverythingElseWasAlreadyTaken, line 6: Replace the hardcoded VNC password with a secret . 
Codespaces supports CODESPACE_* secrets / environment secrets

.devcontainer/codespace/devcontainer.json
```diff
@@ -0,0 +1,48 @@
...
+            "webPort": "6080",
+            "vncPort": "5901"
```
EverythingElseWasAlreadyTaken, line 8: This can be an security issue, in Codespaces ports can be public.

AI suggests something like this:

```
Add portsAttributes to the customizations.codespaces section:
json"customizations": {
    "codespaces": {
        "openFiles": [],
        "portsAttributes": {
            "6080": {
                "label": "Desktop (noVNC)",
                "onAutoForward": "ignore",
                "visibility": "private"
            },
            "5901": {
                "label": "VNC",
                "onAutoForward": "ignore",
                "visibility": "private"
            }
        }
    }
}
"onAutoForward": "ignore" prevents VS Code from popping up a notification and auto-opening the port. "visibility": "private" ensures only you (authenticated with GitHub) can access it — not anyone with the URL.
```

.devcontainer/codespace/devcontainer.json
```diff
@@ -0,0 +1,48 @@
+{
+    "name": "Remote (codespace)",
+    "image": "ghcr.io/fpga-research/fabulous:dev",
```
EverythingElseWasAlreadyTaken, line 3: should we always point to dev tag here? 
I guess something like stable or a fixed version would be better for default, that the version not changes unexpectedly. 
 
But this can be changed later, when 2.0 is officially released.

.devcontainer/local/devcontainer.json
```diff
@@ -0,0 +1,47 @@
+{
+    "name": "Local",
+    "image": "ghcr.io/fpga-research/fabulous:dev",
```
EverythingElseWasAlreadyTaken, line 3: See previous file.

.devcontainer/local/devcontainer.json
```diff
@@ -0,0 +1,47 @@
...
+    "features": {
+        "ghcr.io/devcontainers/features/desktop-lite:1": {
+            "password": "fabulous",
+            "webPort": "6080",
+            "vncPort": "5901"
```
EverythingElseWasAlreadyTaken, line 8: see previous file

.devcontainer/local/devcontainer.json
```diff
@@ -0,0 +1,47 @@
...
+                    "editor.codeActionsOnSave": {
+                        "source.fixAll.ruff": "explicit",
+                        "source.organizeImports.ruff": "explicit"
+                    }
+                },
+                "terminal.integrated.cwd": "/workspaces/${localWorkspaceFolderBasename}"
+            }
+        }
+    },
+    "remoteUser": "root"
```
EverythingElseWasAlreadyTaken: same as before

.devcontainer/local/devcontainer.json
```diff
@@ -0,0 +1,47 @@
...
+        }
+    },
+    "workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
+    // --- LOCAL LINUX SPECIFIC SETTINGS ---
+    "runArgs": [
+        "--net=host",
+        "--ipc=host",
+        "--env=DISPLAY",
+        "--security-opt=label=disable"
+    ],
```
EverythingElseWasAlreadyTaken, line 19: These settings bring some security risks, if a container might get compromised. 

Especially in combinations with the VNC settings before, since now the whole container network stack, including VNC is exposed to the network. 
Also the container has access rights to X11, runs as root, has access to the whole network, and IPC namespace. 

Instead of exposing the whole network stack, only open ports you need.
Is IPC shared memory needed? 
Why is security-opt disables?

.devcontainer/local/devcontainer.json
```diff
@@ -0,0 +1,47 @@
...
+    "mounts": [
```
EverythingElseWasAlreadyTaken, line 20: Instead of  X forwarding, we could consider using noVNC.

.devcontainer/display-switch.sh
```diff
@@ -0,0 +1,38 @@
...
+# 1. GitHub Codespaces (Keep VNC)
+if [ "$CODESPACES" = "true" ]; then
+    echo "☁️  Running in Codespaces."
+    export DISPLAY=:1
+    if ! grep -q "export DISPLAY=:1" ~/.bashrc; then
+        echo "export DISPLAY=:1" >> ~/.bashrc
+    fi
+
+# 2. Local Linux (Network Host Mode)
+else
```
EverythingElseWasAlreadyTaken, line 10: should be also kicked out, when we kick out X forwarding.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,63 @@
+FROM ubuntu:24.04
+
+# Install minimal dependencies and set up Nix user in one layer
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    curl ca-certificates xz-utils git sudo && \
+    useradd -m -s /bin/bash nixuser && \
+    mkdir -p /nix && chown -R nixuser:nixuser /nix
+
+# Install Nix as nixuser and configure for root
+RUN su - nixuser -c 'curl -L https://nixos.org/nix/install | sh -s -- --no-daemon' && \
```
EverythingElseWasAlreadyTaken, line 14: Why do we do this setup on our own, can't we just use a pre-build nix container instead?

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,84 @@
+name: Build and Push Docker Image
```
EverythingElseWasAlreadyTaken, line 1: please rename the file to something more descriptive.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+    mkdir -p /root/.nix-profile /root/.config/nix && \
+    ln -s /nix/var/nix/profiles/per-user/nixuser/profile /root/.nix-profile && \
+    chown -R root:root /nix
```
EverythingElseWasAlreadyTaken: Running everything as root, can be also some security concern.

docs/source/getting_started/codespaces.md
```diff
@@ -0,0 +1,158 @@
...
+## Quick Start
+
+1. Navigate to the [FABulous repository](https://github.com/FPGA-Research/FABulous) on GitHub
+2. Click the green **"Code"** button
+3. Select the **"Codespaces"** tab
+4. Click **"Create codespace on main"** (or your desired branch)
+
+GitHub will create and configure your development environment automatically.
+
+<!-- ## Selecting the Correct Configuration
```
EverythingElseWasAlreadyTaken, line 15: Why is this commented out?

KelvinChung2000: I want to include an image later, once this is up. Since the interface, when it is up in main, is different from when it is not in main.

.devcontainer/Dockerfile
KelvinChung2000: Being the root of a Docker should be fine?

.devcontainer/Dockerfile
KelvinChung2000: We cannot use NixOS, as that does not work with the dev env, as all the usual Ubuntu tools are gone. So what this setup does is like what we have, a base with Nix as the package manager

.devcontainer/codespace/devcontainer.json
KelvinChung2000: Yes, until the next stable, there won't be a stable Docker release.

.devcontainer/codespace/devcontainer.json
KelvinChung2000: I recall VNC user account is root.

.devcontainer/local/devcontainer.json
KelvinChung2000: I prefer X forwarding because it provides a native window. I will see if we can do it with more security in mind.

.devcontainer/Dockerfile
EverythingElseWasAlreadyTaken: Depends, if the container does not mount anything, and we don't have open ports, it's OK.
 
But since the dev container is based on this, it could be a risk.

.devcontainer/Dockerfile
EverythingElseWasAlreadyTaken: ok, makes sense.

.devcontainer/codespace/devcontainer.json
KelvinChung2000: https://github.com/devcontainers/features/issues/1168

So seems like password is not enforced

.devcontainer/codespace/devcontainer.json
KelvinChung2000: https://docs.github.com/en/codespaces/reference/security-in-github-codespaces

> All forwarded ports are private by default, which means that you will need to authenticate before you can access the port. Access to a codespace's private forwarded ports is controlled by authentication cookies with a 3-hour expiry period. When the cookie expires, you will need to reauthenticate.

.devcontainer/local/devcontainer.json
KelvinChung2000: the net host is required for unix based X11 forwarding and the security opt is for RHEL. But still, need to find a better way

KelvinChung2000:
I believe everything is working and safe. I will let AI do another round of checks to make sure we didn't miss anything. 

So to test, you can go into codespace button at home, and do `New with options...`:
image

Then you use the following settings, since this is not merged yet. 
image

Then this will bring you to an online instance of vscode, and you can go to the port and enter the port `6080`
image

And click open in browser, which will bring you to noVNC.

And you can in the VSCode terminal to type `openroad -gui`, this will spawn the GUI in the noVNC tab. 

For testing in local just `ctrl+shift+p` to get to the command pattlet and do the following option:
image

This will pull from the dev image from the GitHub registry.

PR #565, opened 2025-12-16
feat: extending gds helper and testing it

KelvinChung2000:
Extracted from `the-flow` branch, to reduce noise.

### DIFF
--- a/FABulous/fabric_generator/gds_generator/helper.py
+++ b/FABulous/fabric_generator/gds_generator/helper.py
@@ -116,12 +116,17 @@ def get_routing_obstructions(
     Parameters
     ----------
     config : Config
-        The configuration object from liberlane.config.config
+        The configuration object from liberlane.
 
     Returns
     -------
-    list[tuple[str, Decimal | int, Decimal | int, Decimal | int, Decimal | int]]
+    list[tuple[str, Decimal, Decimal, Decimal, Decimal]]
         A list of obstruction tuples.
+
+    Raises
+    ------
+    ValueError
+        If the entry is not a valid obstruction.
     """
     obstructions = config.get("ROUTING_OBSTRUCTIONS") or []
     _, _, width, height = config["DIE_AREA"]


FABulous/fabric_generator/gds_generator/helper.py
```diff
@@ -100,10 +116,10 @@ def get_routing_obstructions(config: Config) -> list[tuple[int, int, int, int]]:
 
     parsed_obstructions = defaultdict(list)
     for obs in obstructions:
-        if len(obs) != 4:
+        if len(obs) != 5:
             raise ValueError(
                 f"Invalid obstruction {obs}. Each obstruction must be a tuple of "
-                "4 integers."
+                "metal layer flowed by 4 decimals"
```
IAmMarcelJung:
```suggestion
                "the metal layer name followed by 4 decimals"
```

I think you meant this (plus a short extension of the description)?

tests/gds_flow_test/test_gds_generator_helper.py
```diff
@@ -0,0 +1,347 @@
...
+        mock_config.__getitem__.side_effect = lambda key: {
+            "DIE_AREA": (0, 0, 100, 100),
+            "FP_IO_VLAYER": "M1",
+            "FP_IO_HLAYER": "M1",
+        }.get(key)
+
+        result = get_routing_obstructions(mock_config)
+
+        # When both layers are the same, the function checks VLAYER first (lines 111-114)
+        # and adds horizontal obstructions. Then it checks HLAYER (lines 116-119) but layer
```
IAmMarcelJung: I think keeping track of the exact lines will be hard to keep track of if something changes.

tests/gds_flow_test/test_gds_generator_helper.py
```diff
@@ -0,0 +1,347 @@
...
+            **kwargs,
+        }
+
+        result = round_die_area(mock_config)
+
+        # Result should have DIE_AREA with rounded dimensions
+        assert result["DIE_AREA"][0] == 0
+        assert result["DIE_AREA"][1] == 0
+        # Width: 100/10 = 10, 10/0.28 = 35.714..., round up to 36, 36 * 0.28 * 10 = 100.8
+        # Height: 200/10 = 20, 20/0.56 = 35.714..., round up to 36, 36 * 0.56 * 10 = 201.6
```
IAmMarcelJung: Probably I'm not getting it right, but shouldn't this round up to the next multiple of 0.28/0.56 for 100/200, which would be 100.24/200.48?

FABulous/fabric_generator/gds_generator/helper.py
IAmMarcelJung: BTW we should put the obstructions on all routing layers, since it can otherwise still cause DRC errors. And to have it fully optimized it has to be set half of min distance away from the macro for each routing layer.

tests/gds_flow_test/test_gds_generator_helper.py
```diff
@@ -0,0 +1,347 @@
...
+        }.get(key)
+
+        result = get_routing_obstructions(mock_config)
+
+        # Should have default M1 horizontal obstructions
+        assert ("M1", 0, -1, 100, 0) in result
+        assert ("M1", 0, 100, 100, 101) in result
+        # Should have default M2 vertical obstructions
+        assert ("M2", -1, 0, 0, 100) in result
+        assert ("M2", 100, 0, 101, 100) in result
```
IAmMarcelJung: Other than the length check and the comments I don't see the difference even when diffing the lines, am I missing something?

tests/gds_flow_test/test_gds_generator_helper.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def test_get_routing_obstructions_invalid_obstruction(self, mock_config):
+        """Test that ValueError is raised for invalid obstructions."""
+        obstructions = [("M1", 10, 10)]  # Only 3 values instead of 4
```
IAmMarcelJung:
```suggestion
        obstructions = [("M1", 10, 10)]  # Only 3 values instead of 5
```

tests/gds_flow_test/test_gds_generator_helper.py
```diff
@@ -0,0 +1,347 @@
...
+        mock_config.get.return_value = None
+        mock_config.__getitem__.side_effect = lambda key: {
+            "DIE_AREA": (0, 0, 100, 100),
+            "FP_IO_VLAYER": "M1",
+            "FP_IO_HLAYER": "M2",
+        }.get(key)
+
+        result = get_routing_obstructions(mock_config)
+
+        # Should have default obstructions for M1 and M2
```
IAmMarcelJung: Technically speaking the obstructions are not the defaults for the specific metal layers, just for the IO pin layers.

tests/gds_flow_test/test_gds_generator_helper.py
IAmMarcelJung:
```suggestion
        obstructions = [("M1", 10, 10)]  # Only 3 values instead of 5
```

tests/gds_flow_test/test_gds_generator_helper.py
```diff
@@ -0,0 +1,347 @@
...
+        result = get_routing_obstructions(mock_config)
+
+        # When both layers are the same, the function checks VLAYER first (lines 111-114)
+        # and adds horizontal obstructions. Then it checks HLAYER (lines 116-119) but layer
+        # is already in parsed_obstructions, so no vertical obstructions are added.
+        m1_obstructions = [obs for obs in result if obs[0] == "M1"]
+        # Should have only 2 obstructions for M1 (the horizontal ones, vertical not added)
+        assert len(m1_obstructions) == 2
+        assert ("M1", 0, -1, 100, 0) in result
+        assert ("M1", 0, 100, 100, 101) in result
```
IAmMarcelJung: I think this is not the intended behavior. In case we do put the pins on the same layer, the obstruction should be put on all sides in that layer, since otherwise we might get DRC violations on the sides where we didn't put obstructions.

IAmMarcelJung (changes requested):
Thanks for the PR and for extracting it from the other PR! It makes reviewing quite a bit easier and is also useful to keep track of the changes if needed at a later point in time. The comments range from minor stuff to potentially untended behavior of the current code and tests.

EverythingElseWasAlreadyTaken (approved):
lgtm

FABulous/fabric_generator/gds_generator/helper.py
```diff
@@ -47,6 +47,22 @@ def get_pitch(config: Config) -> tuple[Decimal, Decimal]:
     return x_pitch, y_pitch
 
 
+def get_offset(config: Config) -> tuple[Decimal, Decimal]:
+    """Read the FP_TRACKS_INFO file and return min pitches for X and Y.
+
+    Returns a tuple (x_pitch, y_pitch) where x_pitch is the minimum pitch along X-axis
+    (FP_IO_VLAYER X direction) and y_pitch is minimum pitch along Y-axis (FP_IO_HLAYER Y
```
IAmMarcelJung:
```suggestion
    """Read the FP_TRACKS_INFO file and return the offset for X and Y.

    Returns a tuple (x_offset, y_offset) where x_offset is the offset along X-axis
    (FP_IO_VLAYER X direction) and y_pitch is the offset along Y-axis (FP_IO_HLAYER Y
```

PR #566, opened 2025-12-16
fix: Fix tile io place script and extend script testing

KelvinChung2000:
Extracted from `the-flow` branch to reduce noise.


EverythingElseWasAlreadyTaken (approved):
lgtm

PR #500, opened 2025-10-15
feat: Librelane flows and commands

KelvinChung2000:
Part of #492

Stacked on #499

Adding all the commands and flows, except for the full automation flow, which need a bit more checking.

PR #572, opened 2025-12-19
feat: Extending CLI utils for gds flow

KelvinChung2000:
extracted from #500


FABulous/processpool.py
```diff
@@ -0,0 +1,44 @@
...
+import multiprocessing
+from concurrent.futures import ProcessPoolExecutor
+from typing import Any
+
+import dill
+
+
+def _init_worker() -> None:
+    """Initialize worker process to use dill for pickling."""
+    from multiprocessing.reduction import ForkingPickler
```
EverythingElseWasAlreadyTaken: move to top

FABulous/processpool.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    def __init__(
+        self,
+        max_workers: int | None = None,
+        initargs: tuple[Any, ...] = (),
+        max_tasks_per_child: int | None = None,
+    ) -> None:
+        # Patch the main process to use dill BEFORE calling parent init
+        from multiprocessing.reduction import ForkingPickler
```
EverythingElseWasAlreadyTaken: here too

FABulous/processpool.py
```diff
@@ -0,0 +1,44 @@
...
+
+        ForkingPickler.dumps = dill.dumps
+        ForkingPickler.loads = dill.loads
```
EverythingElseWasAlreadyTaken, line 33: feels  redundant with the init_worker but not sure how this is handled internally.

KelvinChung2000: Forking a process involves pickling the data structure and sending it to the other process. This basically overrides the default pickler to use the dill pickler.

Issue #416, opened 2025-08-08
Make Buffer Insertion Optional


EverythingElseWasAlreadyTaken:
Not all users need the inserted buffers, since they need to be constrained properly.

We should make the buffer insertion optional.

PR #581, opened 2026-01-27
feat: allow disable UserCLK port adding

KelvinChung2000:
Added the `DisableUserCLK` attribute, and setting it to `TRUE` can disable that port from being added.

### DIFF
--- a/FABulous/fabric_generator/gen_fabric/gen_fabric.py
+++ b/FABulous/fabric_generator/gen_fabric/gen_fabric.py
@@ -376,9 +376,7 @@ def generateFabric(writer: CodeGenerator, fabric: Fabric) -> None:
                 else:
                     for i, j in tileLocationOffset:
                         # prefix for super tile port
-                        pre = ""
-                        if superTile:
-                            pre = f"Tile_X{i}Y{j}_"
+                        pre = f"Tile_X{i}Y{j}_"
 
                         # UserCLK signal
                         if (


FABulous/fabric_generator/gen_fabric/gen_fabric.py
```diff
@@ -353,44 +355,49 @@ def generateFabric(writer: CodeGenerator, fabric: Fabric) -> None:
...
-                        portsPairs.append(
-                            (f"{pre}UserCLK", f"Tile_X{x + i}Y{y + j + 1}_UserCLKo")
-                        )
+                    # for userCLKo
+                    portsPairs.append(("UserCLKo", f"Tile_X{x}Y{y}_UserCLKo"))
+                else:
+                    for i, j in tileLocationOffset:
+                        # prefix for super tile port
+                        pre = ""
+                        if superTile:
```
EverythingElseWasAlreadyTaken: this is always true

FABulous/fabric_generator/gen_fabric/gen_tile.py
```diff
@@ -354,11 +355,12 @@ def generateTile(writer: CodeGenerator, fabric: Fabric, tile: Tile) -> None:
 
             added.add((port.sourceName, port.destinationName))
 
-    writer.addInstantiation(
-        "clk_buf",
-        "inst_clk_buf",
-        portsPairs=[("A", "UserCLK"), ("X", "UserCLKo")],
-    )
+    if not fabric.disableUserCLK:
```
EverythingElseWasAlreadyTaken, line 358: Not an issue for now, but in general should we make the buffer insertion also deactivatable.

KelvinChung2000: Sure

EverythingElseWasAlreadyTaken: I remembered, that we had already an issue for this #416

PR #589, opened 2026-02-10
chore: gds flow doc

KelvinChung2000:
redo of #549

### DIFF
--- a/docs/source/_ext/generate_configvar_docs.py
+++ b/docs/source/_ext/generate_configvar_docs.py
@@ -13,6 +13,13 @@ logger = logging.getLogger(__name__)
 # Base project directory resolved from this extension's location
 _PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent.parent
 
+# Scope labels used to categorize settings by their variable name prefix.
+_SCOPE_GLOBAL = "Global Environment Variables"
+_SCOPE_PROJECT = "Project Specific Environment Variables"
+
+# Display order for scopes (global before project-specific).
+_SCOPE_ORDER = [_SCOPE_GLOBAL, _SCOPE_PROJECT]
+
 
 def setup(app: Sphinx) -> dict[str, str]:  # noqa: ARG001
     """Set up the Sphinx extension.
@@ -234,17 +241,53 @@ def get_user_value_example(field_type: str) -> str:
     return "-"
 
 
-def extract_fabulous_settings() -> dict:
-    """Extract configuration variables from FABulousSettings class using AST parsing.
+def _sort_settings_by_scope(
+    settings: dict[str, dict[str, list]],
+) -> dict[str, dict[str, list]]:
+    """Return *settings* ordered by ``_SCOPE_ORDER`` with sorted subcategories.
 
-    Returns a dictionary with categorized settings variables with descriptions.
-    Categories are determined by the 'title' field in Field() calls. Variables without a
-    title are placed in the 'Miscellaneous' category.
+    Within each scope the subcategories are sorted alphabetically, except
+    "General" which is placed last (it acts as the catch-all bucket).
+
+    Parameters
+    ----------
+    settings : dict[str, dict[str, list]]
+        Unsorted settings produced by the categorization loop.
 
     Returns
     -------
-    dict
-        Dictionary of settings by category.
+    dict[str, dict[str, list]]
+        A new dictionary with deterministic ordering.
+    """
+    result: dict[str, dict[str, list]] = {}
+
+    for scope in _SCOPE_ORDER:
+        if scope not in settings:
+            continue
+        subcats = settings[scope]
+        sorted_subcats = dict(
+            sorted((k, v) for k, v in subcats.items() if k != "General")
+        )
+        general = subcats.get("General")
+        if general is not None:
+            sorted_subcats["General"] = general
+        result[scope] = sorted_subcats
+
+    return result
+
+
+def extract_fabulous_settings() -> dict[str, dict[str, list]]:
+    """Extract configuration variables from FABulousSettings class using AST parsing.
+
+    Returns a two-level nested dictionary: outer key is the scope
+    ("Global Environment Variables" / "Project Specific Environment Variables"),
+    inner key is the subcategory from the ``title`` field in ``Field()`` calls
+    (falling back to "General").
+
+    Returns
+    -------
+    dict[str, dict[str, list]]
+        Nested dictionary of settings by scope and subcategory.
     """
     settings_file = _PROJECT_ROOT / "fabulous" / "fabulous_settings.py"
     source = settings_file.read_text()
@@ -261,13 +304,14 @@ def extract_fabulous_settings() -> dict:
                     if field_info and not field_info["deprecated"]:
                         field_info_list.append(field_info)
 
-    # Dynamically categorize settings based on title field
-    settings: dict[str, list] = {}
+    # Two-level categorization: scope (by prefix) -> subcategory (by title)
+    settings: dict[str, dict[str, list]] = {}
 
     for info in field_info_list:
-        category = info["title"] or "Miscellaneous"
+        scope = _SCOPE_PROJECT if info["name"].startswith("proj_") else _SCOPE_GLOBAL
+        subcategory = info["title"] or "General"
 
-        settings.setdefault(category, []).append(
+        settings.setdefault(scope, {}).setdefault(subcategory, []).append(
             {
                 "name": info["name"],
                 "env_var": f"FAB_{info['name'].upper()}",
@@ -278,13 +322,7 @@ def extract_fabulous_settings() -> dict:
             }
         )
 
-    # Sort categories alphabetically, but keep "Miscellaneous" at the end
-    misc = settings.pop("Miscellaneous", None)
-    sorted_settings = dict(sorted(settings.items()))
-    if misc is not None:
-        sorted_settings["Miscellaneous"] = misc
-
-    return sorted_settings
+    return _sort_settings_by_scope(settings)
 
 
 def extract_cli_settables() -> list:
--- a/docs/source/_templates/flow_variable.md.jinja
+++ b/docs/source/_templates/flow_variable.md.jinja
@@ -1,18 +1,15 @@
-(fabulous-variables)=
-# FABulous Configuration Variables
+{% for scope, subcategories in settings_vars.items() %}
+## {{ scope }}
 
-This is an auto-generated reference of all FABulous configuration variables.
+{% for subcategory, variables in subcategories.items() %}
+{% if subcategory != "General" %}
+### {{ subcategory }}
 
-## Environment Variables
-
-FABulous settings can be configured via environment variables with the `FAB_` prefix. These can be set in your shell or in a `.env` file in your project's `.FABulous` directory.
-{% for category, variables in settings_vars.items() %}
-
-### {{ category }}
-
+{% endif %}
 | Variable | Environment Variable | Type | Default | User Value | Description |
 |----------|---------------------|------|---------|------------|-------------|
-{% for var in variables %}| `{{ var.name }}` | `{{ var.env_var }}` | {{ var.type }} | {{ var.default if var.default else "-" }} | {{ var.user_value if var.user_value else "-" }} | {{ var.description if var.description else "-" }} |
+{% for var in variables %}| `{{ var.name }}` | `{{ var.env_var }}` | {{ var.type }} | {{ var.default | default("-", true) }} | {{ var.user_value | default("-", true) }} | {{ var.description | default("-", true) }} |
+{% endfor %}
 {% endfor %}
 {% endfor %}
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -9,12 +9,14 @@ project = "FABulous: an Embedded FPGA Framework and CAD Tools"
 copyright = "2021, University of Manchester"
 author = "Jing, Nguyen, Bea, Bardia, Dirk"
 
+
 # Automated version management from installed package
 def get_version():
     """Get version from installed package, git tags, or commit ID."""
     # Try to get version from the installed package
     try:
         from importlib.metadata import version
+
         return version("fabulous-fpga")
     except Exception:
         pass
@@ -50,6 +52,7 @@ def get_version():
     # Fallback to unknown version
     return "unknown"
 
+
 version = get_version()
 release = version
 
@@ -76,10 +79,8 @@ extensions = [
     "sphinx.ext.napoleon",
     "sphinx.ext.viewcode",
     "sphinx.ext.imgconverter",
-
     # Modern documentation automation
     "autoapi.extension",  # Keep existing AutoAPI
-
     # Enhanced documentation features (scikit-learn additions)
     "myst_parser",  # Markdown support
     "sphinx_design",  # Modern UI components
@@ -87,10 +88,8 @@ extensions = [
     "sphinx_copybutton",  # Copy code button
     "sphinx_remove_toctrees",  # Clean up AutoAPI navigation noise
     "sphinx_prompt",
-
     # Utility extensions
     "sphinxcontrib.bibtex",
-
     # Custom FABulous extensions
     "generate_cli_docs",
     "generate_configvar_docs",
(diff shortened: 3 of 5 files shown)


docs/source/_ext/generate_configvar_docs.py
```diff
@@ -0,0 +1,372 @@
+#!/usr/bin/env python3
+# Copyright 2024 Efabless Corporation
```
EverythingElseWasAlreadyTaken: If this is something from Efabless, we should rather include it, but it seems like you also edited it.

docs/source/_templates/autoapi/index.rst
```diff
@@ -1,13 +1,13 @@
...
+The API is organized hierarchically - click on a package to see its submodules and classes.
+
 .. toctree::
-   :maxdepth: 1
    :titlesonly:
-   :glob:
 
-   FABulous/index
-   FABulous/**/index
+   {% for page in pages|selectattr("is_top_level_object") %}
```
EverythingElseWasAlreadyTaken, line 11: Should we make it skip the first hierarchies, since currently it is like 
API Reference 
   FABulous
        ... modules

FABulous is not really needed I think.

docs/source/_templates/flow_variable.md.jinja
```diff
@@ -0,0 +1,32 @@
...
+
+## Environment Variables
+
+FABulous settings can be configured via environment variables with the `FAB_` prefix. These can be set in your shell or in a `.env` file in your project's `.FABulous` directory.
+{% for category, variables in settings_vars.items() %}
+
+### {{ category }}
+
+| Variable | Environment Variable | Type | Default | Description |
+|----------|---------------------|------|---------|-------------|
```
EverythingElseWasAlreadyTaken: This is really nice! 
But maybe we should also provide which values people can set, since currently it shows the internal data types, but users can only set strings, which can lead to confusion. 

Image

docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -0,0 +1,241 @@
+# Convert your design into GDSII format
+
+Once you have compiled your design into Verilog using the [building fabric](./building_fabric.md) guide, you can then convert your design into a GDS file for fabrication. The harden process is a 2 stage process. We will first harden all the tiles, and then stitching them together. We chose to do this instead of compile the whole fabric as a flat net list is because in an FPGA there are a lot of repeated components, which means synthesis will be repeatably synthesis similar logic, and the subsequent place and route will also be doing the same. To speed up the hardening process, we deploy this two stage strategy with two key benefits:
```
EverythingElseWasAlreadyTaken:
```suggestion
Once you have compiled your fabric RTL using the [building fabric](./building_fabric.md) guide, you can then convert your design into a GDS file for fabrication. The harden process is a 2 stage process. We will first harden all the tiles, and then stitching them together. We chose to do this instead of compile the whole fabric as a flat net list is because in an FPGA there are a lot of repeated components, which means synthesis will be repeatably synthesis similar logic, and the subsequent place and route will also be doing the same. To speed up the hardening process, we deploy this two stage strategy with two key benefits:
```

docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -0,0 +1,241 @@
...
+
+1. Make fabric development time much faster.
+2. Make scaling your fabric to larger design much easier.
+
+Per tile hardening is much faster as the size of the design is just much smaller. And since a fabric have a lot of repeated elements, optimizing one majority tile might potentially give you huge benefits it terms of, power, performance and area (PPA), which enables you to have a performant fabric in a shorter time window. Another benefit is this allows multiple tile development happening simultaneously, which allow different parties with different specialty to fully optimize a part of the design.
+
+Because we are reusing the existing each of the pre harden tile, this allows you to avoid re-harden all the tiles once again. To get a larger fabric you can simply "add more tiles" to the `fabric.csv` and you can have a much larger design. If you need to add new tiles for new functionality, you can only need to harden the newly created tile, without needing redoing the rest of the tile, which in turn, speed up your development process.
```
EverythingElseWasAlreadyTaken:
```suggestion
We are reusing existing pre hardened tiles, which avoids re-harden all the tiles on a new run.  To get a larger fabric you can simply "add more tiles" to the `fabric.csv`. If you need to add new tiles for new functionality, you can harden the newly created tile, without needing redoing the rest of the tiles.
```

docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -0,0 +1,241 @@
...
+
+To take advantage of fabric stitching, there are two limitations to the tile physical implementation. First is the interface of the adjacent tile need to line up in the exact order, and they need to align physically with the same spacing. Second is the size of the tile of the adjacent tile in the same row must have the same height and tile in the same width must be the same width in order to have perfect stitching. For the tile interface alignment, this is something that our framework will handle, however the tile sizing is something that need special handle. In the following will describe all the details of each stage.
```
EverythingElseWasAlreadyTaken:
```suggestion
To take advantage of fabric stitching, there are two limitations to the tile physical implementation. 

1. The interface of adjacent tiles need to line up in the exact order, and they need to align physically with the same spacing.
2. Adjacent tiles in the same row must have the same height and tiles in the same column must be the same width in order to have perfect stitching. For the tile interface alignment, this is something that our framework will handle, however the tile sizing is something that need special handling, which we will describe in the following for each stage.

```

docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Prerequisite
+
+### Install tools
+
+We use [librelane](https://github.com/librelane/librelane) as our main flow. To access the full flow, you will need to use the [Nix based installation method](../../getting_started/installation/nix-env.md). Which will provide you the full environment with all the required tools.
```
EverythingElseWasAlreadyTaken:
```suggestion
We use [librelane](https://github.com/librelane/librelane) as our main flow. To access the full flow, you should to use the [Nix based installation method](../../getting_started/installation/nix-env.md), which will provide the full environment with all the required tools.
```

docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -0,0 +1,241 @@
...
+
+To compile the design we will also need to install the PDK. In the Nix installation method we have also packaged in a PDK version manger [ciel](https://github.com/fossi-foundation/ciel). By default, we have set up the project to target the `ihp-sg13g2` process (130nm). To install the PDK run the following command:
+
+```bash
+ciel enable --pdk-family ihp-sg13g2 cb7daaa8901016cf7c5d272dfa322c41f024931f
+```
+
+## Changing PDK
+
+We support all PDK that is supported by librelane. As a result you can switch to targeting other process node such as the Sky130A and gf180mcu. To switch to those PDK, you will need to modify the `FAB_PDK_ROOT` and `FAB_PDK`, in `./<project>/.FABulous/.env` or set them in the shell as an environment variable. For `FAB_PDK` will be the PDK you are using, and the `FAB_PDK_ROOT` will be where the PDK is located. If you are installing the PDK from `ciel`, it will be located at `.ciel` under the user directory. An example of the `.env` file will be:
```
EverythingElseWasAlreadyTaken:
```suggestion
We support all PDKs that are supported by librelane. As a result you can switch to targeting other process node such as the Sky130A and gf180mcu. To switch to those PDKs, you will need to modify the `FAB_PDK_ROOT` and `FAB_PDK`, in `./<project>/.FABulous/.env` or set them in the shell as an environment variable. For `FAB_PDK` will be the PDK you are using, and the `FAB_PDK_ROOT` will be where the PDK is located. If you are installing the PDK from `ciel`, it will be located at `.ciel` under the user directory. An example of the `.env` file will be:
```

docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -0,0 +1,241 @@
...
+
+```bash
+fabulous> gen_all_tile_macros
+fabulous> gen_all_tile_macros --parallel      # Run in parallel for faster compilation
+fabulous> gen_all_tile_macros --optimise      # With optimization (balance mode)
+```
+
+### Tile Config
+
+You can change and customise any setting you want via modifying the `gds_config.yaml` file. There are two layers of configuration. There is a `gds_config.yaml` located at the `<project>/Tile/include` and in each of the tile, they have it respective `gds_config.yaml`. The one in the `include` is the base configuration which applies to all the tile, you can put all the settings that is common to all the tile in that file. For per tile specific configuration, you can set them using the `gds_config.yaml` at the tile.
```
EverythingElseWasAlreadyTaken:
```suggestion
You can change and customise any setting you want via modifying the `gds_config.yaml` file. There are two layers of configuration. There is a `gds_config.yaml` located at  `<project>/Tile/include` and in each of the tiles, they have it respective `gds_config.yaml`. The one in the `include` is the base configuration which applies to all  tiles, you can put all the settings that is common to all  tiles in that file. For per tile specific configuration, you can set them using the `gds_config.yaml` at the tile.
```

docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -0,0 +1,241 @@
...
+
+The per tile `gds_config.yaml` is particularly useful and important as you can set per tile `die_area`. In order for the tile to perfectly stitch together, as mention before all tile in the same row much have the same height, and tile in the same column must have the same width, and you can control the tile sizing by using it. For what variable can be configured please check the [flow variable table](/generated_doc/gds_variable.md)
```
EverythingElseWasAlreadyTaken: Rather use labels for references then files.

docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -0,0 +1,241 @@
...
+      - Co
+      reverse_result: false
+      sort_mode: bus_major
+    ...
+X0Y1:
+    ...
+```
+
+The entry key `X0Y0` represent the location of the pin of a tile. For normal tile, it will just have `X0Y0`, this is mainly useful for supertile and to set where the pin should locate relative to the tile shape, having this control allow us to also align supertile and normal tile pin automatically. Then the second layer of key is setting along which side of the tile should have the allocated pin. Each entry within a `side` is controlling a pin placement group. The `pins` is a list of entry of the pin that you want to set. The entry of the pin can be either a regular expression that matches actual pins or an integer which indicate a virtual pin. A virtual pin is basically a placeholder to space out the pins. The value of the integer represent the number of virtual pin to be added. In the group you can set the `min_distance` and `max_distance` and the placement script will try it best to fulfil the requirement, and yield an error in the event of the constraint cannot be achieved. The order of the pin layout will be in the following format:
+
```
EverythingElseWasAlreadyTaken, line 117: Is the example missing here?

docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -0,0 +1,241 @@
...
+# Bit Minor
+addr_bus[0]  # Index 0 first
+data_bus[0]  # Index 0 second (different bus)
+addr_bus[1]  # Index 1 first
+data_bus[1]  # Index 1 second (different bus)
+```
+
+## Stitching the tiles
+
+Once all the tile is compiled to GDS format with correct sizing, we then can stitch them together, and we can do this by using the following command:
```
EverythingElseWasAlreadyTaken:
```suggestion
Once all the tile is compiled to GDS format with correct sizing, we then can stitch them together. This can be done by using the following command:
```

docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -0,0 +1,241 @@
...
+
+```bash
+fabulous>gen_fabric_macro
+```
+
+And the full fabric will be stitched together.
+
+We have a custom top level IO placement script which will align all the pins with the IO pins around the perimeter. You will notice there is a small halo ring around the fabric as we will need some extra space to get the clock leader routed. Same as tile implementation, there is a `gds_config.yaml` file under the `Fabric` folder where you can set additional variables. Check the [flow variable table](/generated_doc/gds_variable.md) for available options.
```
EverythingElseWasAlreadyTaken: use labels for references

docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Full Automated Flow
+
+For a fully automated flow that handles tile size optimization and fabric stitching, use:
```
EverythingElseWasAlreadyTaken, line 164: are there any drawbacks with this ? maybe long runtime?

docs/source/user_guide/cli_doc/cli_feature.md
```diff
@@ -0,0 +1,217 @@
+# FABulous CLI — Features and Usage
+
+In FABulous, we provides two primary method to drive the tool. First is the interactive REPL mode, which is good for performing design space exploration. Second is the batch mode, which is good for CI and automation. This document summarizes the key features of both modes and provides guidance on when to use each.
```
EverythingElseWasAlreadyTaken:
```suggestion
In FABulous, we provides two primary methods to drive the tool. First is the interactive REPL mode, which is good for performing design space exploration. Second is the batch mode, which is good for CI and automation. This document summarizes the key features of both modes and provides guidance on when to use each.
```

docs/source/user_guide/cli_doc/cli_feature.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Interactive mode
+
+To start the REPL, simply run `FABulous start` and this will lunch the interactive shell in the current working directory.
```
EverythingElseWasAlreadyTaken:
```suggestion
To start the REPL, simply run `FABulous start`, which will lunch the interactive shell in the current working directory.
```

docs/source/user_guide/cli_doc/cli_feature.md
```diff
@@ -0,0 +1,217 @@
...
+
+```bash
+FABulous start
+
+# you will see a prompt like this:
+FABulous>
+```
+
+But some times you might want to start the REPL in for a sepecific project directory, and you can do this by providing the `--project-dir/-p` arguments to chooses which project directory to load.
```
EverythingElseWasAlreadyTaken:
```suggestion
If you want to start the REPL in for a sepecific project directory, and you can do this by providing the `--project-dir/-p` arguments to chooses which project directory to load.
```

docs/source/user_guide/cli_doc/cli_feature.md
```diff
@@ -0,0 +1,217 @@
...
+
+Once you are in the REPL there are many features that help you with the flow and we will list out a few useful ones below. For more details please refer to the [cmd2 documentation](https://cmd2.readthedocs.io/en/stable/) since FABulous CLI is built on top of cmd2.
```
EverythingElseWasAlreadyTaken:
```suggestion
Once you are in the REPL, you will find may features that help you with the flow. 
We will list out a few useful ones below. For more details please refer to the [cmd2 documentation](https://cmd2.readthedocs.io/en/stable/) since FABulous CLI is built on top of cmd2.
```

docs/source/user_guide/cli_doc/cli_feature.md
```diff
@@ -0,0 +1,217 @@
...
+
+For more details about the available commands, please refer to the [Interactive CLI Commands Reference](/generated_doc/interactive_cli_commands.md).
```
EverythingElseWasAlreadyTaken: Use lables for references.

docs/source/user_guide/cli_doc/cli_feature.md
```diff
@@ -0,0 +1,217 @@
...
+FABulous start
+
+# you will see a prompt like this:
+FABulous>
+```
+
+But some times you might want to start the REPL in for a sepecific project directory, and you can do this by providing the `--project-dir/-p` arguments to chooses which project directory to load.
+
+Once you are in the REPL there are many features that help you with the flow and we will list out a few useful ones below. For more details please refer to the [cmd2 documentation](https://cmd2.readthedocs.io/en/stable/) since FABulous CLI is built on top of cmd2.
+
```
EverythingElseWasAlreadyTaken, line 21: Maybe we should reference the help/? command here.

docs/source/user_guide/cli_doc/cli_feature.md
```diff
@@ -0,0 +1,217 @@
...
+### Tab completion
+
+Argument and filesystem completion are available — press Tab to complete filenames and subcommands.
+
+### Editor integration (`edit`)
+
+You can open files from the REPL in your preferred editor and return to the CLI once the editor closes. The CLI chooses the editor from the following (in order): the environment variable `FABULOUS_EDITOR` (recommended), then `EDITOR` / `VISUAL`.
+The main use of the `FABULOUS_EDITOR` is to provide per-project editor settings, but without it will try resolving with the `$EDITOR` which is set for most terminal environments.
+
```
EverythingElseWasAlreadyTaken, line 48: Maybe we should als referr to the env var section here.

docs/source/_templates/flow_variable.md.jinja
EverythingElseWasAlreadyTaken: How about this one ? Can we change this? 

Another thought I just had, that we don't mention the global .env file in the ~/.fabulous folder.

KelvinChung2000: I have updated it 

image

And this is how it looks. The description field is where we can add documentation about the variable. We could move to using that to explain what the expected value should be. But in this case, I will remove the User Value entry

docs/source/_templates/autoapi/index.rst
KelvinChung2000: This is updated. You can check the API page to see if this is what you're looking for.

docs/source/_templates/flow_variable.md.jinja
EverythingElseWasAlreadyTaken: Yes it's there. The horizontal scrolling is not really nice and can be overlooked quickly, but it is how it is with all those themes. 

What about the global .env? 
This should be documented.

KelvinChung2000: I have updated. Later, we can use `field(title="...", description= "..." )` to replace the current value setting, and all the docs will be generated and automatically separated into multiple tables.

PR #593, opened 2026-02-14
fix: fix LVS error

KelvinChung2000:
This fixes the LVS error due to misconfiguration. Diode on ports is added to ensure the eFPGA macro is antenna error-free.


EverythingElseWasAlreadyTaken (approved):
Overall looks OK to me. 

One thing I don't fully understand is, why the tiles are so big? 
Leo got something like 210x210 for the LUT4AB, we have now 245x245.
Also the IO tiles, they are huge compared to Leos tiles.

PR #598, opened 2026-02-17
chore: project upgrade

KelvinChung2000:
Renamed FP_IO_VLAYER/FP_IO_HLAYER to IO_PIN_V_LAYER/IO_PIN_H_LAYER.

Remove DRT_MAX_LAYER as RT_MAX_LAYER control the variable

Update the IHP PDK hash in the documentation from cb7daaa8 to
c4b8b4e5, which has DRC scripts compatible with the new librelane
parameter names.

Fixing nix


docs/source/user_guide/building_doc/fabric_gds.md
```diff
@@ -31,7 +31,7 @@ As of writing, we are using custom build of librelane, as a result, the upstream
 To compile the design, we will also need to install the PDK. In the Nix installation method we have also packaged in a PDK version manager [ciel](https://github.com/fossi-foundation/ciel). By default, we have set up the project to target the `ihp-sg13g2` process (130nm). To install the PDK run the following command:
 
 ```bash
-ciel enable --pdk-family ihp-sg13g2 cb7daaa8901016cf7c5d272dfa322c41f024931f
+ciel enable --pdk-family ihp-sg13g2 c4b8b4e5e7a05f375cca3815d51b3a37721fbf5c
```
EverythingElseWasAlreadyTaken, line 34: We should maybe also add the PDK hash in the project .env.
Since for the rest, we have the nix .lock files.

EverythingElseWasAlreadyTaken (approved):

docs/source/user_guide/building_doc/fabric_gds.md
KelvinChung2000: We could actually also add this to the nix and make it part of the shell env variable. Then we only need to update the hash in the flake without updating the doc. So the installation instructions become `ciel enable --pdk-family ihp-sg13g2 $ENV_PDK_HASH` etc

KelvinChung2000: Do you want to do this in this PR or do we do a separate one?

EverythingElseWasAlreadyTaken: Good Idea, it is better to have  for the default PDK in the nix .lock, since it makes updating easier.
 
But this does not necessarily mean, that people use this pdk, so I think it would be still good to leave a note on the PDK version in the project.

EverythingElseWasAlreadyTaken: This can also be a separate PR

PR #603, opened 2026-02-19
fix: validate tile name matches folder name and guard empty tile list

EverythingElseWasAlreadyTaken:
parseTilesCSV now warns when the parsed tile name does not match the parent directory name.

FABulous_CLI raises an error when no intersecting tiles found on disk with those defined in fabric.csv, since this crashed previously with a unrelated message.

### DIFF
--- a/fabulous/fabulous_cli/fabulous_cli.py
+++ b/fabulous/fabulous_cli/fabulous_cli.py
@@ -460,6 +460,13 @@ class FABulous_CLI(Cmd):
         superTileByFabric = list(self.fabulousAPI.fabric.superTileDic.keys())
         self.allTile = list(set(tileByPath) & set(tileByFabric + superTileByFabric))
 
+        if not self.allTile:
+            logger.error(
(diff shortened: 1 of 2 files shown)


fabulous/fabulous_cli/fabulous_cli.py
```diff
@@ -460,6 +460,13 @@ def do_load_fabric(self, args: argparse.Namespace) -> None:
         superTileByFabric = list(self.fabulousAPI.fabric.superTileDic.keys())
         self.allTile = list(set(tileByPath) & set(tileByFabric + superTileByFabric))
 
+        if not self.allTile:
```
KelvinChung2000, line 463: This is fine, but I'm just curious what actually led to needing this test, as this means that all the tiles in the directory do not intersect with the tile defined in the fabric.csv.

EverythingElseWasAlreadyTaken: Was during my thesis, when I generated tile configs and accidentally made the tile names slightly different to the tile folder names.

PR #609, opened 2026-03-02
fix(tests): isolate FAB_USER_CONFIG_DIR from real ~/.fabulous/.env

EverythingElseWasAlreadyTaken:
FAB_USER_CONFIG_DIR is computed at module import time, so the Path.home() monkeypatch arrived too late to redirect it. Patch the constant directly in both modules that bind it and add reset_context() teardown to the autouse fixture so every test cleans up. This way, the local .env file gets not overwritten when running pytest locally.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,6 +11,8 @@ from loguru import logger
 from fabulous.fabulous_cli.fabulous_cli import FABulous_CLI
 from fabulous.fabulous_cli.helper import create_project, setup_logger
 from fabulous.fabulous_settings import init_context, reset_context
+import fabulous.fabulous_settings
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -11,6 +11,8 @@
 from fabulous.fabulous_cli.fabulous_cli import FABulous_CLI
 from fabulous.fabulous_cli.helper import create_project, setup_logger
 from fabulous.fabulous_settings import init_context, reset_context
+import fabulous.fabulous_settings
+import fabulous.fabulous
```
KelvinChung2000, line 15: We might want to re-export fabulous later in `__init__.py`.

PR #617, opened 2026-03-05
fix: fix when ciel is never used before

KelvinChung2000:
With the full auto ciel PDK resolving feature implemented, we no longer need to resolve the PDK root at project creation. 

Ciel always resolves to `~/.ciel`, but on a system where ciel is never used (e.g., a Docker container), this will raise a warning about a missing `~/.ciel` and leave `FAB_PDK` unset, which the auto-resolving logic cannot resolve.

### DIFF
--- a/docs/source/user_guide/cli_doc/fabulous_variable.md
+++ b/docs/source/user_guide/cli_doc/fabulous_variable.md
@@ -1,4 +1,5 @@
 (fabulous-variables)=
+
 # FABulous Configuration Variables
 
 FABulous can use environment variables to configure options, paths and projects. We distinguish between two types of environment variables: **global** and **project specific** environment variables.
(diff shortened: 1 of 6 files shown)


docs/source/user_guide/cli_doc/fabulous_variable.md
```diff
@@ -21,5 +22,30 @@ All environment variables can be set in the shell before running FABulous or can
 Environment variables set in the **shell** always have the **highest priority**, followed by project-specific `.env` files, then global `.env` files.
 :::
 
+## PDK Resolution Logic
```
EverythingElseWasAlreadyTaken, line 26: Thanks for also adding docs for the PDK resolution!
Would be nice if you could add a reference for the GDS flow docs to this here.

EverythingElseWasAlreadyTaken (approved):
Just a small comment on the docs, that would be nice if changed. 
Other than that can be merged.

EverythingElseWasAlreadyTaken (approved):
Looks good now! 

Just needs a rebase and can be merged then.

EverythingElseWasAlreadyTaken (approved):

PR #621, opened 2026-03-07
chore(doc): fix and enhance doc page

KelvinChung2000:
Instead of writing my thesis, making Docs look nicer is, of course, is more important...

Updated the autoapi doc generation to make the docs look nicer and fixes #620 . 

Updated the banding on the top left, which borders me for a very long time. 

I have made the right sidebar toc tree collapsible, since not everyone likes that sidebar. I also make page content render according to available screen space. 

close #620


docs/source/_static/toc_sidebar.js
```diff
@@ -0,0 +1,85 @@
+document.addEventListener("DOMContentLoaded", () => {
```
EverythingElseWasAlreadyTaken, line 15: A few comments whatever all this is doing would be quite helpful, especially if any issues come up in the future.

ruff.toml
```diff
@@ -26,7 +26,7 @@ exclude = [
   "node_modules",
   "site-packages",
   "venv",
-  "conf.py",
+  "docs",
```
EverythingElseWasAlreadyTaken, line 29: Why exclude the docs here?

EverythingElseWasAlreadyTaken (commented):
Besides the comments, why do we actually need that much custom formatting for the autoapi? 
Isn't it the idea of autoapi that we don't need to do all this by ourselves?

ruff.toml
KelvinChung2000: Since I think linting on them is not that important. I could go with `conf.py` + `conf_helper.py`, or we could fix the linting

EverythingElseWasAlreadyTaken: What's the issue with the linting here? 

I mean its code, and it needs to be maintained at some point. 
This is why we lint our stuff and if the linter throwing false positives or so, we can deactivate for this lines / files.

EverythingElseWasAlreadyTaken (approved):

PR #622, opened 2026-03-09
chore: update to pin ghdl v6.0.0

KelvinChung2000:
Minor cleanup, so Linux ghdl is also installed via tarball.


IAmMarcelJung (approved):
LGTM, also tried it locally.

PR #625, opened 2026-03-10
chore: nix pre-commit

KelvinChung2000:
close #624 

This basically just sets up the pre-commit. 

The only issue is that, to install the pre-commit hook, you will need to have Haskell installed if you are not in the Nix shell. 

Will be a draft for now until #622 is merged.

### DIFF
--- a/nix/overlay/python.nix
+++ b/nix/overlay/python.nix
@@ -1,39 +1,49 @@
 final: prev: {
-    # Override fasm to use GitHub source instead of PyPI and add missing build deps
-    fasm = prev.fasm.overrideAttrs (old: {
-        src = final.pkgs.fetchFromGitHub {
-        owner = "chipsalliance";
-        repo = "fasm";
-        rev = "v0.0.2";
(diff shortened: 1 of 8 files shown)


nix/overlay/python.nix
```diff
@@ -1,39 +1,49 @@
...
-        repo = "fasm";
-        rev = "v0.0.2";
-        sha256 = "sha256-AMG4+qMk2+40GllhE8UShagN/jxSVN+RNtJCW3vFLBU=";
-        };
-        nativeBuildInputs = (old.nativeBuildInputs or []) ++ final.resolveBuildSystem {
-        setuptools = [ ]; wheel = [ ]; cython = [ ];
-        };
-        propagatedBuildInputs = (old.propagatedBuildInputs or []) ++ [ prev.textx ];
-    });
+  # Override fasm to use GitHub source instead of PyPI and add missing build deps
```
EverythingElseWasAlreadyTaken, line 2: We could also think about making the FASM package our own and throw it on pypi. 
Could simplify things in the long run.

EverythingElseWasAlreadyTaken (approved):

PR #659, opened 2026-03-21
chore: Validate frame address register capacity at fabric creation

### LINKED ISSUE #490: Number of frames never cross checked with the available bits for frame adress
The number of frames in a fabric is limited by the number of bits available for the frame adresses. Creating a fabric that exceeds this frame limit however does not generate any feedback or error in the flow. A user is able to generate a bitstream.bin file. This file of course does not work, frame adresses above the limit are nonsensical. But this can only be found out by inspecting the binary bitstream file.

Checking for adress overflow during bitstream generation might be a good place to indicate to a user that the written bitstream is not valid, and changes to the fabric are required.

KelvinChung2000:
Add checks in Fabric.__post_init__() to catch frame address overflow early

Closes #490

### DIFF
--- a/fabulous/fabric_definition/fabric.py
+++ b/fabulous/fabric_definition/fabric.py
@@ -118,6 +118,65 @@ class Fabric:
         The wires are used during model generation to work with wire that going cross
         tile.
         """
+        # The FrameAddressRegister (frameBitsPerRow bits wide)
+        # is partitioned into three non-overlapping regions:
+        # upper bits for column select, a single desync bit,
+        # and lower bits for one-hot frame strobe.
+        # See docs/fabric_definition.md for the bit layout.
+        if self.desync_flag >= self.frameBitsPerRow:
+            raise ValueError(
+                f"desync_flag bit position "
+                f"({self.desync_flag}) exceeds "
+                f"FrameAddressRegister width "
+                f"({self.frameBitsPerRow}). "
+                f"Reduce desync_flag or increase "
+                f"frameBitsPerRow."
+            )
+
+        col_select_lo = self.frameBitsPerRow - self.frameSelectWidth
+        if self.maxFramesPerCol > col_select_lo:
+            raise ValueError(
+                f"FrameAddressRegister overflow: "
+                f"frame strobe bits [0:{self.maxFramesPerCol - 1}]"
+                f" overlap with column select bits "
+                f"[{col_select_lo}:"
+                f"{self.frameBitsPerRow - 1}]. "
+                f"Reduce maxFramesPerCol or increase "
+                f"frameBitsPerRow."
+            )
+
+        if self.desync_flag < self.maxFramesPerCol:
+            raise ValueError(
+                f"desync_flag bit position "
+                f"({self.desync_flag}) falls within the "
+                f"frame strobe region "
+                f"[0:{self.maxFramesPerCol - 1}]. "
+                f"Increase desync_flag or reduce "
+                f"maxFramesPerCol."
+            )
+
+        if self.desync_flag >= col_select_lo:
+            raise ValueError(
+                f"desync_flag bit position "
+                f"({self.desync_flag}) falls within the "
+                f"column select region "
+                f"[{col_select_lo}:"
+                f"{self.frameBitsPerRow - 1}]. "
+                f"Adjust desync_flag, frameSelectWidth, "
+                f"or frameBitsPerRow."
+            )
+
+        max_cols = 1 << self.frameSelectWidth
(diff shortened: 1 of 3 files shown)


fabulous/fabric_definition/fabric.py
```diff
@@ -118,6 +118,65 @@ def __post_init__(self) -> None:
...
+        # See docs/fabric_definition.md for the bit layout.
+        if self.desync_flag >= self.frameBitsPerRow:
+            raise ValueError(
+                f"desync_flag bit position "
+                f"({self.desync_flag}) exceeds "
+                f"FrameAddressRegister width "
+                f"({self.frameBitsPerRow}). "
+                f"Reduce desync_flag or increase "
+                f"frameBitsPerRow."
+            )
```
IAmMarcelJung: For now shouldn't we check just if the desync flag is at position 20 directly, instead of splitting it into two checks? Also the error message also might be misleading, since we currently assume the desync flag to be at this position in different places in the flow and I think users can't easily change it without taking good care of every occurrence.

fabulous/fabric_definition/fabric.py
```diff
@@ -118,6 +118,65 @@ def __post_init__(self) -> None:
...
+        if self.desync_flag >= col_select_lo:
+            raise ValueError(
+                f"desync_flag bit position "
+                f"({self.desync_flag}) falls within the "
+                f"column select region "
+                f"[{col_select_lo}:"
+                f"{self.frameBitsPerRow - 1}]. "
+                f"Adjust desync_flag, frameSelectWidth, "
+                f"or frameBitsPerRow."
+            )
```
IAmMarcelJung, line 141: See above

IAmMarcelJung (commented):
Having these checks is great, but I'm not sure if these should currently be dynamic and not just hardcoded, since the header format is currently quite inflexible and I think we are making assumption about the header format throughout the flow. Shouldn't we just check the exact bit regions for now?

fabulous/fabric_definition/fabric.py
KelvinChung2000: I would assume this is the intended design, but somehow things are hardcoded to work with only a single case for now. I think the only part that makes the assumption is at the FSM, only?

IAmMarcelJung: I'm not 100% sure about all locations, but e.g. in bit-gen we also assume the 20th bit currently.

KelvinChung2000: Maybe the right move is to update the bit-gen to read the desync value from the bitstream spec, and change it there. Then we avoid the hard-coding.  

But now I am thinking about this, why do we need to set the desync flag in the first place? Since the number of rows and columns are set anyway. We can always keep the following structure

```
onehot/enum                 onehot/enum
[column select] | [desyn] | [frame strobe select]
``` 

Then we only need to verify if we have enough column select width or strobe select width, and the desyn flag is always between them.

IAmMarcelJung: Sure we can do this, but we just have to make sure that this is completely transparent to the user since otherwise this would cause confusion. And we also wanted to rework the whole config logic anyway (although this could also still take some time), so that's why I would tend to not change anything there for now.

KelvinChung2000: The change will not happen soon. But what would be the best check to put in place for now? 

At most 32 columns
At most 32 rows (row_select is a hard-coded parameter)
At most 20 frames per tile (In theory, we can do 27 since the address mask is 32bit - 5bit(col mask) = 27bit, but the one-hot frame index is hard coded to 20bit)
At most 26 bels per tile (hard-coded limit by the number of capital alphabet available)

Maybe put the hard check on these?

IAmMarcelJung: I think for now checking for these hardcoded limits would be the way to go. Better have something that is working than changing this now and creating a new potential source of errors. But I'm also over careful on this because I had so many issues with different versions and so on when working with the boards ^^

EverythingElseWasAlreadyTaken: I also think the hardcoded checks are fine for now. 
Changing the Config Logic will be part of FABulous 3.0. 
So we can also add a hard check for the desync flag at 20, since the normal user should not change this for now.

I started working on a refactor of bit_gen, I will foresee, that the values for desync etc. can be read from the bistream spec and have a hardcoded fallback for legacy support.

EverythingElseWasAlreadyTaken (approved):

IAmMarcelJung (approved):
LGTM, thanks for changing it!

PR #661, opened 2026-03-24
fix: fix wrong parse csv case

KelvinChung2000:
close #660 

We should write some tests for this function.


IAmMarcelJung (approved):
Thanks, LGTM. I guess the `_matrix` was just a leftover from some old version then, although I could not trace back the change further than some refactor.

PR #664, opened 2026-03-25
fix: fix deprecated name lead to fail variable propagate

KelvinChung2000:
fixes #662


IAmMarcelJung (approved):
LGTM, thanks for the fix!

PR #668, opened 2026-03-29
fix: fix tile_io_place script

KelvinChung2000:


KelvinChung2000:
If you try this locally right now, it will fail because the tool version is not right. After #672  this should be working.

EverythingElseWasAlreadyTaken (approved):
LGTM

PR #670, opened 2026-03-30
fix: fix PDK path resolution error

KelvinChung2000:
I think this should solve https://github.com/FPGA-Research/FABulous/issues/666#issuecomment-4148779211, but it is better to test it on a different machine.


KelvinChung2000:
This will not run through and error until #672  is merged

IAmMarcelJung (approved):
Thanks, this fixed the error!

PR #672, opened 2026-03-31
chore: project upgrade

KelvinChung2000:
Once https://github.com/FPGA-Research/FABulous-fasm/pull/1 is done, I will update all the locks again and the gds-flow should be working again.


IAmMarcelJung (approved):
LGTM, if there is not dependency with the other PR, we can merge this.

PR #667, opened 2026-03-29
chore: add CI for gds-flow

KelvinChung2000:

PR #681, opened 2026-04-08
feat: improve variable loading

KelvinChung2000:


fabulous/fabric_generator/gds_generator/helper.py
```diff
@@ -75,6 +77,62 @@ def round_up_decimal(value: Decimal, pitch: Decimal) -> Decimal:
...
+) -> dict[str, Any]:
+    """Merge two mappings with one-level nested dict merging semantics."""
+    merged_config = dict(base_config)
+    for key, override_value in override_config.items():
+        base_value = merged_config.get(key)
+        if isinstance(base_value, dict) and isinstance(override_value, dict):
+            merged_config[key] = {**base_value, **override_value}
+        else:
+            merged_config[key] = override_value
+    return merged_config
```
IAmMarcelJung, line 92: Maybe add an info or even warning if a value gets overridden?

KelvinChung2000: If we need to issue a warning, we may do it at the call site, since this is a generic helper for merging.

IAmMarcelJung: Ah okay, I see. It's not mandatory but would be a nice convenience feature if it's not much overhead to implement.

KelvinChung2000:
I will do this in #667  since the gf180 also needs this overriding, and with https://github.com/librelane/librelane/issues/926, this is basically not necessary anymore.

PR #690, opened 2026-04-17
fix: allow nix develop from a different directroy

IAmMarcelJung:
Currently it is not possible to just run `nix develop /path/to/fabulous/` to get a nix shell with our environment in which ever directory you want, the `REPO_ROOT` is always evaluated using `git rev-parse`. Since `flake.nix` is in the root of the repo anyway, this can be change to just `"${toString ./.}"`.

At least this is what Claude suggested, so please let me know if this might break anything, although it makes sense to me and _it works on my machine_ ...

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -221,7 +221,7 @@
             ];
             devshell.startup.fabulous-setup = {
               text = ''
-                export REPO_ROOT=$(git rev-parse --show-toplevel)
+                export REPO_ROOT="${toString ./.}"
                 ORIGINAL_PS1="$PS1"
 
                 . ${virtualenv}/bin/activate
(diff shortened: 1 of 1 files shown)


KelvinChung2000:
By the way, you should now be able to do `fabulous nix-env` anywhere with fabulous installed, and it will drop you into the shell.

IAmMarcelJung:
Ah you mean if it is installed using uv? Okay that's also an option then. But I think it would still be nice if it works with the flake.nix file. Btw we could also allow running it directly from the GitHub path, but this would then ofc not work for the `create_project` command (probably this here too).

KelvinChung2000:
This will break the `fabulous nix-env` command. With the original implementation, if you run `fabulous nix-env` in the fabulous root, it will make the edits take effect immediately. With the change, the `toString ./.` will reference `fabulous_nix/`, not `.` , and any edit to the project will not reflect.

PR #695, opened 2026-04-23
feat: export as librelane plugin

KelvinChung2000:


fabulous/fabric_definition/define.py
```diff
@@ -25,25 +26,36 @@ class IO(Enum):
...
 
     NORTH = "NORTH"
-    SOUTH = "SOUTH"
     EAST = "EAST"
+    SOUTH = "SOUTH"
     WEST = "WEST"
     JUMP = "JUMP"
     SJUMP = "SJUMP"
 
+    def __lt__(self, other: "Direction") -> bool:
```
EverythingElseWasAlreadyTaken, line 52: AFAIK. to make the class sortable, you should also implement` __eq__` to make proper use of totalordering.

fabulous/fabric_generator/gds_generator/flows/plugin_tile_flow.py
```diff
@@ -0,0 +1,327 @@
...
+    with tile_csv.open("r", encoding="utf-8") as f:
+        f.readline()
+        for raw in f:
+            line = raw.strip()
+            if "EndSuperTILE" in line:
+                break
+            for cell in line.split(","):
+                name = cell.strip()
+                if name:
+                    subtile_names.append(name)
```
EverythingElseWasAlreadyTaken: I think we have similar logic somewhere.
But anyway, it could be actually useful to make this whole function a helper function, that can be reused through the fabulous API.

fabulous/fabric_generator/gen_fabric/gen_fabric.py
```diff
@@ -276,97 +276,91 @@ def generateFabric(writer: CodeGenerator, fabric: Fabric) -> None:
...
-                            i.name
-                            for i in fabric.tile[y + j][x + i].getWestPorts(IO.INPUT)
-                        ]
-
-                    westInput = [
-                        f"Tile_X{x + i + 1}Y{y + j}_{p.name}"
-                        for p in fabric.tile[y + j][x + i + 1].getWestPorts(IO.OUTPUT)
-                    ]
-                    portsPairs += list(zip(westPorts, westInput, strict=False))
+                        portsPairs += [(p, "") for p in westPorts]
```
EverythingElseWasAlreadyTaken: The whole code is basically copied for each tile side, so it would make sense to make this somehow a function. But anyhow, the entire fabric gen code could need a refactor at some point, so we can postpone it.

tests/assets/librelane_plugin/README.md
EverythingElseWasAlreadyTaken, line 1: The tile library is already moved from leos space to FPGA-Resereach this needs an update here.

EverythingElseWasAlreadyTaken, line 1: Also with the assets, I'm not really shure if we should keep this here in the FABulous repo, since it bloats the repo more. 

We could just pull it in on testing, like we do it with the reference tests.

build_hooks.py
```diff
@@ -5,11 +5,114 @@
...
+def _render_librelane_plugin_init() -> str:
+    """Render the librelane_plugin_fabulous/__init__.py content."""
+    subpkgs = ", ".join(repr(s) for s in _LIBRELANE_PLUGIN_SUBPKGS)
+    return f'''\
+"""LibreLane plugin: re-export of the FABulous GDS steps and flows.
+
+Ships as a side-package inside the FABulous-FPGA wheel. LibreLane
+auto-discovers packages whose name matches ``librelane_plugin_*`` — importing
+this package walks every submodule under
+``{_UPSTREAM_PKG}.{{steps,flows}}`` so their
```
EverythingElseWasAlreadyTaken: Remove double quotes

EverythingElseWasAlreadyTaken: and maybe replace the em dashes.

build_hooks.py
```diff
@@ -5,11 +5,114 @@
...
+
+``FABulousTile`` and ``FABulousFabric`` are additionally re-exported at the
+package root as a drop-in replacement for ``mole99/librelane_plugin_fabulous``.
+
+Generated at build time by ``build_hooks.BuildPyWithFabulousNix`` — do not
+edit by hand. The source of truth for every Step and Flow lives in
+``{_UPSTREAM_PKG}``.
+"""
+
+import importlib as _importlib
```
EverythingElseWasAlreadyTaken, line 32: move imports to top

tests/gds_flow_test/flow_test/test_plugin_fabric_flow.py
```diff
@@ -0,0 +1,426 @@
...
+        mutable = GenericDict({"FABULOUS_FABRIC_CONFIG": resolved_list})
+        _, value = var.compile(mutable, warning_list_ref=[])
+
+        # Concrete fabric path must be recoverable from whatever type we accept.
+        fabric_path = Path(value[0]) if isinstance(value, list) else Path(value)
+        assert fabric_path == fabric_csv
+
+
+class TestBuildMacros:
+    """`_build_macros` turns a tile-name → macro-dir mapping into Macros."""
```
EverythingElseWasAlreadyTaken: remove non ascii chars

EverythingElseWasAlreadyTaken (commented):
Just a few comments on this.
Also I think we should add some docs how to use the plugin.

KelvinChung2000:
The docs of the plugin will be basically the same as Leo's librelane plugin docs. @mole99 Can I just copy them over?

fabulous/fabric_definition/define.py
KelvinChung2000: enum provides an `__eq__` that is correct for enum classes, and doesn't need a custom implementation.

build_hooks.py
KelvinChung2000: This is a script template not a real import

tests/assets/librelane_plugin/README.md
KelvinChung2000: We will have test assets in some way. I have thought that we could fetch them during testing, but cloning the whole tile repo for a single case is kind of stupid. We also have not fully confirmed how we want libraries to move forward anyway, so the structure might break as well

EverythingElseWasAlreadyTaken: Ok, let's leave it for now.
Maybe when we integrate the libraries, we'll find a proper solution.

fabulous/fabric_generator/gds_generator/flows/plugin_tile_flow.py
KelvinChung2000: I am trying to find a place to improve reuse. This is specifically for generating a tile without needing fabric csv.   I don't know whether this decoupling is a good thing, but this is how the tile lib working now

KelvinChung2000: I have moved this to the parser side now, which we can reuse later, depends on how things moving

EverythingElseWasAlreadyTaken (approved):

EverythingElseWasAlreadyTaken (approved):

mole99:
@KelvinChung2000 Are you referring to these docs? https://github.com/mole99/librelane_plugin_fabulous/tree/main/docs
If so, feel free to use whatever you find useful :)

Does the plugin already produce a compatible layout for existing fabrics? For example, https://github.com/mole99/tt-fabulous-ihp-26a
Of course, the pin placement can change, which is fine since all tiles are recompiled.

If so, I'd like to give this a try soon!
Should it be possible to simply specify the flake here and enable the Nix shell? https://github.com/mole99/tt-fabulous-ihp-26a/blob/main/flake.nix#L12

KelvinChung2000:
Last time I checked, it should be just a drop-in replacement with no modification. I am not sure how Nix will handle a multi-package repo. If it is basically doing a `pip install .`, I assume you can just point it to the commit.  But if possible, it would be better if you could also use our Nix flake, since the nix/tool version is also causing issues during integration when I am working on this. 

So once this is merged, in theory you can do 
```
pip install .
fabulous nix-env
librelane <config.yml>
```

And everything will work.

EverythingElseWasAlreadyTaken:
Not sure why the merge queue checks run into this error with our pytest tests, maybe a rebase helps.

mole99:
@KelvinChung2000 I do want to use the FABulous Nix flake, but like I said, as input to the flake in fabulous-tiles.

Enabling the Nix shell via `fabulous nix-env` kind of defeats the purpose of Nix, as I can't easily override the version of LibreLane (if needed) and cannot add additional tools to the flake.

I'm trying to run the following now as a plug-in replacement for [this line](https://github.com/mole99/tt-fabulous-ihp-26a/blob/main/flake.nix#L12):

```
  inputs = {
    librelane_plugin_fabulous.url = "github:KelvinChung2000/FABulous/feat/as-librelane-plugin";
  };
```

And the Nix shell has been successfully enabled. However, LibreLane does not list the plugin:

```
$ librelane --version
LibreLane v3.0.4

Copyright ©2025-2026 LibreLane Contributors

Adapted from OpenLane 2.0
Copyright ©2020-2025 Efabless Corporation

Available under the Apache License, version 2. Included with the source code,
but you can also get a copy at https://www.apache.org/licenses/LICENSE-2.0

Included tools and utilities may be distributed under stricter licenses.
```

KelvinChung2000:
You can. You just modify our flake. What the `fabulous nix-env` does is run `nix develop` using the repo's Nix flake. We are shipping the flake as part of the package `fabulous_nix`. The plugin is packaged and shipped as the `librelan_plugin_fabulous` package. This is why I say I am not sure how a Python package is handled on your side. Like, how is a multi-package repo being handled?

mole99:
What do you mean by "modify our flake"? Do you mean cloning the repository and changing the file?
One should be able to take the Flake as an input and override some of its inputs if required.

I am not yet sure how this Python package should be handled. This is my first time trying to integrate it like this.

KelvinChung2000:
I would prefer we all use the same environment. Cloning and having feature branches update our environment as you develop your own fabric would be ideal. Since we can keep track of the latest features/software needed to build a fabric, we can always update to meet the latest requirements. 

For all the fabric repos, I would like instructions that lean toward installing fabulous (at a specific version or commit, if wanted) and building the fabric from that point. 

But I think it should be possible to create a flake that installs both packages to the Python path. It's just what needs updating in the flake.

mole99:
If the environment is properly managed by Nix, then we all use the same environment. That's one of the key features of Nix. However, being able to override certain inputs is essential: even with my current plugin, I sometimes had to override LibreLane and apply certain patches to get a tapeout done. This is just standard functionality of Nix Flakes.

> For all the fabric repos, I would like instructions that lean toward installing fabulous (at a specific version or commit, if wanted) and building the fabric from that point.

That's what a Flake does. It pins the inputs at a specific commit or version.
Take this repo for example: https://github.com/mole99/tt-fabulous-ihp-26a
You can run `nix-shell` and build the fabric.

> But I think it should be possible to create a flake that installs both packages to the Python path. It's just what needs updating in the flake.

I finally understand what's the issue here: there is no actual FABulous package in this repo.

Currently, the Flake is (ab)used to setup the FABulous Python package as part of the shell hook, if I understand correctly.
This is done here: https://github.com/KelvinChung2000/FABulous/blob/aaec792bb707c89a2a3c1038a090ef139aa224af/flake.nix#L303

If there were a package definition for FABulous itself, then the Flake could be simplified considerably, and it could replace my existing LibreLane Nix plugin.
For reference, here is the FABulous package from my `librelane_plugin_fabulous`: https://github.com/mole99/librelane_plugin_fabulous/blob/main/nix/fabulous-fpga.nix
It is not complete since I only require the actual generation part of FABulous, but it can be used as a starting point for a complete package definition.

EverythingElseWasAlreadyTaken (approved):

PR #696, opened 2026-04-25
feat: multilayer pdn

mtetrault:
A few updates for the pdn section:

- fix the multilayer pdn for tiles to contact on stripes
- Fabric macro: automatic metal layer selection (supports multiplayer pdn)
- Fabric macro: changed hardcoded power nets for VDD_NETS/GND_NETS 
- Fabric macro: Initial support for multiple power domains, but no test case available here


KelvinChung2000:
Rather than taking it from the environment, supplying it through the script will be better.

https://github.com/FPGA-Research/FABulous/pull/667/changes#diff-2348359ea181ccf985faae07a6d85d45c8b8af3e0bef8d868fc2d63467bb03e7R26-R57

KelvinChung2000:
To run the formatting locally, you can use `pre-commit run --all-files`, which runs all our formatting and linting checks.

fabulous/fabric_generator/gds_generator/script/odb_power.py
```diff
@@ -20,99 +20,95 @@
...
+    type=str,
+    multiple=True,
+    help="The name(s) of the ground port(s). Repeat the option for multiple ports.",
 )
 @click.command()
 @click_odb
 def power(
     reader: Any,  # noqa: ANN401
-    metal_layer_name: str,
+    power_names: Tuple[str],
```
KelvinChung2000:
```suggestion
    power_names: tuple[str],
```

fabulous/fabric_generator/gds_generator/script/odb_power.py
```diff
@@ -20,99 +20,95 @@
...
+    ground_names: Tuple[str],
```
KelvinChung2000:
```suggestion
    ground_names: tuple[str],
```

fabulous/fabric_generator/gds_generator/script/odb_power.py
```diff
@@ -11,7 +11,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-from typing import Any
+from typing import Any, Tuple
```
KelvinChung2000:
```suggestion
from typing import Any
```

fabulous/fabric_generator/gds_generator/steps/odb_connect_pdn.py
```diff
@@ -25,7 +27,29 @@ def get_script_path(self) -> str:
...
+            gnd_pins.append("VGND")
+        else:
+            for power_net in self.config["GND_NETS"]:
+                gnd_pins.append("--ground-names")
+                gnd_pins.append(power_net)
+
+        #[
+        #    "--metal-layer-name",
+        #    self.config["RT_MAX_LAYER"],
+        #] +
```
KelvinChung2000:
```suggestion
```

fabulous/fabric_generator/gds_generator/script/odb_power.py
```diff
@@ -20,99 +20,95 @@
...
+            iterm_sigtype = iterm.getMTerm().getSigType()
 
-            if iterm_name == "VPWR":
-                info("Connecting VPWR")
-                iterm.connect(vpwr_net)
-
-            if iterm_name == "VGND":
-                info("Connecting VGND")
-                iterm.connect(vgnd_net)
+            if iterm_name == supply_name:# and iterm_sigtype == POWER:
```
KelvinChung2000:
```suggestion
            if iterm_name == supply_name:
```

KelvinChung2000 (changes requested):
Just some minor corrections. Since @mole99 has a better understanding of this, I will get him to take a look.

mtetrault:
Thanks for the pre-commit tip, I'll keep it in mind. 

The difference in this use case between "Tuple" (from typing lib) and "tuple" (basic keyword) in Python is not something I have a clear understanding, so I can go with which ever you prefer.

KelvinChung2000:
As far as I understand, `Tuple` from typing is basically used before Python had proper typing support. You can use them interchangeably, but `tuple` is the preferred one for newer versions of Python.

mtetrault:
I've worked out the linter issues; it took me a few tries, but I'm learning on the github tools and processes. Thanks for the ointers.

I'll wait for @mole99 's review before going through the pytest CI.

mole99 (approved):
Looks good to me!

I tested it with my [FABulous LibreLane plugin](https://github.com/mole99/librelane_plugin_fabulous).

Perhaps you could rename `draw_supply_net` to something like `promote_supply_pins` to make it clearer what it does.

As for creating the nets, iirc depending on the order of the flow it can happen that the nets don't exist at this stage. I think [Odb.SetPowerConnections](https://github.com/librelane/librelane/blob/04eabcd48ded94fb566797ece132f258de6e76c9/librelane/flows/classic.py#L56) creates the nets in the Classic flow, but I would need to check. I'd keep it in any case.

mole99:
@KelvinChung2000 There needs to be a copyright statement for this contribution. The easiest way to track the copyright for all contributors is to add:

```
# Copyright © 2026 FABulous Contributors
```

and create an [authors](https://github.com/librelane/librelane/blob/main/Authors.md) file in the project root, like LibreLane does.
If someone is interested in individual contributions, they can check the Git history.

mtetrault:
After looking in the logs and actual tests, I don't understand how the step and script pytests work. Any guidance will be welcome.

KelvinChung2000:
In `tests/gds_flow_test/step_test/conftest.py` This is where we set up the `mock_config`.

In the config, it used to be `VDD_PIN` and `GND_PIN`; since the variable name is updated, they need to be renamed to `VDD_NETS` and `GND_NETS.` 

```py
@pytest.fixture
def mock_config() -> Config:  # type: ignore[name-defined]
    """Create a mock Config object for testing."""
    # Add common config values that steps might use

    return Config(
        {
            "DESIGN_NAME": "test_design",
            "RSZ_CORNERS": None,
            "STA_CORNERS": ["typical"],
            "PDN_VERTICAL_LAYER": "met2",
            "IO_PIN_V_LENGTH": None,
            "IO_PIN_H_LENGTH": None,
            "AUTO_ECO_DIODE_INSERT_MODE": "none",
            "FABULOUS_RUN_TILE_OPTIMISATION": False,
            "FABULOUS_IGNORE_ERROR": False,
            "FABULOUS_IGNORE_ERRORS": False,
            "IGNORE_ANTENNA_VIOLATIONS": False,
            "IGNORE_DEFAULT_DIE_AREA": False,
            "FABULOUS_OPTIMISATION_WIDTH_STEP_COUNT": 5,
            "FABULOUS_OPTIMISATION_HEIGHT_STEP_COUNT": 5,
            "FABULOUS_IO_MIN_WIDTH": 1,
            "FABULOUS_IO_MIN_HEIGHT": 1,
            "FABULOUS_OPT_MODE": OptMode.FIND_MIN_WIDTH,
            "RT_MAX_LAYER": "met2",
            "VDD_NETS": "VDD",
            "GND_NETS": "GND",
        }
    )
```


This will fix one of them. The other would be something like the following:

```py
def test_get_command_uses_custom_metal_layer(
    mock_config: Config, mock_state: State, mocker: MockerFixture
) -> None:
    """Test that get_command() uses the configured PDN_VERTICAL_LAYER value."""
    mocker.patch(
        "librelane.steps.odb.OdbpyStep.get_command",
        return_value=["python", "script.py"],
    )

    # Use a different net name layer
    custom_config = mock_config.copy(VDD_NETS=("A", "C"))
    custom_config = custom_config.copy(GND_NETS=("B", "D"))

    step = FABulousPDN(custom_config, mock_state)
    step.config = custom_config
    command = step.get_command()

    VDD_index = command.index("--power-names")
    VDD_name = command[VDD_index + 1]

    GND_index = command.index("--ground-names")
    GND_name = command[GND_index + 1]

    assert VDD_name == "A", f"Expected 'A', got '{VDD_name}'"
    assert GND_name == "B", f"Expected 'B', got '{GND_name}'"
```

The `custom_config = mock_config.copy(VDD_NETS="A")` is basically overriding and setting up the config.  `command` gives the command of how it is being called. And we are basically checking what is getting passed into the script. You could add a new test to check if it is `None`, it will return "VPWR", etc.

KelvinChung2000 (approved):

PR #699, opened 2026-04-30
feat: support keep attribute and VHDL open mapping in code generators

KelvinChung2000:
- Add ``add_keep`` parameter to ``addComponentInstantiation`` so Verilog instances can emit the ``(* keep *)`` synthesis attribute.
- Emit a per-module ``attr_pack_<name>`` package in the VHDL generator for parity (VHDL accepts the parameter as a no-op).
- Map empty signals to the VHDL ``open`` keyword in component instantiations so unconnected ports are syntactically valid.

### DIFF
--- a/fabulous/fabric_generator/code_generator/code_generator_VHDL.py
+++ b/fabulous/fabric_generator/code_generator/code_generator_VHDL.py
@@ -51,6 +51,9 @@ class VHDLCodeGenerator(CodeGenerator):
             The indentation level
         """
         #   library template
+        self._add(f"package attr_pack_{name} is")
+        self._add(" attribute keep : string;")
+        self._add("end package;")
         self._add("library IEEE;", indentLevel)
         self._add("use IEEE.STD_LOGIC_1164.ALL;", indentLevel)
         self._add("use IEEE.NUMERIC_STD.ALL;", indentLevel)
@@ -429,6 +432,7 @@ end process;
         portsPairs: list[tuple[str, str]],
         paramPairs: list[tuple[str, str]] | None = None,
         emulateParamPairs: list[tuple[str, str]] | None = None,
+        add_keep: bool = False,  # noqa: ARG002 — accepted for API parity; VHDL has no keep attr
         indentLevel: int = 0,
     ) -> None:
         """Add a component instantiation.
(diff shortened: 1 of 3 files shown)


fabulous/fabric_generator/code_generator/code_generator_Verilog.py
```diff
@@ -265,7 +267,10 @@ def addInstantiation(
             paramPairs = []
         if paramPairs:
             port = [f".{i[0]}({i[1]})" for i in paramPairs]
-            self._add(f"{compName}", indentLevel=indentLevel)
+            if add_keep:
+                self._add(f"(* keep *) {compName}", indentLevel=indentLevel)
+            else:
+                self._add(f"{compName}", indentLevel=indentLevel)
```
IAmMarcelJung, line 273: Should we consider moving this logic into `_add` by adding also an `add_keep` parameter there and setting it to default false and only pass the `add_keep` here?

KelvinChung2000: `_add` is a generic newline, and likely not everything will need `keep`.

IAmMarcelJung: That's why I though we add it as default false, but if it does not make sense we can also leave it ofc.

KelvinChung2000: I will prefer keeping the `_add` single function, with no RTL-specific behaviour.

EverythingElseWasAlreadyTaken (approved):

fabulous/fabric_generator/code_generator/code_generator_VHDL.py
```diff
@@ -480,7 +486,8 @@ def addInstantiation(
                 )
             split = signal.split(",")
             if len(split) == 1:
-                connectPair.append(f"{port} => {signal}")
+                rhs = signal or "open"
```
EverythingElseWasAlreadyTaken, line 489: I was thinking, actually we should not generate unconnected pins, so a warning would make sense here. 
But this should be also added to Verilog then.

KelvinChung2000: For Verilog, this will be caught by Verilator linting. The idea is that if someone has done a fabric with a null tile at the border, those "mid ports" will stay unconnected, rather than a port missing. I guess if a warning needs to be raised, it should be at the fabric generation level, not at the code generator level.

PR #700, opened 2026-04-30
chore: decouple fabric from component generator interfaces

KelvinChung2000:
Update the public signatures of the per-component generators (`gen_tile`, `gen_configmem`, `gen_switchmatrix`, `gen_io_pin_config_yaml`) to make their inputs explicit, drop unused parameters, and align return shapes. Tests for each generator are updated to the new interfaces.

### DIFF
--- a/fabulous/fabric_generator/gen_fabric/gen_switchmatrix.py
+++ b/fabulous/fabric_generator/gen_fabric/gen_switchmatrix.py
@@ -24,7 +24,6 @@ from fabulous.fabric_definition.define import (
     Direction,
     MultiplexerStyle,
 )
-from fabulous.fabric_definition.fabric import Fabric
 from fabulous.fabric_definition.tile import Tile
 from fabulous.fabric_generator.code_generator.code_generator import CodeGenerator
 from fabulous.fabric_generator.code_generator.code_generator_VHDL import (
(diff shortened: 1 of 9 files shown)


fabulous/fabric_generator/gen_fabric/gen_switchmatrix.py
```diff
@@ -68,6 +67,12 @@ def genTileSwitchMatrix(
         `.list` format. If None, the CSV is written to the same directory as the
         source `.list` file. This parameter is ignored when the input is already
         a `.csv` file.
+    config_bit_mode : ConfigBitMode
+        The configuration-bit mode for the tile (frame-based or flip-flop chain).
+    multiplexer_style : MultiplexerStyle
+        The multiplexer style used to implement switch-matrix muxes.
+    generate_delay_in_switch_matrix : int
```
EverythingElseWasAlreadyTaken: I would rather call it default pip delay or similar, to make it more obvious what It's actually about.

fabulous/fabric_generator/gen_fabric/gen_switchmatrix.py
```diff
@@ -238,9 +243,9 @@ def genTileSwitchMatrix(
     # for example in terminating switch matrices at the fabric borders,
     # we may just change direction without any switching
     if noConfigBits > 0:
-        if fabric.configBitMode == "ff_chain":
+        if config_bit_mode == "ff_chain":
```
EverythingElseWasAlreadyTaken, line 246: Not directly related to this PR, but the naming ff_chain should be changed or even kicked out at some point. Not sure if this is even supported in the rest of the flow. 
 
But the whole FlipFlopChain and ff_chain stuff needs to be reviewed and tested again anyways.

EverythingElseWasAlreadyTaken (approved):
LGTM, just a small comment

fabulous/fabric_generator/gen_fabric/gen_switchmatrix.py
KelvinChung2000: I will make this into an issue. While the FF chain is the worst in all cases, people might still prefer this for simplicty. It will likely stay, but we also need it tested. I will open an issue on this.

EverythingElseWasAlreadyTaken (approved):

PR #701, opened 2026-04-30
chore: disable pdk download during test

KelvinChung2000:


IAmMarcelJung:
If I got it right, this disables the PDK download for the tests? If so could you maybe update the title and/or add a short description?

EverythingElseWasAlreadyTaken (approved):

PR #704, opened 2026-05-01
feat: to allow preserve .list-file mux input order

### LINKED ISSUE #437: Order of inputs to multiplexer not preserved
The entry in my list file is:

```
[GNT|GNT|GNT|GNT],[GND0|VCC0|S2MID0|S2MID1]
[GNT|GNT|GNT|GNT],S1END[3|2|1|0]
```

If I change it to this:

```
[GNT|GNT|GNT|GNT],[GND0|S2MID0|S2MID1|VCC0]
[GNT|GNT|GNT|GNT],S1END[1|3|2|0]
```

Both produce a mux8 with the same order of inputs:

```
 //switch matrix multiplexer GNT MUX-8
assign GNT_input = {VCC0,GND0,S2MID1,S2MID0,S1END3,S1END2,S1END1,S1END0};
cus_mux81_buf inst_cus_mux81_buf_GNT (
    .A0(GNT_input[0]),
    .A1(GNT_input[1]),
    .A2(GNT_input[2]),
    .A3(GNT_input[3]),
    .A4(GNT_input[4]),
    .A5(GNT_input[5]),
    .A6(GNT_input[6]),
    .A7(GNT_input[7]),
    .S0(ConfigBits[48+0]),
    .S0N(ConfigBits_N[48+0]),
    .S1(ConfigBits[48+1]),
    .S1N(ConfigBits_N[48+1]),
    .S2(ConfigBits[48+2]),
    .S2N(ConfigBits_N[48+2]),
    .X(GNT)
);
```

It seems that the entries are sorted? However, I have to rely on the order of the inputs to the multiplexer, as I need to ensure that GND0 is selected for an all-zero bitstream.

Is there any way to preserve the order as given in the list file? I'm using FABulous2.0-development.

ThVerg:
This adds a fabric-level toggle `PreserveListOrder` (default `FALSE`),
so the change is non-breaking and people can opt in.

### How it works

- `list2CSV` writes a per-row 1-based index to each connection cell
  (1, 2, 3, ...) instead of just `1` when the toggle is on, encoding
  the connection's position in the .list file.
- `parseMatrix` sorts cells by `(value, column_index)`. The integer
  recovers the user-specified .list order. The column-index secondary
  key means hand-written all-1 CSVs degrade to today's column-order
  behaviour, so existing tooling keeps working.
- With the toggle off (default) `list2CSV` writes `1` and the sort is
  column-order — bit-identical to current behaviour.

I also reordered the constant entries in the LUT4AB / DSP_bot / RAM_IO /
W_IO .list templates so `GND0` / `VCC0` land at A0 (or A0/A1 where both
are present).

### Verification

- toggle off + pristine templates → counter demo passes (today's
  behaviour, non-breaking)
- toggle on + reordered templates → counter demo passes (.list order
  preserved end-to-end through RTL and bitstream)
- toggle off + reordered templates → counter demo passes (template
  reorderings are a no-op when order is ignored)
- toggle on + pristine templates → counter demo fails — confirms the
  template reorderings are needed for the opt-in path


KelvinChung2000:
Can you please also add some documentation for the new option at `docs/source/user_guide/building_doc/fabric_definition.md`

ThVerg:
> Can you please also add some documentation for the new option at `docs/source/user_guide/building_doc/fabric_definition.md`

Yeah, will do

fabulous/fabric_generator/gen_fabric/gen_helper.py
```diff
@@ -80,6 +82,12 @@ def list2CSV(InFileName: Path, OutFileName: Path) -> None:
         The input file name of the list file
     OutFileName : Path
         The directory of the CSV file to be written
+    preserveListOrder : bool, optional
+        When True, every connection is written as its 1-based position in
+        the .list file for that row (1, 2, 3, ...) so parseMatrix can
+        recover the user-specified mux input order. When False (default),
+        every connection is written as `1` and order is determined by
+        CSV-column position when read back, matching legacy behaviour.
```
KelvinChung2000:
```suggestion
        By default, every connection is written as `1`, and order is determined by
        CSV-column position when read back. When it is set to `True`, every 
        connection is written as its 1-based position in the .list file for that row (1, 2, 3, ...), 
        so `parseMatrix` can recover the user-specified mux input order.
```

This might need to be reformatted.

fabulous/fabric_generator/gen_fabric/gen_helper.py
```diff
@@ -108,7 +116,14 @@ def list2CSV(InFileName: Path, OutFileName: Path) -> None:
     destination = file[0].strip("\n").split(",")[1:]
     source = [file[i].split(",")[0] for i in range(1, len(file))]
 
-    # set the matrix value with the provided connection pair
+    # When `preserveListOrder` is on, give each new connection a unique
+    # 1-based per-row index so the order can be recovered later. Otherwise
+    # all connections are marked with 1 (legacy behaviour). The per-row
+    # counter starts past any existing non-zero values that the bootstrap
+    # CSV may already contain.
```
KelvinChung2000: Move this to the function doc string. Can remove the "(legacy behaviour)"

fabulous/fabric_generator/parser/parse_switchmatrix.py
```diff
@@ -55,8 +55,19 @@ def parseMatrix(fileName: Path, tileName: str) -> dict[str, list[str]]:
         port_name, row = fields[0], fields[1:]
         if not port_name:
             continue
-        # collect destinations where the connection bit is set
-        connections[port_name] = [dest_list[k] for k, v in enumerate(row) if v == "1"]
+        # B1: any non-zero integer is a connection; the integer encodes the
+        # mux input position (lower = earlier). Sort by (value, column) so
+        # legacy all-1 rows degrade to CSV-column order via the secondary key.
```
KelvinChung2000: can move to the function doc string. Can remove the mention of legacy.

fabulous/fabric_generator/parser/parse_switchmatrix.py
```diff
@@ -55,8 +55,19 @@ def parseMatrix(fileName: Path, tileName: str) -> dict[str, list[str]]:
...
+        items: list[tuple[int, int, str]] = []
+        for k, v in enumerate(row):
+            try:
+                value = int(v)
+            except ValueError:
+                continue
```
KelvinChung2000: Rather than silently skip, we should raise an error.

KelvinChung2000 (changes requested):
Just some minor doc changes. In general, rather than stating it as legacy behaviour, it should be phrased more like "sort by column"/"no sorting".

KelvinChung2000:
To run the linting, you can call `pre-commit run --all-files`, which will run all the checks that we have.

KelvinChung2000 (approved):

mole99:
Hi @ThVerg, thanks for your great work!

My only suggestion is to swap the bit order.

As you mentioned in the issue, you had to swap _SR,[J_SR_END0|GND0] → _SR,[GND0|J_SR_END0] because you chose A0 to be on the left. However, my intuition would be for A0 to be on the very right (i.e. MSB first). This matches the "downto" behavior of Verilog and VHDL, and was probably the original intention of the author of this list file.

KelvinChung2000 (approved):

EverythingElseWasAlreadyTaken (approved):
Looks good to me.
I've tested it, and it seems to not brick anything and should be fully legacy compatible. 👍 

Thanks a lot for the PR!

PR #706, opened 2026-05-05
chore: bump dependency

KelvinChung2000:


pyproject.toml
```diff
@@ -175,3 +175,6 @@ FABulous-bit-gen = "FABulous_bit_gen"
 
 [tool.uv.sources]
 sdf-timing = { git = "https://github.com/FPGA-Research/f4pga-sdf-timing.git" }
+
+[tool.uv.extra-build-dependencies]
```
EverythingElseWasAlreadyTaken, line 179: Why is this needed here? 
Shouldn't we rather add this directly in the SDF repo?

EverythingElseWasAlreadyTaken (approved):

pyproject.toml
KelvinChung2000: It's just simpler for me to do it this side than to patch that. I will update on that side.

EverythingElseWasAlreadyTaken: ok, should we change it or leave it? 

I just noticed, that we have the sdf-timing from git, which we might have to change for pypi deployment, since they require all dependency packages to be available on pypi as well.

KelvinChung2000: This is what I just noticed as well when I made the change. I have pushed on that side, so if you install via `uv`, things should work, but only via `uv`.  For the use case of SDF, it will almost always pair with Nix, so technically speaking, a wrongly behaved package is "fine"

EverythingElseWasAlreadyTaken: I remember. 
Let's see if PyPI allows this. :) 

So I'll merge this if the checks are done.

KelvinChung2000: I can't get it working without the following line. It seem like is something specifc to `uv2nix` needing it for resolving.

EverythingElseWasAlreadyTaken (approved):

PR #708, opened 2026-05-06
fix(geometry): support T-shape and thin fabrics

### LINKED ISSUE #600: Geometry generation does not support T-shaped fabrics
**Describe the bug**
The current geometry generation does not fully support T-shaped fabrics. 

**To Reproduce**

Use https://github.com/IAmMarcelJung/ICESOC_FABulous_user_project/tree/fix/correct_fabric_csv as a FABulous project and generate the geometry. 

**Expected behavior**
The geometry should be correctly generated.

**Other hints**
The issue lies somehow in the handling of the termination tiles that are not in the first or last row of the fabric, but in the side of the fabric, as it is the case with the `N_term_RAM_IO` in the provided example. Instead of `Border.NORTHSOUTH`, `Border.EASTWEST` gets assigned to the `tileGeom.border` in `fabric_geometry.py`. Also the handling in https://github.com/FPGA-Research/FABulous/blob/main/fabulous/geometry_generator/fabric_geometry.py#L84-L90 does not cover the T-shape case.

ThVerg:
Closes #600

Two related issues in `fabulous/geometry_generator/`:

- **T-shape concavity tiles** were classified as non-border because border
  detection only checked grid-edge position, not whether a NULL cell sat next
  to the tile. Their wires then routed off into empty cells.
- **Thin (1×N / N×1) fabrics** crashed: every interior tile is a border-axis
  tile but has no same-axis neighbour, so `neighbourConstraints` was `None`
  and the alignment block dereferenced it.

Changes:

- `fabric_geometry.py` — widen `northSouth` / `eastWest` to also flag NULL
  adjacency; bounds-check the neighbour-pick so we try the opposite side if
  the first candidate is NULL; switch `tileGeom == queried` to identity
  (`is`) since `TileGeometry` is a dataclass and two distinct instances with
  default fields compare equal.
- `tile_geometry.py` — gate the alignment block with
  `neighbourConstraints is not None`; when the constraint queue is exhausted
  mid-loop, fall through to the same local-stride pattern as the non-border
  branch instead of clamping to 0.

### DIFF
--- a/fabulous/geometry_generator/fabric_geometry.py
+++ b/fabulous/geometry_generator/fabric_geometry.py
@@ -81,18 +81,24 @@ class FabricGeometry:
                 tileName = tile.name
                 tileGeom = self.tileGeomMap[tileName]
 
-                if tileGeom == queried:
+                if tileGeom is queried:
                     searchedTile = None
+                    # Pick the in-fabric neighbour on the relevant axis:
+                    # for NORTHSOUTH borders that means the tile above or
+                    # below (whichever is non-NULL); for EASTWEST it's the
+                    # tile to the left or right. T-shaped fabrics can have
+                    # one side be NULL even when i / j is not at the grid
+                    # bounds, so we fall back to the opposite direction.
                     if queried.border == Border.NORTHSOUTH:
-                        if i == 0:
+                        if i + 1 < self.fabric.numberOfRows:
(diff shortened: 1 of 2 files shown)


KelvinChung2000 (approved):

IAmMarcelJung (approved):
Thanks for implementing this! LGTM and works for a tested T-shaped fabric.

PR #714, opened 2026-05-11
feat: make NVC the default VHDL simulator (GHDL is fallback)

EverythingElseWasAlreadyTaken:
NVC is significantly faster than GHDL for FABulous simulation and will be
used as the default simulator when available, with automatic fallback to GHDL.

Simulation setup now supports both NVC and GHDL with:
- Auto-detection of available simulator (NVC preferred, GHDL fallback)
- SIMULATOR variable to explicitly choose nvc/ghdl/auto
- NVC-specific and GHDL-specific configuration flags

VHDL attribute naming fixes:
NVC enforces stricter rules for VHDL attribute names - they cannot conflict
with signal or other element names. Updated affected RTL files:
- MULADD.vhdl
- MUX8LUT_frame_config_mux.vhdl
- RegFile_32x4.vhdl

Documentation updates:
- Updated simulation setup guide to recommend NVC
- Added NVC installation and configuration details
- Updated CAD tool installation guide
- Updated Taskfile and Makefile with simulator selection logic

CI Update:
- Add nickg/setup-nvc@v1 to the prepare_FABulous_container composite action so all CI jobs have NVC available.
- Fix backward compatibility check action was silently failing.

Misc: 
- Add NVC to nix setup.

### DIFF
--- a/.github/actions/prepare_FABulous_container/action.yaml
+++ b/.github/actions/prepare_FABulous_container/action.yaml
@@ -92,15 +92,7 @@ runs:
         github-token: ${{ github.token }}
         version: ${{ steps.oss-cad-version.outputs.version }}
     - name: Install NVC VHDL simulator
-      shell: bash
-      run: |
-        UBUNTU_VERSION=$(lsb_release -rs)
-        NVC_TAG=$(curl -sf -H "Authorization: Bearer ${{ github.token }}" \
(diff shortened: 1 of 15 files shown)


KelvinChung2000:
The nix flake will also want the nvc. See whether you can just add it in `allPackages` with something like `pkgs.vnc`. If that does not work, then probably in a separate PR that includes it later

KelvinChung2000 (approved):

PR #715, opened 2026-05-11
feat: allow nix to start anywhere

KelvinChung2000:
This should do what #690 need to do.


IAmMarcelJung (approved):
Thanks for this, works perfectly!

IAmMarcelJung (approved):

PR #748, opened 2026-05-12
fix: avoid temp file collisions on multi-user machines

EverythingElseWasAlreadyTaken:
Three places wrote to fixed, shared paths in the system temp directory, causing permission errors or silent data corruption when multiple users (or parallel test workers) ran FABulous concurrently:

- yosys_obj.py: `my_package.vhd` was always written to the same path (`/tmp/my_package.vhd`). Replaced with a `NamedTemporaryFile` scoped to the VHDL branch with a `finally` cleanup, so each invocation gets a unique path.
- fabulous_cli.py: the OpenROAD TCL script used a unique name (no collision risk) but was never deleted. Added `try/finally` cleanup.
- test_yosys_obj.py: `setup_mocks` wrote `models_pack.v` to the shared temp dir. Switched to pytest's `tmp_path` fixture, which gives each test an isolated directory, safe for parallel workers.

### DIFF
--- a/fabulous/fabulous_cli/fabulous_cli.py
+++ b/fabulous/fabulous_cli/fabulous_cli.py
@@ -1539,13 +1539,16 @@ class FABulous_CLI(Cmd):
             script_file.write(f"read_db {db_file}\n")
             file_name = script_file.name
         logger.info(f"Start OpenROAD GUI with odb: {db_file}")
-        sp.run(
-            [
-                str(openroad),
-                "-gui",
-                str(file_name),
-            ]
-        )
+        try:
+            sp.run(
+                [
+                    str(openroad),
+                    "-gui",
+                    str(file_name),
+                ]
+            )
+        finally:
+            Path(file_name).unlink(missing_ok=True)
 
     @with_argparser(gui_parser)
     @with_category(CMD_TOOLS)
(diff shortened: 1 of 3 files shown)


fabulous/fabulous_cli/fabulous_cli.py
```diff
@@ -1539,13 +1539,16 @@ def do_start_openroad_gui(self, args: argparse.Namespace) -> None:
...
-        )
+        try:
+            sp.run(
+                [
+                    str(openroad),
+                    "-gui",
+                    str(file_name),
+                ]
+            )
+        finally:
```
KelvinChung2000, line 1550: Just double-check if the behaviour is intended? Or should it raise and clean up?

KelvinChung2000 (approved):

fabulous/fabulous_cli/fabulous_cli.py
EverythingElseWasAlreadyTaken: I would temp files expect to be always cleaned up, or should we save it?

KelvinChung2000: I would just leave it and let the os to clean up.

KelvinChung2000 (approved):

PR #755, opened 2026-05-12
fix: resolve pre-commit errors after dependency updates

EverythingElseWasAlreadyTaken:
Recent pre-commit package updates surfaced several linting and formatting issues that now need to be addressed:

- ruff: fix E501 (line too long) in fabulous.py docstring
- pydoclint: fix DOC001/101/103 in code_generator.py — four Parameters sections had excess indentation that pydoclint could not parse; also reorder Parameters before Examples in addLogicEnd
- docformatter: add [tool.docformatter] config to pyproject.toml and move --black flag from pre-commit args into the config file; disable wrap-summaries/wrap-descriptions (set to 0) to stop the line-joining oscillation with ruff format that affected ~25 files; exclude custom_exception.py (docformatter v1.7.8 hardcodes one blank line after module docstrings, conflicting with ruff format's E302) and verilog_gate_level.py (verbose r-string regex patterns passed to re.compile() are misidentified as docstrings and reformatted)
- apply docformatter and ruff format auto-fixes across remaining files


KelvinChung2000:
I have just done my fixes as well... 

Anyway, took me a while to figure out what is wrong. I think we should just drop the docformatter pre-commit. Ruff covers it.

EverythingElseWasAlreadyTaken:
> I have just done my fixes as well...
> 
> Anyway, took me a while to figure out what is wrong. I think we should just drop the docformatter pre-commit. Ruff covers it.

Yeah, was a bit weird that they started to work against each other, also took me a while to fix. 

From what I understood from the ruff docs, it only supports formatting code snippets in docstrings:
https://docs.astral.sh/ruff/formatter/#docstring-formatting

KelvinChung2000:
It is implied by the D rule. It covers pydocstyle, which covers PEP257, which covers what docformatter does.

KelvinChung2000:
If we want anything, it might actually be https://numpydoc.readthedocs.io/en/latest/validation.html#built-in-validation-checks which is tracking at ruff via https://github.com/astral-sh/ruff/issues/8425

EverythingElseWasAlreadyTaken:
> If we want anything, it might actually be https://numpydoc.readthedocs.io/en/latest/validation.html#built-in-validation-checks which is tracking at ruff via [astral-sh/ruff#8425](https://github.com/astral-sh/ruff/issues/8425)

This should be also covered by pydoclint.

PR #769, opened 2026-05-18
fix: show full version including dev tag on RTD

EverythingElseWasAlreadyTaken:
get_display_version was stripping the .dev suffix, causing the sidebar to show e.g. v2.0.0 instead of v2.0.0.dev+{hash}, leading to the assumption, that this is the doc of a stable version not the current git HEAD.

### DIFF
--- a/docs/source/conf_helper.py
+++ b/docs/source/conf_helper.py
@@ -46,5 +46,8 @@ def get_version() -> str:
 
 
 def get_display_version(version_text: str) -> str:
-    """Strip development and local suffixes from the displayed version tag."""
-    return re.sub(r"\.dev.*$", "", version_text)
+    """Return bare version for tagged releases, or version+dev for dev builds."""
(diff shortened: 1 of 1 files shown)


KelvinChung2000:
I did the stripping because the hash looks ugly. 

Right now it is still showing the `b6` if you want to be more explict, I would go with adding `+dev` but omitting the hash
Screenshot 2026-05-18 at 12 50 14

EverythingElseWasAlreadyTaken:
> I did the stripping because the hash looks ugly.
> 
> Right now it is still showing the `b6` if you want to be more explict, I would go with adding `+dev` but omitting the hash Screenshot 2026-05-18 at 12 50 14

I would go for +dev(main).
Once the release is out, we can do a fixed version on every major. 
Have to see how to automate this.

KelvinChung2000 (approved):

PR #804, opened 2026-05-26
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `Medium` | **File**: `fabulous/fabric_files/FABulous_project_template_common/Test/makehex.py:L12`

The makehex.py script directly uses command-line arguments (argv[1], argv[2], argv[3]) without any sanitization or validation. While the script itself is simple, if this pattern is used elsewhere or if this script is called with untrusted input, it could lead to command injection. More critically, the script reads binary files and writes output without path validation, which could lead to path traversal if attacker-controlled paths are passed.

## Solution

Add input validation for command-line arguments, use argparse with proper type checking, validate file paths to prevent directory traversal, and consider using sys.stdin/stdout instead of file arguments for safer operation.

## Changes

- `fabulous/fabric_files/FABulous_project_template_common/Test/makehex.py` (modified)
- `fabulous/processpool.py` (modified)

### DIFF
--- a/fabulous/fabric_files/FABulous_project_template_common/Test/makehex.py
+++ b/fabulous/fabric_files/FABulous_project_template_common/Test/makehex.py
@@ -8,21 +8,46 @@ software, either in source code form or as a compiled binary, for any purpose,
 commercial or non-commercial, and by any means.
 """
 
+import argparse
+import os
 from pathlib import Path
-from sys import argv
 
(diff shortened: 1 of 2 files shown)


KelvinChung2000:
Unrelated.

PR #836, opened 2026-06-10
feat: add standard-cell spec and switch-matrix construct

KelvinChung2000:
Part of the changes factor out from #835 

This creates a cell spec that allows us to use standard cell information for whatever purpose in the future. Flexibility has been considered. So let's say that later, if we want a Multiplex cell spec, we can inherit and extend it from the base cell spec.

### DIFF
--- a/fabulous/fabric_definition/fabric.py
+++ b/fabulous/fabric_definition/fabric.py
@@ -682,6 +682,25 @@ class Fabric:
 
         return result
 
+    def get_super_tile_containing(self, tile_name: str) -> SuperTile | None:
+        """Return the SuperTile that contains the named tile, if any.
+
+        Parameters
+        ----------
+        tile_name : str
+            Name of the (sub-)tile to look up.
+
+        Returns
+        -------
+        SuperTile | None
+            The SuperTile whose constituent tiles include ``tile_name``, or None
+            if the tile is not part of any SuperTile.
+        """
+        for super_tile in self.superTileDic.values():
+            if any(tile.name == tile_name for tile in super_tile.tiles):
+                return super_tile
+        return None
+
     def get_tile_row_column_indices(self, tile_name: str) -> tuple[set[int], set[int]]:
         """Get all row and column indices where a tile type appears.
 
--- a/fabulous/fabric_definition/switch_matrix.py
+++ b/fabulous/fabric_definition/switch_matrix.py
@@ -1,0 +1,71 @@
+"""Switch matrix construct for the FABulous fabric model.
+
+A tile's switch matrix is the programmable interconnect: which sources may drive
+each destination inside the tile. The connectivity is declared in the tile's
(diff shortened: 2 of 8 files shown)


fabulous/fabric_definition/fabric.py
```diff
@@ -542,6 +542,25 @@ def get_all_unique_tiles(self) -> list[Tile | SuperTile]:
 
         return result
 
+    def get_super_tile_containing(self, tile_name: str) -> SuperTile | None:
```
EverythingElseWasAlreadyTaken, line 685: This function is only used in tests, shouldn't we just implement it there as a helper? 

Also, would maybe make more sense, that the Tile class has a back pointer to the supertile, which we can use instead of the boolean isPartOfSupertile we currently have.

fabulous/fabric_definition/switch_matrix.py
```diff
@@ -0,0 +1,71 @@
+"""Switch matrix construct for the FABulous fabric model.
+
+A tile's switch matrix is the programmable interconnect: which sources may drive
+each destination inside the tile. The connectivity is declared in the tile's
+matrix file (a ``.csv`` adjacency matrix or a ``.list`` of pairs) and parsed by
+:mod:`fabulous.fabric_generator.parser.parse_switchmatrix`. This light dataclass
+gives that switch matrix a first-class home in the fabric model so callers work
+with a ``SwitchMatrix`` object instead of re-deriving the file dispatch and
+parsing at every site.
+"""
```
EverythingElseWasAlreadyTaken, line 10: I'm not sure what the point of the switch matrix class is, if it does not hold the actual data. 
It might be better if the switch matrix class just gets filled with the actual data, so parsing is only done once. 

The parsers should be class methods, CSV and list are basically different representations of the same data, so we could just read whatever is there instead of converting list->csv->rtl.
For the rest, we can offer conversion functions through the CLI, that convert list to csv, an vice versa. 
(I also have a rtl to list function lying around somewhere, that could be added here).

EverythingElseWasAlreadyTaken: I already prepared something like this, that we could use.

fabulous/fabric_definition/tile.py
```diff
@@ -292,6 +293,17 @@ def getTileOutputNames(self) -> list[str]:
...
+    @property
+    def switch_matrix(self) -> SwitchMatrix:
+        """Get the tile's switch matrix as a fabric-model construct.
+
+        Returns
+        -------
+        SwitchMatrix
+            The switch matrix wrapping this tile's matrix file and config bits.
+        """
+        return SwitchMatrix(self.name, self.matrixDir, self.matrixConfigBits)
```
EverythingElseWasAlreadyTaken, line 324: Shouldn't this be actually a variable that hold a switch matrix object?

EverythingElseWasAlreadyTaken (approved):
Discussed in Call, that we leave it basic as it is and extend the swtichmatrix class as described in 2.1 release.

PR #837, opened 2026-06-10
feat: add Yosys, OpenSTA and GHDL subprocess wrappers

KelvinChung2000:
Part of the changes factor out from #835

Add the Tool base class (single subprocess entry point + Jinja template rendering) and the GHDL, Yosys and OpenSTA wrappers, with the yosys_synth and opensta_sdf script templates. Adds jinja2 as a dependency.

This allows us to more easily extend the use of tools and script these external tools.


fabulous/template/yosys_synth.j2
EverythingElseWasAlreadyTaken, line 1: If I see this right, this is only used for testing?  
Might be a bit overkill to have and test this, if there is no real use.

EverythingElseWasAlreadyTaken (approved):
Overall ok, just the question if we need the yosys_synth script, since it is only for testing, and we are already testing with the opensta script.

fabulous/template/yosys_synth.j2
KelvinChung2000: This is something used inside the timing engine for doing analysis

PR #843, opened 2026-06-15
fix(gen_switchmatrix): fix generic switch matrix config bit slicing

q574500110:
## Summary

Fix an invalid bit-slice generated by the generic switch matrix path.

With `multiplexerStyle = generic`, the first configurable mux could end up using `ConfigBits[-1:0]`, which is not valid Verilog. This change corrects the selection-bit indexing so single-bit muxes use `ConfigBits[pos]` and wider muxes use a valid descending slice.

## Related Issues

None.

## Testing

Checked the generic switch matrix generation logic and confirmed the first mux no longer produces a negative bit slice.

## Checklist

- [ ✓] I have read [`CONTRIBUTING.md`](../CONTRIBUTING.md).
- [ ✓] I agree that my contribution is licensed under the project's [Apache License 2.0](../LICENSE).
- [ ✓] I have run `pre-commit run --all-files` and `pytest` locally.
- [ ✓] I have updated documentation where relevant.

### DIFF
--- a/fabulous/fabric_generator/gen_fabric/gen_switchmatrix.py
+++ b/fabulous/fabric_generator/gen_fabric/gen_switchmatrix.py
@@ -365,10 +365,17 @@ def genTileSwitchMatrix(
                     )
             else:
                 # generic multiplexer
+                select_width = paddedMuxSize.bit_length() - 1
+                if select_width == 1:
+                    select_index = f"ConfigBits[{configBitstreamPosition}]"
+                else:
+                    select_index = (
+                        f"ConfigBits[{configBitstreamPosition + select_width - 1}:"
+                        f"{configBitstreamPosition}]"
+                    )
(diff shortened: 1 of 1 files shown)


KelvinChung2000:
Thanks for the PR and bug fixes.

During the review, I noticed another problem that might lead to an issue. So wires like `LA_I0_input` are used but never assigned any value. I will open a PR to fix this for you asap.

KelvinChung2000:
I have opened #850.

EverythingElseWasAlreadyTaken (approved):
Thanks for the PR.
This looks good overall. 

@KelvinChung2000 We should add more tests for the default mux gen and also the non power of 2 mux warning, is currently only for custom mux, so it should be moved.

KelvinChung2000:
Suppressed by #851

PR #851, opened 2026-06-15
fix: fix and improve generic mux generation

KelvinChung2000:


fabulous/fabric_generator/code_generator/code_generator_VHDL.py
```diff
@@ -425,6 +425,42 @@ def addAssignVector(
...
+            Index of the lowest select bit.
+        selectWidth : int
+            Number of select bits.
+        delay : int
+            Delay in picoseconds.
+        indentLevel : int
+            The indentation level.
+        """
+        selectHigh = selectLow + selectWidth - 1
+        index = f"to_integer(unsigned({selectVector}({selectHigh} downto {selectLow})))"
```
EverythingElseWasAlreadyTaken, line 461: Should we also add the selectWidth == 1 case here?

EverythingElseWasAlreadyTaken (approved):
Looks good.

fabulous/fabric_generator/code_generator/code_generator_VHDL.py
KelvinChung2000: We could; this should make things cleaner. I will include that as well

fabulous/fabric_generator/code_generator/code_generator_Verilog.py
```diff
@@ -460,6 +460,33 @@ def addAssignVector(
...
+            selectLow: Index of the lowest select bit
+            selectWidth: Number of select bits
+            delay: Assignment delay
+            indentLevel: The indentation level
+        """
+        if selectWidth == 1:
+            index = f"{selectVector}[{selectLow}]"
+        else:
+            index = f"{selectVector}[{selectLow + selectWidth - 1}:{selectLow}]"
+        self._add(f"assign {output} = #{delay} {inputVector}[{index}];", indentLevel)
```
EverythingElseWasAlreadyTaken: I think we usually don't use the delay in Verilog.

KelvinChung2000: But having them should allow for more "truthful" testing?

KelvinChung2000: Since you mentioned before having an issue with the 0 delay simulator race condition

fabulous/fabric_generator/code_generator/code_generator_VHDL.py
KelvinChung2000: Turns out vhdl for single bit select needs `0 downto 0`

fabulous/fabric_generator/code_generator/code_generator_Verilog.py
EverythingElseWasAlreadyTaken: This is true, but haven't used it till now for any Verilog write function.

fabulous/fabric_generator/code_generator/code_generator_VHDL.py
EverythingElseWasAlreadyTaken:
```
index = f"to_integer(unsigned({selectVector}({selectLow})))" 
```
won't work?

EverythingElseWasAlreadyTaken (approved):

PR #852, opened 2026-06-15
chore: speed up and extend CI

KelvinChung2000:


EverythingElseWasAlreadyTaken (commented):
Testing the whole pytest run in nix containers takes quite some time, not sure if we should do this on every PR, since there should be fundamentally no difference, if there are just some FABulous changes. 

The longest nix run took ~50 minutes, while the longest container run was ~8 minutes:
https://github.com/FPGA-Research/FABulous/actions/runs/27574099282/usage

I would suggest, we run the nix pytest just on nightly for nix and on changes on the nix env.

EverythingElseWasAlreadyTaken (approved):

PR #878, opened 2026-06-29
fix: fix broken 2+ wide super tile and extend testing

### LINKED ISSUE #875: Wrong internal supertile wires when >1 column
**Describe the bug**
When hardening my BRAM tile, which has 5 rows and 2 columns of tiles, the verilog file for the supertile contains errors in the X indexes for the wires that pass the FrameData output of the column 0 into the column 1 of the supertile. The X index is shited by 1, so it creates wires that duplicate the outputs of the right column, instead of creating the wires that connect both columns.

Image

**To Reproduce**
```shell
cd FABulous
# Activate venv
source ~/FABulous/.venv/bin/activate
# Activate Nix env
FABulous nix-env
cd <my_poject>
fabulous start
# Generate fabric
FABulous> load_fabric
FABulous> gen_fabric
FABulous> gen_top_wrapper
# GDS flow
fabulous> gen_tile_macro BRAM_32 --optimise balance

```

**Expected behavior**
These are the generated wires:
Image

Which duplicate these outputs:
Image

And these are the expected wires:
Image

I have not tried with wider supertiles, but I assume it would translate to those as well.
I can share my supertile files if anyone wants to run the exact flow.

KelvinChung2000:
close #875 

I extend testing to a few other chaotic shapes.

### DIFF
--- a/fabulous/fabric_definition/supertile.py
+++ b/fabulous/fabric_definition/supertile.py
@@ -92,6 +92,8 @@ class SuperTile:
         internalConnections = []
         for y, row in enumerate(self.tileMap):
             for x, tile in enumerate(row):
+                if tile is None:
+                    continue
                 if (
                     0 <= y - 1 < len(self.tileMap)
                     and self.tileMap[y - 1][x] is not None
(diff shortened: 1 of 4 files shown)


IAmMarcelJung (approved):
LGTM

PR #887, opened 2026-07-03
build: adopt cmd2 4.0 from PyPI

KelvinChung2000:
Stacked PRs:
 * #896
 * #895
 * #894
 * #893
 * #892
 * #891
 * #890
 * #889
 * #888
 * __->__#887

--- --- ---

### build: adopt cmd2 4.0 from PyPI

Bump cmd2 to >=4 for the annotated command API. cmd2 4.0.0 is now
published on PyPI, so no git source pin is required. cmd2 4.0 pulls in
rich>=15, which ciel caps at rich<15, so add the project-local uv
override-dependencies for both the root and docs projects.

### DIFF
--- a/docs/pyproject.toml
+++ b/docs/pyproject.toml
@@ -14,12 +14,12 @@ dependencies = [
   "sphinx-prompt==1.10.2",
   "sphinxcontrib-bibtex>=2.6.0",
   "sphinx-autodoc-typehints>=3.12.1", # Automatic type hint cross-references
-  "sphinx-autoapi>=3.0.0", # Modern alternative to autosummary
-  "myst-parser>=5.1.0", # Markdown support (industry standard)
-  "sphinx-design>=0.5.0", # Modern UI components
-  "sphinxext-opengraph>=0.9.0", # Social media preview cards
+  "sphinx-autoapi>=3.0.0",            # Modern alternative to autosummary
+  "myst-parser>=5.1.0",               # Markdown support (industry standard)
+  "sphinx-design>=0.5.0",             # Modern UI components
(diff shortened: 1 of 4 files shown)


docs/pyproject.toml
```diff
@@ -26,12 +26,18 @@ dependencies = [
   "sphinxcontrib-mermaid>=2.0.0",
 ]
 
+# cmd2 4.0 (pulled in transitively via the editable FABulous package) requires
+# rich>=15, which ciel's conservative rich<15 cap conflicts with. uv overrides are
+# project-local, so the docs project repeats the override the root project applies.
+[tool.uv]
```
EverythingElseWasAlreadyTaken: Could this become an issue if you don't use uv? Like just normal pip?

pyproject.toml
```diff
@@ -161,20 +158,21 @@ markers = [
...
 
 [tool.deptry.per_rule_ignores]
-DEP002 = ["pyreadline3", "gnureadline", "go-task-bin", "FABulous-bit-gen"]
+DEP002 = ["go-task-bin", "FABulous-bit-gen"]
 
 [tool.deptry.package_module_name_map]
 FABulous-bit-gen = "FABulous_bit_gen"
 
+[tool.uv]
+override-dependencies = ["rich>=15.0.0"]
```
EverythingElseWasAlreadyTaken: same here.

docs/pyproject.toml
KelvinChung2000: But if in pip, how would this be resolved? Does it mean it would never work?

EverythingElseWasAlreadyTaken (approved):

PR #790, opened 2026-05-25
chore: doc setup audit

KelvinChung2000:

PR #895, opened 2026-07-03
docs: rewrite command-doc generator for cmd2 CommandSets

KelvinChung2000:
Stacked PRs:
 * __->__#895
 * #894

--- --- ---

### docs: rewrite command-doc generator for cmd2 CommandSets

Generate the REPL command reference by introspecting the live REPL classes
instead of scraping their source with the ast module. FABulousREPL and every
ReplCommandSet subclass are imported, and each do_* method is read through the
objects cmd2 already builds: the @with_category help-category string and the
@with_annotated argument parser. Required-ness, defaults, choices and flag
spellings come straight from cmd2's resolved argparse actions, and parameter
types from the live type hints, so the docs cannot drift from what the REPL
actually accepts.

This removes a class of silent failures the source-scraping approach had:
optional positionals (nargs='?') are now reported optional rather than
required; the category order follows the CMD_* definition order with "Other"
last (Timing Characterization no longer sorts after it); an unknown category
or a decorator/import alias can no longer downgrade a command to "Other"
unnoticed; and markdown pipes in types/help/choices/defaults are escaped so
they can't break the table. Also refresh the REPL command-reference pages.

### DIFF
--- a/docs/source/_ext/generate_repl_docs.py
+++ b/docs/source/_ext/generate_repl_docs.py
@@ -1,12 +1,25 @@
 #!/usr/bin/env python3
-"""Sphinx extension to auto-generate CLI command documentation from FABulousREPL."""
+"""Sphinx extension to auto-generate REPL command documentation from FABulousREPL.
 
-import ast
+Command metadata is read by importing the live REPL classes and introspecting the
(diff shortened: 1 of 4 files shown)


KelvinChung2000:
suppressed by #790

PR #896, opened 2026-07-03
fix(repl): rename --no-switch-matrix flag to --switch-matrix

KelvinChung2000:
Stacked PRs:
 * __->__#896
 * #895
 * #894
 * #893
 * #892

--- --- ---

### fix(repl): rename --no-switch-matrix flag to --switch-matrix

A boolean Option literally named "--no-switch-matrix" is inferred as an
argparse BooleanOptionalAction, which Python 3.13+ rejects because it
reserves the --no- prefix for the auto-generated negation form (the parser
fails to build when the command is invoked). Flip generate_custom_tile_config
to the positive --switch-matrix flag, enabled by default. argparse still
generates --no-switch-matrix as the negation, so that flag keeps working and
keeps its "skip the switch matrix" meaning, while the parser now builds on
Python 3.13+.


KelvinChung2000:
Folded into #893. `generate_custom_tile_config` now introduces the positive `--switch-matrix` flag directly in the migration commit, so the intermediate `--no-switch-matrix` state (which argparse rejects on Python 3.13+) never exists in the history.

PR #911, opened 2026-07-06
chore: remove custom pdn step and scripts

IAmMarcelJung:
LibreLane now supports all functionality we need for our PDN setup by default, so we can remove our custom PDN scripts and the custom PDN step that was using them.


KelvinChung2000:
You likely forgot to remove the related test as well

IAmMarcelJung:
Jup thanks!

KelvinChung2000 (approved):

PR #926, opened 2026-07-13
refactor: make SwitchMatrix class a consistent state and avoid re-reading the config files on every step

EverythingElseWasAlreadyTaken:
We turned SwitchMatrix class from a thin pointer-to-a-file into object that owns its parsed connectivity.
The matrix files (.csv,  .list, or hand-written HDL) is parsed once into canonical order at fabric-parse time; tiles hold the object and consumers read connections directly.
This replaces the old bootstrap round-trip (.list → write CSV → re-parse) and no *_switch_matrix.csv is written by default anymore.
List2CSV and CSV2List are now CLI functions, that can be used for debugging.

Extended the testing and also add tests for the preserve list order function.


fabulous/fabric_definition/supertile.py
```diff
@@ -53,8 +52,7 @@ class SuperTile:
     tileMap: list[list[Tile]]
     bels: list[Bel] = field(default_factory=list)
     withUserCLK: bool = False
-    supertile_matrix_dir: Path | None = None
-    supertile_matrix_config_bits: int = 0
+    switchMatrix: SwitchMatrix | None = None
```
KelvinChung2000: move to snake case

We can rename the rest of the field in the upcoming supertile refactor

fabulous/fabric_definition/tile.py
```diff
@@ -73,8 +70,7 @@ class Tile:
     name: str
     portsInfo: list[Port]
     bels: list[Bel]
-    matrixDir: Path
-    matrixConfigBits: int
+    switchMatrix: SwitchMatrix
```
KelvinChung2000: same in here

fabulous/fabric_generator/parser/parse_csv.py
```diff
@@ -137,13 +134,18 @@ def parsePortLine(line: str) -> tuple[list[Port], tuple[str, str] | None]:
     raise InvalidPortType(f"Unknown port type: {kind}")
 
 
-def parseTilesCSV(fileName: Path) -> tuple[list[Tile], list[tuple[str, str]]]:
+def parseTilesCSV(
```
KelvinChung2000, line 137: should we rename, since we are touching it?

fabulous/fabulous_api.py
KelvinChung2000, line 1: https://github.com/FPGA-Research/FABulous/pull/862

I have done something similar. But your implementation makes more sense; making it a class method is much cleaner. I will adapt mine to yours later, once the cmd2 4.0 migration is done.

fabulous/fabric_definition/switch_matrix.py
```diff
@@ -2,70 +2,388 @@
...
+                if mux_in not in dest_set:
+                    raise InvalidSwitchMatrixDefinition(
+                        f"Switch matrix input {mux_in!r} (driving {mux_out!r}) in "
+                        f"{filename} is not a signal of the tile"
+                    )
+
+    def to_csv_file(self, path: Path, tile_name: str) -> None:
+        """Write the switch matrix connections to a `.csv` file.
+
+        The file is written in the format consumed by :func:`parseMatrix`:
```
KelvinChung2000: Just some leftover style to clean up. Maybe we should find something to detect this so that we can make the agent work in loops to confirm everything is clean

fabulous/fabric_definition/switch_matrix.py
```diff
@@ -2,70 +2,388 @@
...
+            inputs needs none).
+        """
+        total = 0
+        for sources in connections.values():
+            if len(sources) >= 2:
+                total += (len(sources) - 1).bit_length()
+        return total
+
+    @staticmethod
+    def _extract_config_bits_from_hdl(path: Path) -> int:
```
KelvinChung2000, line 412: Do we still need this? We can get it from yosys_obj?

fabulous/fabric_definition/switch_matrix.py
```diff
@@ -2,70 +2,388 @@
...
+from fabulous.fabric_generator.parser.parse_switchmatrix import parseList, parseMatrix
 
-@dataclass
+if TYPE_CHECKING:
+    from fabulous.fabric_definition.bel import Bel
+    from fabulous.fabric_definition.port import Port
+
+
+def switch_matrix_signal_order(
+    ports: "list[Port]", bels: "list[Bel]"
```
KelvinChung2000:
```suggestion
    ports: list[Port], bels: list[Bel]
```

fabulous/fabric_definition/switch_matrix.py
```diff
@@ -2,70 +2,388 @@
...
-                for sink in sinks
-            ]
-        if suffix == ".list":
-            return parseList(self.matrix_dir)
-        raise InvalidFileType(f"File {self.matrix_dir} is not a .csv or .list file")
+    @classmethod
+    def _canonical_list_connections(
+        cls,
+        path: Path,
+        ports: "list[Port]",
```
KelvinChung2000:
```suggestion
        ports: list[Port],
```

fabulous/fabric_definition/switch_matrix.py
```diff
@@ -2,70 +2,388 @@
...
+        bels: "list[Bel]",
```
KelvinChung2000:
```suggestion
        bels: list[Bel],
```

fabulous/fabric_definition/switch_matrix.py
```diff
@@ -2,70 +2,388 @@
...
+        rather than the canonical dest-column order. Recorded once at read time
+        and reused when exporting so a round trip is faithful. Default False.
     """
 
-    name: str
-    matrix_dir: Path
-    config_bits: int = 0
+    matrix_file: Path
+    connections: dict[str, list[str]]
+    no_config_bits: int
```
KelvinChung2000: Make this into a property and compute on demand; it might be better, since this allows us to have a fresh number every time after maybe programmatically modifying the `connections`.

fabulous/fabulous_cli/fabulous_cli.py
```diff
@@ -849,6 +862,28 @@ def do_gen_switch_matrix(self, args: argparse.Namespace) -> None:
...
+        logger.info(
+            "Format conversion only; connectivity is not validated against any "
+            "tile configuration."
+        )
+        self.fabulousAPI.addList2Matrix(args.input, args.output)
+        logger.info(f"Converted {args.input} to {args.output}")
+
+    @with_category(CMD_SETUP)
+    @with_argparser(switch_matrix_convert_parser)
+    def do_csv2list(self, args: argparse.Namespace) -> None:
```
KelvinChung2000:
```suggestion
    def do_csv_to_list(self, args: argparse.Namespace) -> None:
```

I prefer this over the "2"

fabulous/fabulous_cli/fabulous_cli.py
```diff
@@ -849,6 +862,28 @@ def do_gen_switch_matrix(self, args: argparse.Namespace) -> None:
             self.fabulousAPI.genSwitchMatrix(i)
         logger.info("Switch matrix generation complete")
 
+    @with_category(CMD_SETUP)
+    @with_argparser(switch_matrix_convert_parser)
+    def do_list2csv(self, args: argparse.Namespace) -> None:
```
KelvinChung2000:
```suggestion
    def do_list_to_csv(self, args: argparse.Namespace) -> None:
```

fabulous/fabric_definition/supertile.py
```diff
@@ -247,6 +245,16 @@ def total_config_bits(self) -> int:
         """Return the supertile's config bits: switch matrix bits plus BEL bits."""
         return self.supertile_matrix_config_bits + sum(b.configBit for b in self.bels)
 
+    @property
+    def supertile_matrix_dir(self) -> Path | None:
+        """Return the supertile switch matrix file, or None if there is none."""
+        return None if self.switchMatrix is None else self.switchMatrix.matrix_file
+
+    @property
+    def supertile_matrix_config_bits(self) -> int:
```
KelvinChung2000, line 254: Is this now a single SM, or should this return the sum of all subtiles?

KelvinChung2000 (changes requested):

fabulous/fabric_definition/supertile.py
EverythingElseWasAlreadyTaken: This is a single switch matrix, that sits in the supertile.

EverythingElseWasAlreadyTaken:
Btw. the tests currently  fail, because we don't generate a switchmatrix csv by default, so the reference test fails.
I have to update the reference tests to fix this.

fabulous/fabulous_api.py
```diff
@@ -145,41 +142,29 @@ def loadFabric(self, fabric_dir: Path) -> None:
...
-        listFile : Path
-            List data to be converted.
         matrix : Path
-            File path where the matrix data will be saved.
+            CSV matrix data to be converted.
+        listFile : Path
+            Destination `.list` file (created or overwritten).
         """
-        list2CSV(listFile, matrix)
+        SwitchMatrix.from_file(matrix, matrix.stem).to_list_file(listFile)
```
KelvinChung2000: **`csv2list` fails on every real project switch-matrix CSV.**

`SwitchMatrix.from_file(matrix, matrix.stem)` passes the *file stem* where `parseMatrix` expects the *tile name*. For a real project file those differ: the stem is `W_IO_switch_matrix`, but the CSV's top-left header cell is `W_IO`.

Reproduced against the demo project:

```python
>>> SwitchMatrix.from_file(Path("demo/Tile/W_IO/W_IO_switch_matrix.csv"), "W_IO_switch_matrix")
InvalidSwitchMatrixDefinition: Tile name (top left element) in csv file does not
match tile name in tile object
```

`test_switch_matrix_list_csv_conversion` doesn't catch this because it only re-reads a CSV that `list2csv` itself wrote, and `list2csv` writes the stem into that header cell (same root cause — see the comment on line 155). The tile name needs to come from the CSV header itself, or be a separate argument to the command.

<sub>🤖 Posted by Claude (Opus 4.8) on behalf of @KelvinChung2000 — automated code review, please verify before acting.</sub>

KelvinChung2000 (commented):
Nice cleanup — collapsing the `.list` → bootstrap CSV → re-parse round trip into a single read is the right direction, and `_canonical_list_connections` reproduces the old ordering faithfully.

I left inline comments on the things I think need addressing before merge. Two findings don't land on diffed lines, so they're here instead:

**1. `SwitchMatrix` is now deep-copied once per fabric-grid cell.**
`parse_csv.py:577` and `parse_csv.py:809` both `deepcopy(tileDic[...])` for every placement. Previously `Tile.matrixDir` was a bare `Path`; `Tile.switchMatrix` now carries a dict of hundreds of `str -> list[str]` entries, so an N×M fabric allocates N×M copies of the full connectivity. Since `SwitchMatrix` is declared `frozen=True` and documented as safe to share, a `__deepcopy__` returning `self` would keep the sharing free.

**2. `Fabric.preserveListOrder` (`fabric.py:114`) is now dead state.**
`genSwitchMatrix` no longer forwards it (the flag is consumed at parse time), and grep finds no remaining reader — it's set and never used, so a future caller will reasonably assume it still controls generation. Same class of leftover: `configBit = 0` at `parse_csv.py:201` is never read anymore, and the `elif matrix_file.is_dir()` branch in `gen_bitstream_spec.py:76` is unreachable now that `SwitchMatrix.from_file` only accepts suffixed files.

<sub>🤖 Posted by Claude (Opus 4.8) on behalf of @KelvinChung2000 — automated code review, please verify before acting.</sub>

fabulous/fabulous_api.py
```diff
@@ -145,41 +142,29 @@ def loadFabric(self, fabric_dir: Path) -> None:
...
+            Destination `.csv` file (created or overwritten).
         """
-        tile = self.fabric.getTileByName(tileName)
-        if not tile:
-            raise ValueError(f"Tile {tileName} not found in fabric.")
-        bootstrapSwitchMatrix(tile, outputDir)
-
-    def addList2Matrix(self, listFile: Path, matrix: Path) -> None:
-        """Convert list into CSV matrix and save it.
+        SwitchMatrix.from_file(listFile, listFile.stem).to_csv_file(matrix, matrix.stem)
```
KelvinChung2000: **`list2csv` writes the *output file's* stem into the CSV's tile-name cell, so the CSV it produces can't be used as a tile `MATRIX`.**

`to_csv_file(matrix, matrix.stem)` — reproduced on the demo tile:

```
$ list2csv demo/Tile/W_IO/W_IO_switch_matrix.list W_IO_switch_matrix.csv
$ head -c 40 W_IO_switch_matrix.csv
W_IO_switch_matrix,W2MID0,W2MID1,W2MID...
```

The header cell should be `W_IO`. Pointing `MATRIX,./W_IO_switch_matrix.csv` at that file makes `parseMatrix` reject it. Combined with the `csv2list` issue on line 167, neither direction round-trips against real project files — which defeats the stated debugging purpose of the two commands.

<sub>🤖 Posted by Claude (Opus 4.8) on behalf of @KelvinChung2000 — automated code review, please verify before acting.</sub>

fabulous/fabric_definition/switch_matrix.py
```diff
@@ -2,70 +2,388 @@
...
+        Parameters
+        ----------
+        path : Path
+            Destination `.csv` file. Created (or overwritten) by this call.
+        tile_name : str
+            Tile name written to the top-left cell of the CSV header.
+        """
+        preserve_list_order = self.preserve_list_order
+        # Column headers = unique mux-input signals, in first-seen order.
+        mux_inputs_ordered: list[str] = []
```
KelvinChung2000, line 330: **`to_csv_file` isn't order-faithful in the default (non-preserve) mode.**

Column headers are collected in first-seen order *across rows*, which is not the canonical dest order. `parseMatrix` resolves an all-`1` row by column index, so some rows come back reordered:

```python
>>> sm = SwitchMatrix(matrix_file=Path("x.csv"),
...                   connections={"A": ["Y", "Z"], "B": ["X", "Y"]},
...                   no_config_bits=0)
>>> sm.to_csv_file(out, "T")
>>> out.read_text()
'T,Y,Z,X\nA,1,1,0,#,2\nB,1,0,1,#,2\n#,2,1,1'
>>> parseMatrix(out, "T")
{'A': ['Y', 'Z'], 'B': ['Y', 'X']}   # B was ['X', 'Y']
```

Mux-input position determines each PIP's config-bit encoding, so a `.list` → CSV export silently yields a matrix with a *different bitstream mapping*. The removed `list2CSV` avoided this by filling a `bootstrapSwitchMatrix` CSV whose columns were already in canonical dest order — that invariant is gone here.

Related: `to_list_file`'s docstring says the non-preserve order is fine because "the reader re-derives it from the tile's ports", but that's only true of the `.list` reader; `parseMatrix` does not re-derive anything from ports.

<sub>🤖 Posted by Claude (Opus 4.8) on behalf of @KelvinChung2000 — automated code review, please verify before acting.</sub>

fabulous/fabric_cad/gen_npnr_model.py
```diff
@@ -65,31 +62,15 @@ def genNextpnrModel(
...
-                        delay: float = 8
-                        if delay_model is not None:
-                            delay = delay_model.pip_delay(tile.name, sink, source)
-                        pipStr.append(
-                            f"X{x}Y{y},{sink},X{x}Y{y},{source},{delay},{sink}.{source}"
-                        )
-            elif tile.matrixDir.suffix == ".list":
-                connection = parseList(tile.matrixDir)
-                for source, sink in connection:
+            for source, sinkList in tile.switchMatrix.connections.items():
```
KelvinChung2000: **A hand-written-HDL switch matrix now silently contributes zero tile-internal pips instead of raising.**

`SwitchMatrix.from_file` returns `connections={}` for `.v`/`.sv`/`.vhdl`/`.vhd`, so for such a tile this loop emits the `#Tile-internal pips` header and nothing else. The same is true of `generateBitstreamSpec` (`gen_bitstream_spec.py:151`), which now produces no switch-matrix bit mapping — while `no_config_bits` read from the HDL still reserves frame bits.

Before this PR that path raised:

```python
raise InvalidFileType(f"File {tile.matrixDir} is not a .csv or .list file")
```

So we've gone from a loud failure to a quiet wrong result: nextpnr can't route through the tile, and the bitstream leaves those config bits unmapped, with only a parse-time warning to go on. AGENTS.md is explicit here — *"No fallbacks, no silent retries. Surface failures with clear errors. A stack trace beats a wrong result."* Either raise when a consumer that needs connectivity meets an HDL matrix, or make the HDL escape hatch carry real connectivity.

<sub>🤖 Posted by Claude (Opus 4.8) on behalf of @KelvinChung2000 — automated code review, please verify before acting.</sub>

fabulous/fabric_generator/parser/parse_csv.py
```diff
@@ -451,21 +425,22 @@ def parseTilesCSV(fileName: Path) -> tuple[list[Tile], list[tuple[str, str]]]:
...
-                    configBit += (muxSize - 1).bit_length()
 
         new_tiles.append(
             Tile(
                 name=tileName,
                 ports=ports,
                 bels=bels,
                 tileDir=fileName,
-                matrixDir=matrixDir,
+                switchMatrix=SwitchMatrix.from_file(
```
KelvinChung2000: **A tile CSV with no `MATRIX` line now crashes with `AttributeError: 'NoneType' object has no attribute 'suffix'`.**

`matrixDir: Path | None = None` (line 199) stays `None` when a tile declares no `MATRIX` row, and `SwitchMatrix.from_file` opens with `match path.suffix`. Previously such a tile parsed fine and only failed later *if* its switch matrix was actually generated; now every fabric load fails, with an unhelpful traceback.

Either reject it at the boundary with a clear `InvalidTileDefinition("tile X has no MATRIX line")`, or keep matrix-less tiles parseable — but not an `AttributeError`.

<sub>🤖 Posted by Claude (Opus 4.8) on behalf of @KelvinChung2000 — automated code review, please verify before acting.</sub>

fabulous/fabric_definition/switch_matrix.py
```diff
@@ -2,70 +2,388 @@
...
 from dataclasses import dataclass
 from pathlib import Path
+from typing import TYPE_CHECKING
 
-from fabulous.custom_exception import InvalidFileType
+from loguru import logger
 
+from fabulous.custom_exception import InvalidFileType, InvalidSwitchMatrixDefinition
+from fabulous.fabric_definition.define import Direction
+from fabulous.fabric_generator.parser.parse_switchmatrix import parseList, parseMatrix
```
KelvinChung2000: **This module-level import reintroduces the `fabric_definition` → `fabric_generator` dependency the deleted code deliberately avoided.**

The `SwitchMatrix.pips()` this PR removes used a function-local import, with an explicit comment saying why:

> Function-local import keeps `fabric_definition` free of a module-level dependency on `fabric_generator` (the same pattern `Tile` uses for its GDS import); the parser itself only depends on `custom_exception`.

`fabulous/fabric_generator/__init__.py` isn't empty, so importing *any* `fabric_definition` module now pulls in the generator package — exactly the circular-import hazard that guard existed to prevent. Moving the import inside `from_file` keeps the layering and costs nothing.

<sub>🤖 Posted by Claude (Opus 4.8) on behalf of @KelvinChung2000 — automated code review, please verify before acting.</sub>

fabulous/fabric_generator/parser/parse_csv.py
```diff
@@ -732,8 +700,16 @@ def parseFabricCSV(fileName: str) -> Fabric:
     superTileDic = {}
     unusedSuperTileDic = {}
 
+    # PreserveListOrder controls the canonical .list mux-input ordering, so it
+    # must be known before any tile is parsed (a tile may precede it in the CSV).
+    preserveListOrder = False
+    for line in parameters:
+        fields = [f.strip() for f in line.split(",") if f.strip()]
+        if fields and fields[0].startswith("PreserveListOrder"):
```
KelvinChung2000, line 713: **`PreserveListOrder` is now parsed twice, and this pre-scan raises `IndexError` on a valueless line.**

This pre-scan sets `preserveListOrder`, then the main parameter loop re-parses the same key at lines 793-794 into the same variable — that second read is dead.

Separately, a `ParametersBegin` block containing a bare `PreserveListOrder` (or `PreserveListOrder,`) makes `fields[1]` raise `IndexError` rather than the `InvalidFabricParameter` the other parameters raise. Worth folding both call sites into one small helper that validates the value.

<sub>🤖 Posted by Claude (Opus 4.8) on behalf of @KelvinChung2000 — automated code review, please verify before acting.</sub>

fabulous/fabric_definition/switch_matrix.py
KelvinChung2000: This might need a follow-up check, since I assume this will break the Bitstream spec?

fabulous/fabulous_api.py
EverythingElseWasAlreadyTaken: Having Parse Martrix check for the tile name and parse list not, makes no real sense for me. 
I would rather kick out the check.

fabulous/fabulous_api.py
EverythingElseWasAlreadyTaken: same here

fabulous/fabric_definition/switch_matrix.py
EverythingElseWasAlreadyTaken: This is not true, the _canonical_list_connections function serve exactly the purpose to recrate the old csv bootsrapping and the new preserve list order way. 
Also added tests to ensure legacy compatibility.

fabulous/fabulous_api.py
KelvinChung2000: We can update the behaviour. We should update the docs as well

> The adjacency matrix states the tile identifier name in the top left cell. The columns denote the input ports to the switch matrix and the rows denote the output ports. A non-zero entry in the matrix denotes a configurable connection, i.e. a multiplexer input connection. Each non-zero entry corresponds to one <output_port>,<input_port> tuple defined in the adjacency list.

Since this is in the doc. I guess it is there at first to prevent confusion when everything is flat?

fabulous/fabric_cad/gen_npnr_model.py
EverythingElseWasAlreadyTaken: should we kick out the HDL parsing or let it in? 
I've added a warning, if HDL is used.

fabulous/fabric_generator/parser/parse_csv.py
EverythingElseWasAlreadyTaken: I think a tile with no matrix makes no sense? 
So we should throw an error.

KelvinChung2000: More of a random edge case, but a tile with no switch matrix is possible, like a corner tile or a pass-through tile with only frame data and frame strobe. So maybe we should do a warning. 

But for this tile type, we probably should implement an auto-generated NULL tile for this use case.

fabulous/fabric_cad/gen_npnr_model.py
KelvinChung2000: We should keep this in, but I do think the use case is very limited. It will likely be used as a workaround when a specific switch matrix feature is desired but not yet supported.

EverythingElseWasAlreadyTaken:
> **1. `SwitchMatrix` is now deep-copied once per fabric-grid cell.** `parse_csv.py:577` and `parse_csv.py:809` both `deepcopy(tileDic[...])` for every placement. Previously `Tile.matrixDir` was a bare `Path`; `Tile.switchMatrix` now carries a dict of hundreds of `str -> list[str]` entries, so an N×M fabric allocates N×M copies of the full connectivity. Since `SwitchMatrix` is declared `frozen=True` and documented as safe to share, a `__deepcopy__` returning `self` would keep the sharing free.

This is intentional, since Tiles keep their wiring self-contained. 
Currently, this does not seem like an issue, but if RAM becomes a limiting factor, we could use pythons string interning.

fabulous/fabric_generator/parser/parse_csv.py
EverythingElseWasAlreadyTaken: The question is then, how the span wiring is treated in these cases. We have NULL tiles for corners, but there usually no wiring should arrive.  
But for everything else,  it needs special treatment. 
For example, will a quad wire only be routed through to the next tile? And how will the port and tile indexing then be? I guess, this needs some decisions and special treatments, but until this is implemented, we should throw an error.

fabulous/fabulous_api.py
EverythingElseWasAlreadyTaken: Sure, will update the docs.

fabulous/fabric_generator/parser/parse_csv.py
KelvinChung2000: The proper approach is likely to be to do a pass-through for all of them. Right now, in the current form, in the middle of the fabric type without SM, it will require the user to put a "NULL" terminal for this to work correctly. This will be like carving a hole in the fabric, while still allowing the frame data and frame strobe to pass. I don't know why someone would do this, but I guess raising an error might not be a bad idea.

fabulous/fabric_definition/switch_matrix.py
EverythingElseWasAlreadyTaken: As discussed, we leave it for now and added a warning, since it is a special case.

fabulous/fabulous_api.py
```diff
@@ -145,41 +142,45 @@ def loadFabric(self, fabric_dir: Path) -> None:
             logger.error("Only .csv files are supported for fabric loading")
             raise ValueError
 
-    def bootstrapSwitchMatrix(self, tileName: str, outputDir: Path) -> None:
-        """Bootstrap the switch matrix for the specified tile.
-
-        Using 'bootstrapSwitchMatrix' defined in 'fabric_gen.py'.
+    def addList2Matrix(
```
KelvinChung2000: Should we move this to `add_list_to_matrix` and similar below? We could clean this up later as we slowly removing the fabulous_api layer

KelvinChung2000 (approved):
Other than the one new comment everything looks good

EverythingElseWasAlreadyTaken:
I've updated the convention for preserve_list_order (default false).
It is now respected (when set) only while reading a switch matrix for both csv and list. 
So this defines the state (ordering of the connections) of the switchmatrix. 

Writing a switch matrix, so exporting a list, csv or generating RTL is just based on the state, preserve_list_order does not impact this. So if the internal state is edited for some reason, it will be always transparent to the user.

EverythingElseWasAlreadyTaken:
> Other than the one new comment everything looks good

fixed. Also snake cased a few more variables I touched.

KelvinChung2000 (approved):

PR #929, opened 2026-07-15
Add  BEL timing (bel.v3.txt) and placement_estimate.txt generation

EverythingElseWasAlreadyTaken:
genNextpnrModel now emits bel.v3.txt (bel.v2.txt plus per-arc Delay/SetupHold/ ClkToOut/Clock timing lines gated on cell params and port connectivity) and placement_estimate.txt (precomputed values for nextpnr's placement heuristic), alongside the existing pip/bel outputs.

Add tests and document everything in the nextpnr models page.

The nextpnr PR will follow soon.


fabulous/fabric_cad/gen_npnr_model.py
```diff
@@ -12,16 +12,173 @@
...
+CARRY_PREDICT_DELAY = 0.5
+
+# Arbitrary placeholder pip delay used when no delay_model is supplied.
+DUMMY_PIP_DELAY = 8
+
+# Representative FABULOUS_LC timing arcs for nextpnr's placement estimate.
+# Static while every LC instance shares these constants (I0-I3 LUT4); a real
+# per-instance timing model would regenerate this.
+_LC_LUT_INPUTS = ("I0", "I1", "I2", "I3")
+LC_ESTIMATE_LINES: list[str] = [
```
KelvinChung2000: For things like this, I prefer to use Jinja these days. But this might end up being a full rewrite of this gen model...

EverythingElseWasAlreadyTaken:
Should fix the failing tests:
https://github.com/FPGA-Research/FABulous-demo-projects/pull/11

fabulous/fabulous_api.py
```diff
@@ -428,12 +436,12 @@ def genBitStreamSpec(self) -> dict:
         """
         return generateBitstreamSpec(self.fabric)
 
-    def genRoutingModel(self) -> tuple[str, str, str, str]:
+    def genRoutingModel(self) -> tuple[str, str, str, str, str]:
```
KelvinChung2000: Just the usual touch and rename

KelvinChung2000 (approved):
For the string, we now return all V1, V2 and V3. I am thinking should we just keep a single format and generate a single output. Since V3 is back-compat with V2 and V1 anyway.

fabulous/fabulous_cli/fabulous_cli.py
```diff
@@ -1187,9 +1143,17 @@ def do_gen_model_npnr(self, *_ignored: str) -> None:
...
+        logger.info(f"output file: {self.projectDir}/{META_DATA_DIR}/bel.v3.txt")
+        with Path(f"{self.projectDir}/{META_DATA_DIR}/bel.v3.txt").open("w") as f:
+            f.write(npnrModel[3])
+
         logger.info(f"output file: {self.projectDir}/{META_DATA_DIR}/template.pcf")
         with Path(f"{self.projectDir}/{META_DATA_DIR}/template.pcf").open("w") as f:
-            f.write(npnrModel[3])
+            f.write(npnrModel[4])
+
+        estimatePath = Path(f"{self.projectDir}/{META_DATA_DIR}/placement_estimate.txt")
```
KelvinChung2000, line 1199: Maybe directly to the user design folder? Or in the nextpnr side will have the auto reading?

EverythingElseWasAlreadyTaken:
> For the string, we now return all V1, V2 and V3. I am thinking should we just keep a single format and generate a single output. Since V3 is back-compat with V2 and V1 anyway.

I also thought about this, but decided against, for the reason of nexpnr versioning. 
If we only export v3 format and name it like v2, older nextpnr versions will throw errors on the format, since the old parser does not know the new fields. 

The only one we can kick out safely, is the v1, since this is so old, no one will use any more, and it won't even work with the newer project structure with the .FABulous folder.

fabulous/fabric_cad/gen_npnr_model.py
EverythingElseWasAlreadyTaken: Makes sense, will think about it.

fabulous/fabulous_cli/fabulous_cli.py
EverythingElseWasAlreadyTaken: Its auto read by nextpnr, but this could be changed, but I would not place it in user design, since it belongs to the model files, so .FABulous makes more sense in my opinion.

fabulous/fabric_cad/gen_npnr_model.py
EverythingElseWasAlreadyTaken: I've tried templating with jinja here: 
https://github.com/EverythingElseWasAlreadyTaken/FABulous/tree/feat/bel_v3_with_jinja

But for me, it seems like it does not solve any problems.
The code does not look much cleaner, adds more loc and the execution time for run_fab on my laptop jumped from 2 sec, to 3.5 sec. for our demo fabric.

KelvinChung2000 (approved):

PR #935, opened 2026-07-17
feat(fabric_cad): add per-type IO pad placement estimates

EverythingElseWasAlreadyTaken:
The initial bel v3. only added estimates for the FABULOUS_LC, this now allows to add estimates for all BEL types.

Registered pads (InPass4/OutPass4_frame_config[_mux]) had no placement-estimate arcs, so nextpnr fell back to zero-delay estimates for pad-facing paths. Emit one BelBegin/BelEnd block per timed BEL type using nextpnr's historical default constants; P&R behaviour unchanged for FABULOUS_LC.

### DIFF
--- a/fabulous/fabric_cad/gen_npnr_model.py
+++ b/fabulous/fabric_cad/gen_npnr_model.py
@@ -20,7 +20,7 @@ from fabulous.fabric_definition.bel import Bel
 from fabulous.fabric_definition.fabric import Fabric
 
 # Dummy BEL timing values (ns), mirroring nextpnr's historical hardcoded
-# constants (fabulous.cc, update_cell_timing).
+# constants (fabulous.cc, seed_default_estimates).
 LUT_DELAY = 3.0
 CARRY_CICO_DELAY = 0.2
 CARRY_I_DELAY = 1.0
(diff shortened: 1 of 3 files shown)


fabulous/fabric_cad/gen_npnr_model.py
```diff
@@ -71,7 +73,21 @@
             f"delayEpsilon={DELAY_EPSILON}",
             f"ripupPenalty={RIPUP_PENALTY}",
             f"carryPredictDelay={CARRY_PREDICT_DELAY}",
+            "BelBegin,FABULOUS_LC",
             *LC_ESTIMATE_LINES,
+            "BelEnd",
+            "BelBegin,OutPass4_frame_config",
```
KelvinChung2000, line 79: This is hard-coded at the moment, which is fine for now. But down the line, I would want to do something like:

```verilog
(* bel_type="IO" *)
module OutPass4_frame_config(...)
...
```
So we can add settings based on the functions of the Bel. We will likely want more like "arith" and "arbitrary" for things like DSP and LUT. I will make sure something like this is supported in the upcoming representation upgrade.

KelvinChung2000 (approved):

PR #950, opened 2026-07-23
fix(gen): emit VHDL component declarations for supertile HDL

EverythingElseWasAlreadyTaken:
VHDL projects with a supertile switch matrix failed to generate/compile because supertile-related entities were instantiated without the matching VHDL `component` declarations (Verilog is unaffected as it needs none).

- gen_fabric: declare tile/supertile components from each tile's real tileDir instead of rebuilding a flat Tile/<name>/<name>.vhdl path via a split("_") heuristic, which mis-resolved supertile subtiles.
- gen_tile (generateSuperTile): also declare the wrapper's own switch matrix, ConfigMem, and supertile-level BEL components.


KelvinChung2000:
This does not make sense to me, why is our CI not catching? I would assume this is a syntax failure.

EverythingElseWasAlreadyTaken:
> This does not make sense to me, why is our CI not catching? I would assume this is a syntax failure.

It only hit in case of switch matricies in supertile top wrapper, so this is something I overlooked for VHDL.
I've added a test to catch this now.

KelvinChung2000 (approved):

PR #951, opened 2026-07-23
chore(repl): cleanup pass on the REPL refactor

KelvinChung2000:


fabulous/fabulous_repl/fabulous_repl.py
```diff
@@ -295,26 +301,39 @@ def onecmd(
...
+    @with_annotated
+    def do_exit(self) -> bool:
         """Exit the FABulous shell and log info message."""
         logger.info("Exiting FABulous shell")
         return True
 
-    def do_quit(self, *_ignored: str) -> None:
+    @with_category(CMD_OTHER)
+    @with_annotated
+    def do_quit(self) -> None:
```
EverythingElseWasAlreadyTaken: Should also just return True like do exit

fabulous/fabulous_repl/fabulous_repl.py
```diff
@@ -295,26 +301,39 @@ def onecmd(
...
         """Exit the FABulous shell and log info message."""
         self.onecmd_plus_hooks("exit")
 
-    def do_q(self, *_ignored: str) -> None:
+    @with_category(CMD_OTHER)
+    @with_annotated
+    def do_q(self) -> None:
```
EverythingElseWasAlreadyTaken: same here

fabulous/fabulous_repl/fabulous_repl.py
KelvinChung2000: But this mean I will need to dulplicate the log message. Let me check how I can do an alias

EverythingElseWasAlreadyTaken (approved):

PR #956, opened 2026-07-24
feat(fabric): extend Bel with the typed port model

KelvinChung2000:
Stacked PRs:
 * #961
 * #960
 * #959
 * #958
 * #957
 * __->__#956

--- --- ---


KelvinChung2000:
@greptile-apps

fabulous/fabric_cad/gen_design_top_wrapper.py
```diff
@@ -92,13 +92,16 @@ def generateUserDesignTopWrapper(
...
-                if not bel.externalInput and not bel.externalOutput:
+                if not bel.external_input and not bel.external_output:
                     logger.info(
                         f"Skipping bel {bel.name} in tile X{x}Y{y} since it has no "
                         f"external ports"
                     )
                     continue
-                if len(bel.inputs + bel.outputs) == 0:
+                if (
+                    len([p.name for p in bel.inputs] + [p.name for p in bel.outputs])
```
EverythingElseWasAlreadyTaken, line 102: use bel.input_names and output_names

fabulous/fabric_cad/gen_design_top_wrapper.py
```diff
@@ -109,10 +112,10 @@ def generateUserDesignTopWrapper(
                 if bel.name not in bel_count:
                     bel_count[bel.name] = 0
                     bel_inputs[bel.name] = [
-                        port.removeprefix(bel.prefix) for port in bel.inputs
+                        port.name.removeprefix(bel.prefix) for port in bel.inputs
```
EverythingElseWasAlreadyTaken, line 115: Rather than having to remove the prefix every time, BelPort should hold a prefix_name and a non_prefix_name or similar.

EverythingElseWasAlreadyTaken: and similar to bel.input_names, there should be also a bel.input_names_no_prefix.

fabulous/fabric_definition/bel.py
```diff
@@ -4,164 +4,886 @@
...
+        validation runs when a consumer first reads them rather than at construction.
+
+        Raises
+        ------
+        ValueError
+            If an IO BEL has more than one external input or output.
+        """
+        # Accessing filetype validates the source suffix.
+        _ = self.filetype
+        if self.bel_type == BelType.IO:
```
EverythingElseWasAlreadyTaken, line 202: Why is this restricted to one extreneral input and output  per bel?
Also beltype is never defined and has therefore the default BelType LOGIC?

fabulous/fabric_definition/bel.py
```diff
@@ -4,164 +4,886 @@
...
-    """Information about a single BEL.
+    """Immutable, Yosys-backed description of a single BEL.
 
-    The information is parsed from the directory of the BEL in the CSV definition file.
-    There are some things to be noted:
+    A `Bel` stores the top `YosysModule` it was parsed from plus the few pieces of
+    information that do not live in the module (the source path, the BEL prefix, and
+    placement knobs). Every port collection is derived from the module on first
+    access and cached in its own property. The BEL is validated at construction time
+    and never mutated afterwards.
```
EverythingElseWasAlreadyTaken, line 119: I don't fully get why we should do Lazy loading here, instead of just parsing once and having a object as before. We need all elements anyway and this mostly at once. 
The whole handling seems to complicated here.

fabulous/fabric_definition/bel.py
EverythingElseWasAlreadyTaken, line 116: These attributes are never set and always default values.
Also, param_overrie and base_bel need some explanation.

fabulous/fabric_definition/bel.py
```diff
@@ -4,164 +4,886 @@
...
+        bel = cls(
+            src=filename,
+            prefix=prefix,
+            module=module,
+            module_name=module_name,
+        )
+
+        config_bits_port = module.ports.get("ConfigBits")
+        no_config_bits = len(config_bits_port.bits) if config_bits_port else 0
+        if len(bel.bel_feature_map) != no_config_bits:
```
EverythingElseWasAlreadyTaken, line 255: this should be rather checked when the bel feature map is evaluated. 
There could be also a check useful, that every config bit is assigned to a feature

fabulous/fabric_definition/bel.py
```diff
@@ -4,164 +4,886 @@
...
+                    SharedPort(
+                        name=bit_name,
+                        io_direction=direction,
+                        width=1,
+                    )
+                )
+        return shared_port
+
+    @cached_property
+    def carry(self) -> dict[str, dict[IO, str]]:
```
EverythingElseWasAlreadyTaken, line 512: Since we now have a belport class, we could actually move the carry and local_shared properties to the port. But  this can be also a dedicated PR.

fabulous/fabric_definition/supertile.py
```diff
@@ -235,15 +235,15 @@ def get_matrix_port_names(self) -> tuple[set[str], set[str]]:
                         valid_sinks |= names
 
         for bel in self.bels:
-            valid_sinks.update(bel.inputs)
-            valid_sources.update(bel.outputs)
+            valid_sinks.update(p.name for p in bel.inputs)
```
EverythingElseWasAlreadyTaken, line 238: use input_names

fabulous/fabric_definition/supertile.py
```diff
@@ -235,15 +235,15 @@ def get_matrix_port_names(self) -> tuple[set[str], set[str]]:
...
+            valid_sources.update(p.name for p in bel.outputs)
```
EverythingElseWasAlreadyTaken, line 239: use output_names

fabulous/fabric_generator/gen_fabric/fabric_automation.py
```diff
@@ -199,11 +198,11 @@ def generateSwitchmatrixList(
     ) as dummy_file_path:
         CLBDummyFile = dummy_file_path
 
-    belIns = sum((bel.inputs for bel in bels), [])
-    belOuts = sum((bel.outputs for bel in bels), [])
+    belIns = sum(([p.name for p in bel.inputs] for bel in bels), [])
```
EverythingElseWasAlreadyTaken, line 201: use input/output_names

fabulous/geometry_generator/bel_geometry.py
```diff
@@ -83,13 +83,13 @@ def generateGeometry(self, bel: Bel, padding: int) -> None:
         """
         self.name = bel.name
         self.src = bel.src
-        self.internalInputs = bel.inputs
-        self.internalOutputs = bel.outputs
-        self.externalInputs = bel.externalInput
-        self.externalOutputs = bel.externalOutput
+        self.internalInputs = [p.name for p in bel.inputs]
```
EverythingElseWasAlreadyTaken, line 86: use input/output_names

EverythingElseWasAlreadyTaken (changes requested):
I think we should discuss the lazy loading of the bel attributes.
In my opinion it makes the bel object just more complex and its quite inefficient to reparse the yosys model for every attribute, compared to just build the data structures once at init.

KelvinChung2000:
I notice the same as I am doing a re-read of what I did. Since the class is frozen, I am going to move to using a cached property instead, which should do the job without losing anything.  I could make it do all the filtering and populate the property on init, but I think this will make the init function very large.

EverythingElseWasAlreadyTaken:
> I notice the same as I am doing a re-read of what I did. Since the class is frozen, I am going to move to using a cached property instead, which should do the job without losing anything. I could make it do all the filtering and populate the property on init, but I think this will make the init function very large.

I would prefer a large init function over a scattered init function, to have the logic in one place.

PR #969, opened 2026-07-27
fix: respect user die area when said so

### LINKED ISSUE #962: Predefined DIE_AREA not respected
My `no_opt` flow finished, but with DRC errors. I tried to use the `balance`mode, but I see it keeps trying to start from a `DIE_AREA` that is considerably smaller that the area of the SRAM macro itself:
```shell
[08:30:40] INFO     FABulous optimisation is set to balance, default die area is ignored.                                                                               tile_macro_flow.py:182
[08:30:40] INFO     Rounding DIE_AREA from (178.48, 278.80) to (178.48, 278.80) (pitch_x=0.46, pitch_y=0.68)                                                                     helper.py:109
[08:30:40] INFO     Starting a new run of the 'BRAM_32' flow with the tag 'RUN_2026-07-27_08-30-40'.                                                                               flow.py:654
[08:30:40] INFO     Starting…     
```
The area of the macro is `703.02 x 737.94`, so anything smaller will automatically fail. As it starts with a very small area, it never manages to reach a solution. I have tried using the `FABULOUS_IGNORE_DEFAULT_DIE_AREA` to see if I could make it start from the provided `DIE_AREA`, but it does not seem to have any effect. Is there an intended way of providing a starting area?

_Originally posted by @Dario-CP in https://github.com/FPGA-Research/FABulous/discussions/880#discussioncomment-17793645_

### LINKED ISSUE #963: Incorrect documentation of FABULOUS_IGNORE_DEFAULT_DIE_AREA
The `FABULOUS_IGNORE_DEFAULT_DIE_AREA`, in [this page](https://fabulous.readthedocs.io/en/latest/user_guide/building_doc/fabric_gds.html) of the docs it appears as `IGNORE_DEFAULT_DIE_AREA`, which is incorrect.

_Originally posted by @Dario-CP in https://github.com/FPGA-Research/FABulous/discussions/880#discussioncomment-17793645_

KelvinChung2000:
close #962 
close #963

### DIFF
--- a/docs/source/user_guide/building_doc/fabric_gds.md
+++ b/docs/source/user_guide/building_doc/fabric_gds.md
@@ -285,7 +285,7 @@ If no feasible solution can be found after all iterations, the flow will raise a
 - `FABULOUS_OPTIMISATION_WIDTH_STEP_COUNT`: Sites to increase width per iteration (default: 4)
 - `FABULOUS_OPTIMISATION_HEIGHT_STEP_COUNT`: Sites to increase height per iteration (default: 1)
 - `IGNORE_ANTENNA_VIOLATIONS`: If `true`, antenna violations won't trigger size increases
-- `IGNORE_DEFAULT_DIE_AREA`: If `true`, ignores provided die area and starts from instance area
+- `FABULOUS_IGNORE_DEFAULT_DIE_AREA`: If `true`, ignores provided die area and starts from instance area
 
 ## Output Structure
 
--- a/docs/source/user_guide/building_doc/fabric_gds.md
+++ b/docs/source/user_guide/building_doc/fabric_gds.md
@@ -280,12 +280,25 @@ The GDS flow includes an iterative optimisation process to find the minimum viab
 
 If no feasible solution can be found after all iterations, the flow will raise an error and stop the generation.
 
+### Providing a Starting Die Area
+
+The optimisation only ever grows the die, so the area it starts from matters. A `DIE_AREA` set in the tile's `gds_config.yaml` is used as that starting area in every optimisation mode:
+
+- `find_min_width` / `find_min_height` lock the axis that is not being minimised to the provided value.
+- `balance` / `large` grow both axes from the provided value.
+
+Both axes of the provided `DIE_AREA` must still clear the minimum needed to place the tile's IO pins, otherwise the flow raises an error. Set `FABULOUS_IGNORE_DEFAULT_DIE_AREA: true` to discard the provided value and let the flow size the tile from scratch.
+
+:::{note}
+Without a provided `DIE_AREA`, the automatic starting area is derived from the IO pin count and the stdcell area reported by synthesis. Neither accounts for a hard macro instantiated in the tile, so a tile containing a macro (for example an SRAM) starts far too small and the optimisation cannot grow enough to reach a solution. For those tiles, set a `DIE_AREA` that already fits the macro.
+:::
+
 ### Related Variables
 
 - `FABULOUS_OPTIMISATION_WIDTH_STEP_COUNT`: Sites to increase width per iteration (default: 4)
 - `FABULOUS_OPTIMISATION_HEIGHT_STEP_COUNT`: Sites to increase height per iteration (default: 1)
 - `IGNORE_ANTENNA_VIOLATIONS`: If `true`, antenna violations won't trigger size increases
-- `FABULOUS_IGNORE_DEFAULT_DIE_AREA`: If `true`, ignores provided die area and starts from instance area
+- `FABULOUS_IGNORE_DEFAULT_DIE_AREA`: If `true`, ignores the provided `DIE_AREA` and starts from the area calculated by the flow (default: `false`)
 
 ## Output Structure
 
--- a/fabulous/fabric_generator/gds_generator/flows/tile_macro_flow.py
+++ b/fabulous/fabric_generator/gds_generator/flows/tile_macro_flow.py
@@ -32,8 +32,8 @@ configs = Classic.config_vars + [
     Variable(
         "FABULOUS_IGNORE_DEFAULT_DIE_AREA",
         bool,
-        "When is true will ignore the provided die area and "
-        "use the default one instead.",
+        "When true, discards the provided DIE_AREA and sizes the tile from "
+        "scratch instead of using it as the optimisation starting area.",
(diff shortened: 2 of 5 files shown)


IAmMarcelJung (commented):
In general LGTM, I'm only unsure if we should use the default die area as a starting point by default, since that kind of invalidates the point of the optimizations. For the case that uncovered the issue, this is probably the right approach though. Is it maybe also somehow possible to check if there is a macro in the tile and if so, get it's dimensions and add that to the starting area?

@Dario-CP, can you try if the solution from this PR works for you?

KelvinChung2000:
This is largely why the code is broken in the first place: I always start with the assumption that optimisation with predefined values defeats the original purpose, and sometimes you want to provide a larger area at the start to speed up convergence. I always forget that the latter is a wanted case whenever I touch that logic.

IAmMarcelJung:
But isn't that what `FABULOUS_IGNORE_DIE_AREA` is for? I would say the default for the optimization is starting from the minimum possible from synth area/IO pin count and if someone, like in this case, wants a different starting point, they just set the variable to `false` and use some predefined die area as a starting point.

Dario-CP:
> In general LGTM, I'm only unsure if we should use the default die area as a starting point by default, since that kind of invalidates the point of the optimizations. For the case that uncovered the issue, this is probably the right approach though. Is it maybe also somehow possible to check if there is a macro in the tile and if so, get it's dimensions and add that to the starting area?
> 
> @Dario-CP, can you try if the solution from this PR works for you?

I will try it, thanks! Also, I assume this is not done by the flow, but for the cases where a hard macro is involved in an optimized flow it would be nice that the relative position of the macro was kept throughout the flow. Currently, I guess that even if the macro is originally centered, when the DIE_AREA increases, the macro location remains the same, which results in the macro shifting to the bottom left relative to he whole area. I don't know if the best approach would be to calculate the relative position at the beginning and then use that to recalculate the location at each iteration, or just have a boolean variable to automatically center the macro and ignore the given location, so even with a no_opt flow the macro would be automatically centered. I have not yet tried the PR (I am launching the flow in some minutes), but I assume this is how it would work.

KelvinChung2000:
Yes, the location will stay the same. Implementing relative positioning is probably the way to go for now until auto macro placement is implemented. As the area increases, the optimal position will likely change. 

I will implement this feature later. Likely with modes like `relative`,  `fix` and `auto` for the future.

Dario-CP:
> Yes, the location will stay the same. Implementing relative positioning is probably the way to go for now until auto macro placement is implemented. As the area increases, the optimal position will likely change.
> 
> I will implement this feature later. Likely with modes like `relative`, `fix` and `auto` for the future.

Sounds cool!
I have run the flow, and got these errors:
```shell
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[GRT-0229] Vertical edge usage exceeds the maximum allowed. (126, 58) usage=65535 limit=2200
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[GRT-0229] Vertical edge usage exceeds the maximum allowed. (35, 16) usage=65535 limit=2200
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[GRT-0228] Horizontal edge usage exceeds the maximum allowed. (132, 140) usage=65535 limit=3000
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[GRT-0228] Horizontal edge usage exceeds the maximum allowed. (17, 27) usage=65535 limit=3000
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[DRT-0416] Term Tile_X0Y0_E1END[0] contains offgrid pin shape. Pin shape ( 0 936884 ) ( 600 937484 ) is not a multiple of the manufacturing grid 5.
[GRT-0228] Horizontal edge usage exceeds the maximum allowed. (81, 14) usage=65535 limit=3000
[GRT-0228] Horizontal edge usage exceeds the maximum allowed. (125, 73) usage=65535 limit=3000
```
The offgrid pin error remains even after https://github.com/FPGA-Research/FABulous/pull/901
Regarding the starting area, it took the one I specified, so that works.

EverythingElseWasAlreadyTaken (approved):

PR #975, opened 2026-07-29
chore(nix): ship as composable

KelvinChung2000:
## What this does

FABulous ships as a Nix flake a downstream chip or fabric project consumes with one input.

`pkgs.fabulous-python` is an ordinary nixpkgs Python interpreter whose package set is FABulous's `uv.lock`. Every package the lock resolves is converted into that set, so the versions are uv's resolution rather than whatever nixpkgs happens to ship.

```nix
pkgs.fabulous-shell.override {
  extra-python-packages = ps: [ ps.cocotb ps.pytest ];
  extra-packages = with pkgs; [ iverilog gnumake ];
}
```

Because it is a normal nixpkgs Python set, it composes normally:

```nix
pkgs.fabulous-python.withPackages (ps: [ ps.fabulous-fpga ])
pkgs.fabulous-python.pkgs.fabulous-fpga            # as a propagatedBuildInput
```

Also exposed: `pkgs.fabulous` (the CLIs wrapped with the EDA toolchain on `PATH`), `fab-yosys`, `fab-nextpnr`, `fab-ghdl`, `fabulator`, and `nix flake init -t github:FPGA-Research/FABulous` to scaffold a consuming project.

## Numbers

Of the 111 packages in `uv.lock`, measured against nixpkgs' own Python set:

| | plain nixpkgs | this |
| --- | --- | --- |
| at `uv.lock` version | 27 | **106** |
| absent, would need hand-packaging | 20 | **0** |
| diverging from `uv.lock` | 63 | **3** |

Nothing is hand-written. The package list is read from `uv.lock`, so `uv lock` is the only thing that moves it.

## Why not a plain buildPythonPackage

nixpkgs ships 63 of these packages at older versions and lacks 20 entirely (`librelane`, `sdf-timing`, `ciel`, `fabulous-fasm`, and others). Going native there means a hand-maintained dependency list duplicating `pyproject.toml`, ~12 package definitions with their own hashes, and 63 version divergences re-checked on every nixpkgs bump. This keeps `uv.lock` as the single source of truth instead.

## Workarounds this relies on

1. **Convert the whole closure, not part of it.** `toNixpkgs` resolves each converted package's dependencies from the set it lands in, so a partial conversion silently falls back to nixpkgs' versions for everything it did not convert.

2. **Scope it to one interpreter.** Applied through `python3.override { packageOverrides = ...; self = python; }` rather than as a global overlay, so `pkgs.python3` is untouched for everything else in the consumer's nixpkgs.

3. **Restrict conversion to `uv.lock`'s own package names.** `toNixpkgs` probes every candidate with `isDerivation`, and `build-system-pkgs` contributes fixups for packages absent from this lock (`oldest-supported-numpy`) that fail when probed. Nix's `&&` short-circuits before the probe.

4. **Exclude `wheel`, `packaging`, `tomli`.** nixpkgs' `buildPythonPackage` is built from them, so converting them asks them to build themselves — infinite recursion.

5. **Reapply `postInstall` fixups on the nixpkgs side.** `toNixpkgs` reinstalls each package from its wheel, so anything a fixup did in `postInstall` is undone. The `about-time` / `alive-progress` `LICENSE` collision has to be fixed on both sides.

## Caveats

- `toNixpkgs` is marked **experimental** upstream and may change.
- `wheel`, `packaging` and `tomli` are nixpkgs' versions, not the lock's. All three are build tooling rather than anything FABulous imports, and nixpkgs' versions satisfy its constraints.
- **The Python environment cannot be built on a case-insensitive `/nix`.** FABulous installs both a `FABulous` and a `fabulous` command, which are one path there, and the installer refuses to overwrite. The Nix installer creates a case-sensitive store volume on macOS by default, so this normally does not arise.
- `librelane_plugin_fabulous` ships in the FABulous wheel from **2.2.0** onwards. On an older FABulous everything else works, but `librelane` will not discover the FABulous flows.
- Development shells stay on the editable uv2nix virtualenv. That is the one thing the converted set cannot express: it builds from wheels, and a wheel is not an editable checkout.
- A project needing a package `uv.lock` does not resolve, or a different version of one it pins, can build from its own `uv.lock` and pass it as `python-env` (`loadFabulousWorkspace` + `mkFabulousPythonSet`). That route gives up the nixpkgs composability above, so it is the exception rather than the default.

## What you maintain when `uv.lock` changes

| Change | Result |
| --- | --- |
| Version bumps (the common case) | Auto-syncs, no Nix edit |
| A package renamed or dropped | `checks.python-fixups` fails and names it |
| A new **sdist-only** package | Build fails; needs a build-system entry in `nix/overlay/python.nix` |
| A new package installing a colliding top-level file | `buildEnv` fails; needs a `nixpkgsFixups` entry |

Checked by running `uv lock --upgrade` (23 packages bumped): everything evaluated and built with no Nix change.

## Fixed along the way

- The `fasm` fixup had been dead since #672 renamed that package to `fabulous-fasm`. Nix only forces an override when something depends on the package, so it stopped applying without ever failing. `checks.python-fixups` now asserts every patched name is still one `uv.lock` resolves.
- `sdf-timing` stamped a hardcoded `SETUPTOOLS_SCM_PRETEND_VERSION` of `0.0.post134` while the lock said `0.0.post136`. It now takes the resolved version.
- The dev shell ran FABulous's `librelane_plugin_fabulous` generation whenever `editable-root` was set, which for a downstream checkout meant importing a `build_hooks` that is not there. Now gated behind a separate `materialize-plugin`.

## Breaking changes

- `fabulous-shell`'s `fabulous-venv` argument is now `python-env`.
- `extra-python-packages` selects from `fabulous-python` rather than nixpkgs.
- `fabulous.passthru.tkinterPath` and `.venv` are gone; the environment carries tkinter directly and is exposed as `.pythonEnv`.


KelvinChung2000:
@mole99 Have you tested if this is working?

mole99:
Ah, no! I totally forgot after HeiChips... feel free to ping me when there's stuff going stale.

I still need to do some catch-up on things after ORConf, but I'd like to take a look next week.

PR #976, opened 2026-08-01
fix: Added a python function to generate ConfigMem in FlipFlopChain configuration mode 

MrAbhi19:
## Summary of Changes
1. Added a new primitive **config_dff** in `models_pack.v` and `models_pack.vhdl`
2. Updated `generateConfigMem` function to support generating different `_Config_Mem` RTL implementation for `FlipFLopChain`
3. Reused the existing FRAME_BASED custom CSV logic to map and sequence the configuration bits for the new daisy chain (FLIPFLOP_CHAIN) architecture
4. renamed `generateConfigMem` universally to snake_case format
---
## Note:

- This change is strictly limited to the generation of the `config_mem` files. The corresponding tile and switch matrix updates are included in #994


fabulous/fabric_generator/gen_fabric/gen_configmem.py
```diff
@@ -102,7 +102,7 @@ def generateConfigMemInit(
             writer.writerow([entry[field] for field in fieldName])
 
 
-def generateConfigMem(
+def generateConfigMemFramebased(
```
KelvinChung2000:
```suggestion
def generate_config_mem_frame_based(
```

Since touching, we can migrate to snake_case

fabulous/fabric_generator/gen_fabric/gen_configmem.py
```diff
@@ -282,6 +282,89 @@ def generateConfigMem(
     writer.writeToFile()
 
 
+def generateConfigMemFF(
```
KelvinChung2000:
```suggestion
def generate_config_mem_FF(
```

similar here

fabulous/fabric_generator/gen_fabric/gen_configmem.py
KelvinChung2000, line 193: We will also want to include some of these checks in your new function. You can have a think what would be the best way to structure it such that enable more reuse

fabulous/fabric_generator/gen_fabric/gen_configmem.py
```diff
@@ -509,11 +593,21 @@ def generate_super_tile_config_mem(
...
         Number of frames per column.
+    config_bit_mode : ConfigBitMode
+        Memory Config Mode
     """
     st_config_bits = superTile.total_config_bits
     if st_config_bits <= 0:
         return
 
+    if config_bit_mode == ConfigBitMode.FLIPFLOP_CHAIN:
+        generateConfigMemFF(
```
KelvinChung2000: Similarly, calling `super_tile_config_mem_csv` will also be wanted.

KelvinChung2000 (changes requested):
The VHDL side will also need to be included and possibly need some checking.

fabulous/fabric_generator/gen_fabric/gen_configmem.py
```diff
@@ -216,20 +225,35 @@ def generateConfigMem(
...
-    if max_frame_per_col != 0:
-        writer.addParameter(
-            "MaxFramesPerCol", "integer", max_frame_per_col, indentLevel=2
-        )
-    if frame_bits_per_row != 0:
-        writer.addParameter(
-            "FrameBitsPerRow", "integer", frame_bits_per_row, indentLevel=2
-        )
+
+    # Frmae-based specific parameters
```
IAmMarcelJung:
```suggestion
    # Frame-based specific parameters
```

KelvinChung2000 (approved):

KelvinChung2000:
This is looking good now. I will do a few more checks; you can work on the next things based on this branch

KelvinChung2000 (approved):

IAmMarcelJung:
Thanks for the PR, could you please update the description to match it's current state? Thanks!

MrAbhi19:
I have updated the description, let me know if any changes are needed.

tests/fabric_gen_test/conftest.py
```diff
@@ -162,6 +164,9 @@ def fabric_config(request: pytest.FixtureRequest, mocker: MockerFixture) -> Fabr
     fabric.frameBitsPerRow = config.frame_bits_per_row
     fabric.maxFramesPerCol = config.max_frames_per_col
     fabric.name = config.name
+    # Explicitly set FRAME_BASED mode so cocotb RTL generation
+    # includes FrameData/FrameStrobe
+    fabric.configBitMode = getattr(config, "config_bit_mode", ConfigBitMode.FRAME_BASED)
```
IAmMarcelJung, line 169: I think this will always default to `ConfigBitmode.FRAME_BASED` since the `FabricConfig` class does not have a `config_bit_mode` attribute?

IAmMarcelJung (commented):
Thanks for updating the description! A few comments:

1. You could also add the `config_dff` module to the `MODELS_PACK_REQUIRED_MODULES` in `fabulous_settings.py` now that this module is added to the template.
2. You could add tests for the new code added for the flipflop chain mode in `configmem_test`.

I might have some more comments, but I wanted to get these out so you can already think about it.

tests/fabric_gen_test/conftest.py
MrAbhi19: yeah, and it's because we don't have any proper testing for `FlipFLopChain` mode and some simulation test is explicitly looking for `FrameData` and `FrameStrobe` even in the `FlipFlopChain` mode so to not break the testing flow. I had to hardcode it to `FRAME_BASED`

- I have added `config_dff` module to `MODELS_PACK_REQURIED_MODULES`

>You could add tests for the new code added for the flipflop chain mode in `configmem_test`.

I'm not familiar with the cocotb, so I might not be able to add tests. I'm fixing the `tile` and `switch_matrix` generation right now.

KelvinChung2000 (approved):

fabulous/fabric_files/FABulous_project_template_verilog/Fabric/models_pack.v
```diff
@@ -17,6 +17,20 @@ module config_latch (
     end
     /* verilator lint_on LATCH */
 endmodule
+
+module config_dff (
+    input D,
+    input CLK,
```
EverythingElseWasAlreadyTaken:
```suggestion
    input wire D,
    input wire CLK,
```

The rest of the modules will be fixed with #981.

fabulous/fabric_generator/gen_fabric/gen_configmem.py
```diff
@@ -509,24 +560,31 @@ def generate_super_tile_config_mem(
...
-        max_frames_per_col=max_frame_per_col,
-    )
-    generateConfigMem(
-        writer,
-        superTile.name,
-        st_config_bits,
-        output_csv,
+
+    # Only build the master-derived CSV layout when in FRAME_BASED mode
+    if config_bit_mode == ConfigBitMode.FRAME_BASED:
```
EverythingElseWasAlreadyTaken, line 572: In theory, this should also work with a supertile, but might need some adaptions. 

If it's really not possible, we have to throw an error at some point.

EverythingElseWasAlreadyTaken (commented):
Thanks for your PR.
I Just went through the code a have a few comments, but haven't done a functional test yet.


We also need some tests, but this can be also a seperate PR.

KelvinChung2000:
This in itself is not functional. This will also need the tile and fabric level fixed, which is at #994

MrAbhi19:
> Thanks for your PR. I Just went through the code a have a few comments, but haven't done a functional test yet.
> 
> We also need some tests, but this can be also a seperate PR.

Hi @EverythingElseWasAlreadyTaken, thanks for reviewing! I've done a quick functional check with the basic simulation for now. I am currently occupied with another PR, so I will follow up and add the tests in a separate PR as soon as possible.

PR #981, opened 2026-08-02
feat: add xsim support with fixes to HDL for supporting

### LINKED ISSUE #977: PR  #449 violated Verilog Syntax and broke 3rd Party Flows
**General**
PR #449 introduced sweeping changes to the HDL files, some of which are not in accordance to the Verilog standard. Tools with stricter enforcementthan our test setup no longer compile the fabric successfully (for example Vivado). Some of the changes might also accidentallyalter hardware behaviour.

**default_nettype none**
`` `default_nettype none `` is a compiler directive for the sole purpose of forcing explicit signal declarations, rendering implicit signal into compile time errors.
The PR pasted this directive at the top of most of our Verilog files, enforcing explicit signal declaration in our codebase.
In principle, I fully agree with this change, explicit declaration is much safer and forcing it will prevent accidental errors.
However, the previous default was implicit declarations in many modules, and while the rule of explicit declaration was enforced in every file, none of the files were actually changed to adhere to this new rule. So they are no longer able to be compiled. If you want to transition to explicit declaration, I would recommend first changing everything to the desired syntax, and only then using strict enforcement.

Note:
In `models_pack.v`, this directive is disabled again only a few lines later, I don't know if this was intentional or not:
https://github.com/FPGA-Research/FABulous/blob/8eff7dbcd5c2c0ad2dd96fb578d509f52967d30d/fabulous/fabric_files/FABulous_project_template_verilog/Fabric/models_pack.v#L20

To be honest, I am a bit concerned why the automated flow completely ignored this directive, as it did not flag any of the implicit declarations after being specifically told to not allow implicit declarations.

**SystemVerilog Syntax**
This one is quick, easy and not very relevant, but in `ConfigFSM.v`, SystemVerilog Syntax was introduced in a Verilog file, strict compilers will not allow this.

https://github.com/FPGA-Research/FABulous/blob/8eff7dbcd5c2c0ad2dd96fb578d509f52967d30d/fabulous/fabric_files/FABulous_project_template_verilog/Fabric/ConfigFSM.v#L63
 
Correct would have been `row_index <= NumberOfRows[4:0];` in Verilog.

**The Big One: Changing wire to reg**
In `Frame_Data_Reg.v`, `FrameData_I` and `RowSelect` were changed from wire to reg:

https://github.com/FPGA-Research/FABulous/blob/8eff7dbcd5c2c0ad2dd96fb578d509f52967d30d/fabulous/fabric_files/FABulous_project_template_verilog/Fabric/Frame_Data_Reg.v#L8-L11

First of all, Verilog stricly forbids to declare inputs as reg. And that is for a good reason, as an input reg is completely non-sensical (there is no `always` trigger to ever set any value to this reg). I assume that Yosys knows that the only allowed input type is net (wire) and just silently fixed that, but I wouldn't be certain that this will synthesize into the same design as with the correct wire declaration. Anyways, strict compilers will definitely not let you do this.

I am not sure if this change was accidental or if there was an intention of introducing clock dependency.

**Why wasn't this a pull request?**
I did NOT review the whole PR. I only noted the syntax errors, as they are very easy to find and actually need to be fixed to run our fabric in 3rd party flows. I have no idea if there are any other errors, that might introduce logical errors but not come with accompanied syntax violations. And I therefore did not want to instill false confidence in this fix.

Also, it is extremely hard to judge changes in a PR that encompasses a whopping 4000 lines of actual code change with the only goal given being "linting and formatting all hdl files". Judging if a given change works towards this goal or is a mistake gets complicated if to many changes are introduced together.
I would recommend splitting this up into small precise PRs, for example "Enforcing explicit signal declaration", otherwise these become impossible to review to high standards.

**What to do now**
To ensure FABulous remains a reliable framework for actual silicon tape-outs, I would recommend against using these HDL files in the state they are now for actual chip projects. Apart from the fixes listed here, we should either re-review the changes introduced or just roll it back to a known good state, reintroducing the changes made in manageable PRs bit by bit later.

We should also take a look at our automated testing suite. The fact these slipped past into production should be avoided in the future. Also, passing this suite should not be taken as "the proposed changes are without flaw and dont break anything".

KelvinChung2000:
Closes #977.

**fix(hdl)**. Every ANSI port in the template Verilog names `wire`, `input reg` in Frame_Data_Reg is `input wire` again, the `5'()` cast and the unsized `mem [32]` take their Verilog-2005 forms, and the `default_nettype wire` restore moves to end of file in models_pack.v and BlockRAM_1KB.v, where it sat after the first module and left the rest outside the `none` region. Generated fabric HDL, chip database and bitstream are byte-identical. `parameter reg [..]` in three files is left as is because xvlog, Yosys, iverilog and Verilator all accept it.

**feat(test)**. `SIMULATOR=xvlog` (Verilog) or `SIMULATOR=xvhdl` (VHDL) runs the existing testbench under Vivado xsim, from `task run-simulation` in a project, `run_simulation --simulator` in the REPL, or `task smoke-test SIMULATOR=...` at the repo root, where the value also picks the project language. xvlog reads the fabric as IEEE 1364-2005, so this is the check that would have caught #977. The VHDL `run-simulation` dispatcher no longer shells out to `task`, which had dropped CLI variables such as `DESIGN=`.

Verified locally with Vivado 2025.2: `task smoke-test SIMULATOR=<v>` passes for iverilog, xvlog, nvc and xvhdl, and a corrupted bitstream fails both xsim paths through the log grep.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -20,50 +20,30 @@ tasks:
       - uv run --directory . make livehtml
 
   smoke-test:
-    desc: Create the demo project and run the simulation flow
+    desc: Create the demo project and run the simulation flow (SIMULATOR=iverilog|xvlog|nvc|ghdl|xvhdl)
     # FABULOUS is the launcher prefix. It defaults to `uv run FABulous` for the
     # uv-managed dev/container environment; in the Nix devshell FABulous is on
     # PATH directly, so pass `FABULOUS=FABulous` to skip uv (whose .venv is a
     # different Python and ABI-incompatible with the devshell's packages).
+    #
+    # SIMULATOR also picks the project language: iverilog and xvlog build a
+    # Verilog project, nvc, ghdl and xvhdl a VHDL one.
     vars:
       FABULOUS: '{{.FABULOUS | default "uv run FABulous"}}'
+      SIMULATOR: '{{.SIMULATOR | default "iverilog"}}'
+      LANGUAGE_FLAG:
+        sh: 'case "{{.SIMULATOR}}" in nvc|ghdl|xvhdl) echo "-w vhdl" ;; esac'
+    preconditions:
+      # Membership, not emptiness: LANGUAGE_FLAG is legitimately empty for Verilog.
+      - sh: 'case "{{.SIMULATOR}}" in iverilog|xvlog|nvc|ghdl|xvhdl) exit 0 ;; *) exit 1 ;; esac'
+        msg: >-
+          Unknown SIMULATOR '{{.SIMULATOR}}'. Use iverilog or xvlog for Verilog,
+          or nvc, ghdl or xvhdl for VHDL.
     cmds:
       - rm -rf demo-smoke-test
-      - '{{.FABULOUS}} c demo-smoke-test'
+      - '{{.FABULOUS}} c demo-smoke-test {{.LANGUAGE_FLAG}}'
       - '{{.FABULOUS}} -p demo-smoke-test run "run_fab"'
-      - task --dir ./demo-smoke-test/Test fab-sim
-
-  smoke-test-xsim:
-    desc: Create the demo project and run the simulation flow under Vivado xsim
-    # Needs xvlog/xelab/xsim on PATH from a Vivado installation.
-    vars:
-      FABULOUS: '{{.FABULOUS | default "uv run FABulous"}}'
-    cmds:
-      - rm -rf demo-smoke-test
-      - '{{.FABULOUS}} c demo-smoke-test'
-      - '{{.FABULOUS}} -p demo-smoke-test run "run_fab"'
-      - task --dir ./demo-smoke-test/Test fab-sim-xsim
-
-  smoke-test-vhdl:
-    desc: Create the demo project and run the simulation flow
-    vars:
-      FABULOUS: '{{.FABULOUS | default "uv run FABulous"}}'
-    cmds:
-      - rm -rf demo-smoke-test
-      - '{{.FABULOUS}} c demo-smoke-test -w vhdl'
-      - '{{.FABULOUS}} -p demo-smoke-test run "run_fab"'
-      - task --dir ./demo-smoke-test/Test fab-sim
-
-  smoke-test-vhdl-xsim:
-    desc: Create the VHDL demo project and run the simulation flow under Vivado xsim
-    # Needs xvhdl/xelab/xsim on PATH from a Vivado installation.
-    vars:
-      FABULOUS: '{{.FABULOUS | default "uv run FABulous"}}'
-    cmds:
-      - rm -rf demo-smoke-test
-      - '{{.FABULOUS}} c demo-smoke-test -w vhdl'
-      - '{{.FABULOUS}} -p demo-smoke-test run "run_fab"'
-      - task --dir ./demo-smoke-test/Test fab-sim-xsim
+      - task --dir ./demo-smoke-test/Test fab-sim SIMULATOR={{.SIMULATOR}}
 
   ci:
     desc: Run code checks, tests, and the docs build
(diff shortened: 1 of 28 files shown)


NeverInControl:
In case you are open to suggestions:
You mentioned here that the intended purpose for the tailing `` `default_nettype wire`` is to reset the compiler directive so it won't cross file boundries. There is a more convenient way of doing this:
The `` `resetall `` compiler directive resets all set directives to their default value. Pasting this at the end of every file will prevent all compiler directives to ever cross file boundries, which is basically never intended. So you will also prevent bugs of accidentally not resetting directives added in the future. In addition, it spares you from having to know the default value of the compiler directives you try to reset (`wire` in this case).

NeverInControl:
In reguards to the `` `default_nettype none`` issue:

**Practical Discussion**
Independent of what the actual Verilog standard is in this case, the fact of the matter is, Vivado will not compile these HDL files with  `` `default_nettype none`` because it claims the input/output signals were not explicitly declared (preventing the "bare" style of declaration).
If we want the fabric to be emulatable on Xilinx, we need to fix this, either by omitting the compiler directive or by including net or variable type in the signal declaration.
Also, to prevent "It works on my machine"-syndrome, this project should always strive to fullfill the most strict standard available, which in this case is Vivado, as `iverilog` and `verilator` will also happily accept explicitly declared ports.

**Theoretical Discussion**
(Not that important)
In the Verilog standard (`IEEE 1364-2005`), chapter ``19.2 `default_nettype`` states:
"When the `default_nettype is set to none, all nets shall be explicitly declared. If a net is not
explicitly declared, an error is generated."

From chapters ``12.3.3 Port declarations`` and ``12.3.4 List of ports declarations``:
"If a port declaration does not include a net or variable type, then the port can be again declared in a net or
variable declaration." - This means a port declaration without wire or reg is always implicit, it does not make any distictions for ANSI style. And for non-ANSI style, it is quite clear that omitting this makes it implicit:"
module test(a,b,c,d,e,f,g,h);
input [7:0] a; // no explicit declaration - net is unsigned
input [7:0] b;
input signed [7:0] c;
input signed [7:0] d; // no explicit net declaration - net is signed
output [7:0] e; // no explicit declaration - net is unsigned
output [7:0] f;
output signed [7:0] g;
output signed [7:0] h; // no explicit net declaration - net is signed
wire signed [7:0] b; // port b inherits signed attribute from net decl.
wire [7:0] c; // net c inherits signed attribute from port
reg signed [7:0] f; // port f inherits signed attribute from reg decl.
reg [7:0] g; // reg g inherits signed attribute from port
endmodule
"
It is very possible I made a mistake here, but I could not find any hint that "Bare ANSI" is a special case and explicit.

So, as far as I could figure out, bare declarations are implicit, which is specifically prohibited by the compiler directive. Then Vivado flagging this as a violation is actually in accordance to the standard, while iverilog, verilator and Yosys are not. Which brings us back to the point I made in the issue, that the tools might not be as strict as they are needed to be.

[A possible explanation might be that explicit port declarations, especially CLK, were really uncommon for a long time, and the open source tools did not want to break legacy code for people who wanted to use the directive to just scan for implicit declarations in the body of modules?]

KelvinChung2000:
This issue is a mix of Verilator and us enabling Yosys to parse SystemVerilog. I have not checked what the changes claude made yet. But the lint side from Verilator needs a fix

EverythingElseWasAlreadyTaken (approved):
From my side everything looks fine. 
I've tested verilog and vhdl simultion with Vivado 2025.2, an everything seems to work just fine.

EverythingElseWasAlreadyTaken:
Before merging this, I would wait for feedback from @NeverInControl.
He wanted to go through the changes today or tomorrow.

NeverInControl:
The RTL if fixed as far as I can tell, happy with this, thanks!
Would have loved to see `` `resetall`` instead of a hardcoded `` `default_nettype wire`` to reset compiler directives back to default at the end of every file to prevent file boundry crossing for all directives at once and not assume the defaults manually, but thats just style and convenience. The main point of getting the files back to standard is achieved.

EverythingElseWasAlreadyTaken (approved):

PR #992, opened 2026-08-06
docs: overhaul and simplify the documentation setup

KelvinChung2000:
Simplifies the documentation setup and cuts what we maintain by hand.

**Single manifest.** The docs had their own `pyproject.toml`/`uv.lock`, so Dependabot had to open one PR touching two lockfiles — a shape it cannot handle, and the uv updates kept landing broken. The Sphinx dependencies move into a `docs` group in the root `pyproject.toml`, leaving one manifest and one lockfile.

**Docstrings are MyST.** They used to go through a regex pass that rewrote single backticks to double, fences to `.. code-block::`, and inserted blank lines around lists. A `myst-docstring` directive parses them with the MyST parser instead, so the markup lands in a parser that speaks it — the converter is gone.

**No Python API reference.** Nothing linked into it: a grep for python-domain roles across every hand-written page found zero. Keeping it building cost ~790 lines of templates, extensions and cross-reference workarounds, so it goes, along with `sphinx-autoapi`, `astroid` and `napoleon`. `conf.py` drops from 405 to 203 lines. The REPL and variable reference tables are unaffected — different generators, still there.

**Sphinx runs from the Taskfile.** `docs/Makefile` only assembled a `sphinx-build` line; `task docs-build` and `task docs-server` call it directly.

⚠️ **Breaking:** the 213 `generated_doc/fabulous/...` URLs will 404. Only the entry point gets a redirect — `sphinx-reredirects` expands a wildcard source against documents in the *current* build, so a pattern covering deleted pages matches nothing, warns, and fails under `-W`.

Verified with `task ci` (pre-commit + a `-W -n` docs build) and by content-diffing the surviving pages against the previous build: 51 of 53 identical, the exceptions being `genindex` (it listed the API objects) and `index` (the removed toctree entry).


KelvinChung2000:
@greptile-apps

PR #994, opened 2026-08-07
Fix: added support for FlipFlopChain mode in tile and switch_matrix generation. 

MrAbhi19:
**Summary:**
1. Updated the tile generation logic for `FlipFlopChain` config mode 
2. Updated the switch matrix generation logic for `FlipFlopChain` mode
3. Removed `addFlipFlopChain` and `addShiftRegister` generation modules, which will be unused after #976.

### DIFF
--- a/fabulous/fabric_generator/gen_fabric/gen_tile.py
+++ b/fabulous/fabric_generator/gen_fabric/gen_tile.py
@@ -404,7 +404,7 @@ def generateTile(
         )
 
     elif config_bit_mode == ConfigBitMode.FLIPFLOP_CHAIN and tile.globalConfigBits > 0:
-        writer.addComment("configuration storage latches", onNewLine=True)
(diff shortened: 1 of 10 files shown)


fabulous/fabric_generator/gen_fabric/gen_tile.py
```diff
@@ -403,12 +403,26 @@ def generateTile(
             emulateParamPairs=[("Emulate_Bitstream", "Emulate_Bitstream")],
         )
 
+    elif config_bit_mode == ConfigBitMode.FLIPFLOP_CHAIN and tile.globalConfigBits > 0:
+        writer.addComment("configuration storage latches", onNewLine=True)
```
KelvinChung2000:
```suggestion
        writer.addComment("configuration storage FlipFlop", onNewLine=True)
```

PR #1029, opened 2026-09-09
ci: authenticate the svlint release lookup and retry its rate limit

KelvinChung2000:
`pre-commit` failed on #1028 with `HTTP Error 403: rate limit exceeded` from the
svlint hook. Not caused by that PR, and it can hit any branch.

`svlint_wrapper` resolves its pinned release through an unauthenticated
api.github.com call, so it shares a 60 requests per hour quota with everything
else on the runner's IP, and nothing caches the binary between runs. The lookup
now sends `GITHUB_TOKEN`, which the workflow passes, for 1000 per hour instead.

The retry loop was also dead code for this case: GitHub reports an exhausted
rate limit as 403, which is not in `TRANSIENT_STATUSES`, so `fetch_json`
re-raised on attempt 1 of 5. The failing log shows exactly that. A 403 now
counts as transient when the response carries `x-ratelimit-remaining: 0`, and a
403 that is a real permission error still raises immediately.

Verified locally: cold `XDG_CACHE_HOME` downloads with and without a token, the
403 branch retries all 5 attempts, a bare 403 and a 404 still raise at once, and
`pre-commit run --all-files` passes.

### DIFF
--- a/.github/workflows/pre-commit-hooks.yml
+++ b/.github/workflows/pre-commit-hooks.yml
@@ -21,6 +21,10 @@ jobs:
         uv sync --locked --all-extras --dev
 
     - name: run pre-commit
+      # svlint_wrapper resolves its pinned release through the GitHub API, which
+      # rate-limits unauthenticated runners off a shared IP.
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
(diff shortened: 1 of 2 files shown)


KelvinChung2000:
Discarding this; the rate-limit flake will be handled another way.

PR #1039, opened 2026-09-17
feat: configurable UserCLK ladder direction

EverythingElseWasAlreadyTaken:
Add fabric.csv parameter `UserCLKDirection` (S2N default, N2S, W2E, E2W). The ladder is expressed as a predecessor offset keyed by the side the clock enters each tile; gen_fabric and the supertile wrapper derive tile-to-tile chaining and boundary ports from it, and the GDS pin-order config places UserCLK/UserCLKo on the matching sides. Default output is byte-identical.


fabulous/fabric_definition/define.py
```diff
@@ -90,6 +90,38 @@ def opposite(self) -> "Side":
                 return Side.ANY
 
 
+# Grid offset (dx, dy) of the tile whose UserCLKo feeds a tile's UserCLK, keyed by
+# the side the clock enters. y grows downwards, so SOUTH means "the row below".
```
KelvinChung2000, line 94: We really need to get the origin swap in; otherwise, the debt on it will keep rolling.

fabulous/fabric_generator/gen_fabric/gen_tile.py
```diff
@@ -620,6 +628,9 @@ def generateSuperTile(
         If True, the UserCLK port will not be generated or connected
     config_bit_mode : ConfigBitMode
         The configuration bit mode to use (frame-based or FlipFlop chain)
+    user_clk_side : Side
+        Side on which UserCLK enters each subtile; the chain runs towards the
+        opposite side. Default SOUTH (S→N ladder).
```
KelvinChung2000:
```suggestion
        opposite side. Default SOUTH (S->N ladder).
```

KelvinChung2000 (approved):

